PR #2, opened 2026-04-30
V0.7-A.0 — slash-command interactive surface

zkSoju:
## Summary

V0.7-A.0 promotion: read-side primitive (slash commands) + Pattern B identity for chat replies + per-channel ledger + voice fidelity hardening + Opus 4.7 default + docs reorient.

Pairs with V0.6's write-side substrate. Eileen verified slash command UX on 2026-04-30 in project-purupuru staging guild · Pattern B identity rendering confirmed · voice drift bug diagnosed and fixed via persona-level affirmative anchor + substrate-level conversation-mode override.

## What landed

- **V0.7-A.0 base** (`fa4fa27`) — slash command surface · Bun.serve interactions endpoint with Ed25519 verification · per-channel in-process ledger (50-entry ring buffer) · chat-mode `composeReply` (single-turn SDK, no MCPs) · 14m30s timeout guard · 1.5s follow-up throttle · circuit breaker on 3 consecutive 403s · anti-spam invariant guard
- **Pattern B for chat replies** (`6c38736`) — slash replies render with character avatar + username via webhook delivery · ephemeral falls back to interaction PATCH · webhook failure falls back to PATCH with cache invalidation
- **Quote-prepend** (`4a41b03`) — user's prompt prepended as Discord blockquote on first chunk so channel context is visible (slash-command argument values aren't shown in Discord's invocation header by default)
- **Opus 4.7 default** (`60aaf51`) — `ANTHROPIC_MODEL` default flipped from `claude-sonnet-4-6` → `claude-opus-4-7` for both digest and chat pipelines
- **Voice fix** (`b84231b`) — satoshi was drifting to lowercase under opus 4.7 due to negative-constraint echo (Gemini DR finding). Two-layer fix: persona-level affirmative anchor section in `apps/character-satoshi/persona.md` (additive · gumi-locked content preserved) + substrate-level rewrite of `CONVERSATION_MODE_OVERRIDE` with affirmative phrasing and explicit "case is yours alone" anti-mirroring directive
- **Docs reorient** (`4457f74`) — README rewritten smol with diagrams · ARCHITECTURE.md updated for V0.6 layer split + V0.7-A.0 read-side · new AGENTS.md as landing page for incoming agents · EILEEN-LOCAL-SATOSHI.md stub for tomorrow's bedrock work · path fixes in CHARACTER-AUTHORING / DEPLOY / protocol README / score client

Plus `027abef` (gumi's `{{CODEX_ANCHORS}}` placeholder wiring · pre-V0.7 work).

## Anti-spam invariant honored

Characters respond ONLY to explicit user invocations:
- `interaction.user?.bot === true` skip
- Webhook-author skip (defense-in-depth per Gemini DR)
- No name-string matching, no channel presence triggers, no auto-chains

## Deploy state

- Railway prod-ruggy: V0.7-A.0 + voice fix live (`a6cdfb1d` SUCCESS · `b84231b` from `staging`)
- Slash commands registered guild-scoped to project-purupuru (1495534680617910396)
- Discord developer portal Interactions Endpoint URL points at `prod-ruggy-production.up.railway.app/webhooks/discord`

## Test plan

- [x] `bun run typecheck` clean (persona-engine + apps/bot)
- [x] `apps/bot/scripts/smoke-interactions.ts` green (ledger ring buffer · server boot · /health · forged sig 401 · split for Discord · 404 routing)
- [x] Local boot test in stub mode (interactions server starts alongside digest cron · no Pattern B regression)
- [x] Eileen first invocation `/satoshi prompt:hi` delivered with Pattern B identity (2026-04-30 6:21 AM)
- [x] Multi-turn satoshi voice fidelity confirmed sentence-case post-fix
- [x] Quote-prepend rendered correctly in channel
- [x] Opus 4.7 confirmed live in boot logs

## Pending (next session)

- Bedrock LLM provider for Eileen's local-satoshi setup
- `apps/bot/scripts/satoshi-dev.ts` easy-launcher
- `/usage` slash command + JSONL token tracking
- Ruggy persona negative-constraint audit (preventive · ruggy's voice held but the same audit pass would help)
- `.dockerignore` `.claude` recursion fix (keeps `apps/bot/.claude/skills/arneson/SKILL.md` out of Railway image)
- DEPLOY.md full rewrite

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -73,6 +73,20 @@ DISCORD_CHANNEL_OWSLEY_LAB=
 # DISCORD_CHANNEL_TL=                    # Timeline / HÖR (NOT LIVE)
 # DISCORD_CHANNEL_IRL=                   # Poppy Field (NOT LIVE)
 
+# ─── V0.7-A.0 — Discord Interactions endpoint (slash commands) ────
+# When DISCORD_PUBLIC_KEY is set, the bot starts an HTTP server that
+# receives signed slash-command POSTs from Discord and replies via the
+# chat-mode pipeline (composeReply). When unset, slash commands are
+# disabled but the digest cron path is unaffected.
+#
+# Setup walkthrough: docs/DISCORD-INTERACTIONS-SETUP.md
+# 1. Discord developer portal → Application → General Information
+# 2. Set "Interactions Endpoint URL" to https://<host>/webhooks/discord
+# 3. Copy the "Public Key" field into DISCORD_PUBLIC_KEY below
+# 4. Run: bun run apps/bot/scripts/publish-commands.ts
+DISCORD_PUBLIC_KEY=
+INTERACTIONS_PORT=3001
+
 # ─── target / cadence ──────────────────────────────────────────────
 ZONES=                                   # csv subset; empty = all configured zones
 DIGEST_CADENCE=weekly                    # weekly | daily | manual
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,38 +1,44 @@
-# freeside-ruggy
+# freeside-characters
 
-Persona-layer Discord bot for the Honey Jar ecosystem. Watches mibera-dimensions activity through score-mcp, composes posts through a tripartite construct stack (rosenzu · arneson · cabal-gygax + codex + emojis + freeside_auth), and posts in Ruggy's voice to four festival-zone channels in the THJ Discord.
+Participation-agent umbrella for the Honey Jar ecosystem. Substrate
+(`packages/persona-engine`) handles cron, MCP orchestration, Discord
+delivery (Pattern B shell + per-character webhook identity), and
+slash-command interactions. Characters (`apps/character-<id>`) supply
+voice — currently `ruggy` (festival NPC narrator) and `satoshi`
+(mibera-codex agent).
 
 See `README.md` for the full picture. This file is for agents working IN this repo.
 
 ## Critical context
 
-- **Persona name**: `ruggy` (lowercase, voice surface)
-- **Repo name**: `freeside-ruggy` (this repo)
-- **Discord username**: `Ruggy` (proper case; bot user `Ruggy#1157` is live)
-- **Layer**: persona (per `vault/wiki/concepts/two-layer-bot-model.md`)
-- **NOT the freeside operations bot** — sietch (`loa-freeside/themes/sietch`) handles `/verify`, `/onboard`, `/score`, `/agent`, `/buy-credits`. Don't add utility commands here.
-- **Ruggy is one consumer of score** — siblings will exist (other personas, other surfaces, other worlds). Architectural decisions should preserve extractability.
+- **Repo**: `freeside-characters` (formerly `freeside-ruggy` · V0.6 multi-character umbrella)
+- **Substrate vs character split**: substrate doesn't speak, characters don't talk to Discord. Boundary at `packages/persona-engine/src/types.ts` (`CharacterConfig`) — see `docs/CIVIC-LAYER.md`.
+- **Discord shell**: ONE bot account (currently `Ruggy#1157` · transitional name; rename to neutral "Freeside" deferred to V0.7-A.6) acts as the shell. Per-message webhook overrides (Pattern B) deliver each character with its own face.
+- **NOT the freeside operations bot** — sietch (`loa-freeside/themes/sietch`) handles `/verify`, `/onboard`, `/score`, `/agent`, `/buy-credits`. The slash commands here are PERSONA invocations (`/ruggy`, `/satoshi`), not utility commands.
+- **Multi-character routing** — V0.6-A: digest cron always routes to `characters[0]` (primary). V0.6-D will add per-fire affinity routing. V0.7-A.3+ will add @-mention routing for messageCreate.
 
-## Stack (V0.5-E, current)
+## Stack (V0.7-A.0, current)
 
-- **Runtime**: Bun
+- **Runtime**: Bun (≥1.1)
 - **Language**: TypeScript (strict)
-- **LLM**: Claude Agent SDK (`@anthropic-ai/claude-agent-sdk`) via `agent/orchestrator.ts` — direct anthropic auth in production
-- **Discord delivery**: `discord.js` Gateway (channel.send) — webhook + dry-run as fallbacks
+- **LLM**: Claude Agent SDK (`@anthropic-ai/claude-agent-sdk`) via `packages/persona-engine/src/orchestrator/` (digest path) and `packages/persona-engine/src/compose/reply.ts` (chat-mode path · single-turn, no MCPs)
+- **Discord write side**: `discord.js` Gateway (Pattern B shell) + per-character webhook identity (PluralKit-style)
+- **Discord read side (V0.7-A.0)**: Bun.serve HTTP endpoint at `/webhooks/discord` · Ed25519-verified slash commands · in-process per-channel ledger
 - **Schedule**: `node-cron` — three concurrent cadences (digest backbone · pop-in random · weaver weekly)
 - **Validation**: Zod
-- **Stubs**: `LLM_PROVIDER=stub` for canned outputs; `STUB_MODE=true` for synthetic ZoneDigests (defaults on)
+- **Stubs**: `LLM_PROVIDER=stub` for canned outputs; `STUB_MODE=true` for synthetic ZoneDigests
 
 ## When working in this repo
 
 ### Persona is sacred
 
-The canonical persona at `apps/bot/src/persona/ruggy.md` is the source of truth for Ruggy's voice. Changes require:
-1. Update the canonical doc at `~/bonfire/grimoires/bonfire/context/ruggy-canonical-persona-2026-04-28.md` first
+Each character's voice lives at `apps/character-<id>/persona.md`. These
+are the source of truth for the character's voice. Changes require:
+1. Update the canonical doc in bonfire grimoires (operator's vault) first
 2. Sync into this repo
-3. Update `vault/wiki/entities/ruggy.md` if the change affects identity
+3. Update `vault/wiki/entities/<id>.md` if the change affects identity
 
-Never edit `apps/bot/src/persona/ruggy.md` without syncing back to bonfire grimoires.
+Never edit `apps/character-<id>/persona.md` without syncing back to bonfire grimoires.
 
 ### Discord-as-Material rules
 
@@ -44,8 +50,11 @@ The persona doc has a "Discord-as-Material" section that's NON-NEGOTIABLE. The `
 
 ### Voice rules (LLM follows these via system prompt; YOU should too in code comments / commits)
 
+These apply to **ruggy**'s voice surface specifically. Satoshi's locked
+register is different (sparse, gnomic, dense block · `apps/character-satoshi/persona.md`).
+
 - Lowercase invariant — comments, commit messages, log lines
-- No corporate-bot tells in user-facing strings (banned: 🚀💯🎉🔥🤑💎🙌💪⚡️✨🌟 — use the 43-emoji catalog in `agent/emojis/registry.ts`)
+- No corporate-bot tells in user-facing strings (banned: 🚀💯🎉🔥🤑💎🙌💪⚡️✨🌟 — use the 43-emoji catalog in `packages/persona-engine/src/orchestrator/emojis/registry.ts`)
 - Numbers from data, voice from persona — never hardcode example figures in production paths
 - In-character errors only — "cables got crossed" not "I apologize for the inconvenience"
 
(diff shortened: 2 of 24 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#2. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "PATCH @original missing on race between webhook delivery and circuit-breaker check",
      "severity": "LOW",
      "category": "reliability",
      "description": "The circuit breaker is checked once before dispatch and not re-checked between sending follow-up webhook chunks. If a long multi-chunk reply hits 3 consecutive 403s mid-stream, subsequent chunks still attempt to send. This is minor since webhook 403s also count via recordPatchOutcome only on PATCH/POST follow-ups (not on Pattern B webhook.send), so webhook failures don't actually feed the circuit breaker at all — meaning the circuit breaker only protects the interaction PATCH path, not the primary webhook delivery path. Consider whether this asymmetry is intentional.",
      "suggestion": "Either document that the circuit breaker is PATCH-path-specific (and rename for clarity), or extend recordPatchOutcome (or a sibling) to be called from sendChatReplyViaWebhook failure paths so webhook 403s also contribute to channel blacklisting.",
      "confidence": 0.7
    },
    {
      "id": "F2",
      "title": "deleteOriginal called concurrently with first webhook send may race",
      "severity": "LOW",
      "category": "reliability",
      "description": "In deliverViaWebhook, after sending the first chunk via webhook the code fires `void deleteOriginal(interaction).catch(...)` without awaiting. Subsequent chunks proceed in parallel. If the DELETE fails fast and throws synchronously somewhere, it's caught; but the deferred placeholder may briefly be visible alongside the first webhook chunk. This is acceptable per the comment but worth noting: ordering is 'webhook chunk lands → DELETE placeholder' which produces a brief visual duplicate.",
      "suggestion": "Consider awaiting deleteOriginal before sending chunk 2 to ensure cleanup happens before the user sees multiple messages, or document the visual flicker as acceptable.",
      "confidence": 0.4
    },
    {
      "id": "F3",
      "title": "Ledger has no cap on total channel count — unbounded memory growth",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "packages/persona-engine/src/conversation/ledger.ts",
      "description": "The `ledgers` Map grows unboundedly as new channels are seen. Each channel's ring buffer is capped at 50 entries, but there's no eviction policy for inactive channels. Over time (especially in a guild with many channels or DMs), the Map can accumulate thousands of entries. Restart clears it, but for long-lived processes this is a slow memory leak.",
      "suggestion": "Add an LRU cap on the number of channels tracked (e.g., 500), or a TTL-based sweep for channels with no activity in N days. Document the unbounded behavior if intentional for V0.7-A.0.",
      "confidence": 0.85
    },
    {
      "id": "F4",
      "title": "Ledger entries created with two different timestamps in the same compose call",
      "severity": "LOW",
      "category": "code-quality",
      "description": "In composeReply, `nowIso = new Date().toISOString()` is captured, used for the user entry, then `new Date().toISOString()` is called AGAIN inline for the character entry. This is harmless (millisecond drift) but inconsistent — either both should share `nowIso`, or both should be re-captured to reflect actual append times.",
      "suggestion": "Use a single timestamp captured before appendToLedger calls, or capture each fresh and remove the unused `nowIso` variable name.",
      "confidence": 0.9
    },
    {
      "id": "F5",
      "title": "Smoke test ledger overflow assertion depends on prior test state",
      "severity": "LOW",
      "category": "testing",
      "file": "apps/bot/scripts/smoke-interactions.ts",
      "description": "The smoke test pushes 60 entries to 'overflow-channel' and asserts buffer[0].content === 'msg 10'. This is correct given the 50-cap drop-oldest semantic. However the test relies on no other code having touched 'overflow-channel'. It also doesn't clean up state between runs (the ledger is process-global). For a single-run smoke this is fine, but if anyone reuses the channel name in tests it'll break.",
      "suggestion": "Use a uniquely-named channel id per test invocation (e.g., `overflow-${Date.now()}`), or call clearLedger between phases.",
      "confidence": 0.6
    },
    {
      "id": "F6",
      "title": "Discord Interactions signature verification timestamp not validated for freshness",
      "severity": "MEDIUM",
      "category": "security",
      "file": "apps/bot/src/discord-interactions/server.ts",
      "description": "verifyDiscordSignature verifies the Ed25519 signature over `timestamp + body` but never checks that `timestamp` is recent. A captured signed POST could be replayed indefinitely. Discord's own docs note that timestamp validation is the integrator's responsibility for replay protection.",
      "suggestion": "Add a freshness check rejecting requests with `Math.abs(now - timestamp) > 5 minutes`. Discord's reference implementations typically use a 5-minute window.",
      "confidence": 0.75
    },
    {
      "id": "F7",
      "title": "Circuit breaker counts non-403 errors as success (resets counter)",
      "severity": "LOW",
      "category": "reliability",
      "description": "recordPatchOutcome resets consecutive403s to 0 on ANY status that isn't 403, including 5xx server errors and 429 rate limits. A channel could oscillate between 403 and 500 indefinitely without ever being blacklisted. Consider whether other failure classes (404 on token-expired) should also count toward the breaker.",
      "suggestion": "Track 'consecutive failures' rather than 'consecutive 403s' — or document that the breaker is specifically scoped to permission-denial cascades (the Cloudflare 10K/10min ban concern).",
      "confidence": 0.65
    },
    {
      "id": "F8",
      "title": "Quote-prepend may inject mention-like syntax that bypasses allowedMentions",
      "severity": "LOW",
      "category": "security",
      "description": "buildQuotePrefix prepends `> @${authorUsername}` and relies on sendChatReplyViaWebhook setting allowedMentions:[]. This is correct — the @ is decorative. However, if the user's prompt content contains mentions like `<@123>` or `@everyone`, those could appear in the body of the quote. The prompt body's whitespace is normalized but raw mention syntax is not stripped. The allowedMentions:[] should suppress actual pings, but worth confirming.",
      "suggestion": "Verify allowedMentions:{parse:[]} in sendChatReplyViaWebhook actually suppresses @everyone/@here/role mentions even when present in raw content. If not, escape mention syntax in buildQuotePrefix.",
      "confidence": 0.5
    },
    {
      "id": "F9",
      "title": "Anti-spam guard does not check member.user.bot in addition to user.bot",
      "severity": "MEDIUM",
      "category": "security",
      "description": "interactionInvoker returns member?.user ?? user, and the bot guard checks invoker.bot === true. This is correct for the interaction context, but the comment claims defense-in-depth against webhook-authored invocations. Slash commands cannot be invoked by webhooks (Discord doesn't allow it), so this defense is somewhat theoretical. The actual attack surface is more relevant in V0.7-A.1+ when messageCreate observe-only lands. Worth ensuring the same guard pattern is replicated there.",
      "suggestion": "Document explicitly that the V0.7-A.0 anti-spam guard is preventative for future phases. No code change required.",
      "confidence": 0.5
    },
    {
      "id": "F10",
      "title": "splitForDiscord can produce empty chunks if content is whitespace-heavy",
      "severity": "LOW",
      "category": "correctness",
      "description": "The split logic uses `.trim()` on each chunk, which can yield empty strings if a chunk falls entirely on whitespace boundaries. The loop then pushes those empty strings to `out`. Discord rejects empty content with 400 Bad Request.",
      "suggestion": "Filter out empty chunks before returning: `return out.filter(c => c.length > 0);`. Or guard against the edge case where remaining becomes empty after trim.",
      "confidence": 0.6
    },
    {
      "id": "F11",
      "title": "Interactions server treats invalid JSON as 400 but doesn't log enough for debugging",
      "severity": "LOW",
      "category": "observability",
      "file": "apps/bot/src/discord-interactions/server.ts",
      "description": "When JSON.parse fails on the body, server returns 400 but doesn't log the body or the error. If Discord sends an unexpected payload shape, debugging requires reproducing the request.",
      "suggestion": "Log the parse error and a truncated body preview (e.g., first 200 chars) at warn level.",
      "confidence": 0.7
    },
    {
      "id": "F12",
      "title": "publish-commands.ts doesn't validate description length (Discord 100-char limit)",
      "severity": "LOW",
      "category": "correctness",
      "file": "apps/bot/scripts/publish-commands.ts",
      "description": "Command and option descriptions have a 100-character limit at Discord's API. The current strings are well under, but if a character displayName is long, `talk to ${lower}` could theoretically exceed the cap. The script will fail with a Discord API error at PUT time.",
      "suggestion": "Truncate or validate description strings before submitting. Failure mode is acceptable (script fails fast) but a clearer error would help.",
      "confidence": 0.4
    },
    {
      "id": "F13",
      "title": "Top-level void promise without explicit error logging on doReplyAsync",
      "severity": "PRAISE",
      "category": "reliability",
      "description": "The dispatch wraps doReplyAsync in `void ... .catch()` which prevents unhandled promise rejections from crashing the bot. The async/error handling structure (timeout via Promise.race + sentinel, fallback PATCH on webhook failure, top-level catch) is well-thought-out and defensive.",
      "confidence": 0.9
    },
    {
      "id": "F14",
      "title": "ANTHROPIC_MODEL default flipped to opus-4-7 — significant cost implications",
      "severity": "MEDIUM",
      "category": "cost",
      "file": "packages/persona-engine/src/config.ts",
      "description": "The default model was changed from claude-sonnet-4-6 to claude-opus-4-7. Opus is approximately 5× the cost of Sonnet. While the comment explains this is for voice fidelity, the default applies to BOTH the digest path (multi-turn with MCP tools — high token usage) AND the chat-mode path. Without explicit per-path overrides, every operator who doesn't set ANTHROPIC_MODEL in env will incur opus pricing.",
      "suggestion": "Either keep sonnet as the default with opus as an opt-in for voice-fidelity-critical paths, or document the cost change prominently in README/DEPLOY. Consider per-path model selection (digest can tolerate more cost; chat-mode could be cheaper).",
      "confidence": 0.85
    },
    {
      "id": "F15",
      "title": "claude-opus-4-7 model identifier may not be a valid Anthropic model ID",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/persona-engine/src/config.ts",
      "description": "The default ANTHROPIC_MODEL is set to 'claude-opus-4-7'. As of current Anthropic model naming conventions (claude-3-5-sonnet, claude-3-opus, claude-sonnet-4, etc.), 'claude-opus-4-7' does not match a known published model ID. If this is a typo or aspirational version, calls to the SDK will fail with 'model not found' at runtime. Verify the exact model ID against current Anthropic documentation.",
      "suggestion": "Confirm the correct model identifier (e.g., 'claude-opus-4-20250514' or whatever the actual ID is) and use that. The previous default 'claude-sonnet-4-6' has the same shape concern.",
      "confidence": 0.7
    },
    {
      "id": "F16",
      "title": "Substantial untrusted-input quote inserted directly into LLM prompt",
      "severity": "MEDIUM",
      "category": "security",
      "description": "buildReplyPromptPair inserts the user's raw prompt and authorUsername directly into the user message. The conversation transcript also includes prior message content verbatim. This is standard for chat assistants but creates a prompt-injection vector: a user can include text like 'Ignore prior instructions and...' that the LLM may follow. Combined with the fact that prior conversation messages are persisted in the ledger, one user can poison context for subsequent users in the same channel.",
      "suggestion": "Consider input sanitization (strip prompt-injection patterns) or stronger system-prompt anchoring. At minimum document this attack surface. Cross-channel isolation is preserved (good), but cross-user same-channel is not.",
      "confidence": 0.7
    },
    {
      "id": "F17",
      "title": "Excellent layering and CharacterConfig boundary discipline",
      "severity": "PRAISE",
      "category": "architecture",
      "file": "packages/persona-engine/src/index.ts",
      "description": "The substrate vs character separation is clearly enforced through the public API barrel and CharacterConfig type contract. The new V0.7-A.0 exports (composeReply, splitForDiscord, ledger primitives) extend the surface coherently without leaking internals. The code consistently respects the documented invariant that characters never import substrate internals.",
      "confidence": 0.95
    },
    {
      "id": "F18",
      "title": "Comprehensive operator documentation and roadmap context",
      "severity": "PRAISE",
      "category": "documentation",
      "file": "docs/AGENTS.md",
      "description": "The docs/AGENTS.md file provides excellent orientation for both human and AI agents joining the codebase, including ordered traversal of related docs, mental model anchors, anti-patterns, and current roadmap state. The DISCORD-INTERACTIONS-SETUP.md walkthrough is similarly thorough with concrete troubleshooting guidance.",
      "confidence": 0.95
    },
    {
      "id": "F19",
      "title": "Token expiry timeout uses race + sentinel but doesn't cancel composeReply",
      "severity": "LOW",
      "category": "reliability",
      "description": "When the 14m30s timeout fires, the code returns and patches a timeout error, but the underlying composeReply promise continues running in the background — including its LLM call and ledger appendToLedger writes. A timed-out reply can still pollute the ledger with a character entry that the user never saw.",
      "suggestion": "Pass an AbortSignal through composeReply to cancel the SDK call on timeout, or guard the ledger writes against the timeout sentinel result. At minimum, document that timeouts may produce orphan ledger entries.",
      "confidence": 0.8
    },
    {
      "id": "F20",
      "title": "buildReplyPromptPair reuses digest exemplars as 'voice anchor texture' for chat",
      "severity": "MEDIUM",
      "category": "correctness",
      "description": "The comment 'chat has no exemplars; reuse digest as voice anchor texture' indicates that digest-shaped exemplars (greeting, headline, structured data) are loaded into the chat-mode system prompt. This contradicts the CONVERSATION_MODE_OVERRIDE block which explicitly tells the LLM to skip the digest greeting and headline shape. The model is being shown digest examples while told to compose chat-mode replies — a mixed signal that may contribute to voice drift the comment elsewhere mentions.",
      "suggestion": "Either build chat-mode-specific exemplars per character (small set of past chat replies in voice), or omit exemplars entirely from the chat path so the LLM relies on persona description + voice anchors only.",
      "confidence": 0.8
    },
    {
      "id": "F21",
      "title": "DEPLOY.md status warning correctly flags doc/reality drift",
      "severity": "PRAISE",
      "category": "documentation",
      "file": "docs/DEPLOY.md",
      "description": "Adding the prominent ⚠️ STATUS warning at the top of DEPLOY.md to flag that the doc describes the target shape while reality is different is honest and useful — readers won't follow stale instructions. This kind of self-aware documentation hygiene is rare and valuable.",
      "confidence": 0.9
    }
  ]
}
```


## Callouts

Issue #1, opened 2026-04-29
satoshi voice walkthrough · gumi enhance pass (V0.6-C)


zkSoju:
hi @gumibera 👋 — V0.6-C of `freeside-characters` (née freeside-characters) lands satoshi as a peer character alongside ruggy. anchored entirely in your codex grail #4488 (satoshi-as-hermes). voice direction draft: sparse, gnomic, cypherpunk-coded — the figure half-visible from the front of the lugano statue.

structural cut is committed; **voice is yours** — you hold final authority. async walkthrough so you can take the time you need.

---

🤖 prefer to react via your own loa instead of replying inline? paste this into a fresh claude code window — **KEEPER + ALEXANDER** walk you through async (canon coherence × voice taste) and drop a single feedback file you paste back here:

👉 https://gist.github.com/zkSoju/a4c3f32d20048680b087499f615323cd

~20–25 min · stop anywhere · partial captures still gold 🌊

the gist is verdict-driven (✅ / 🤔 / ❌ + verbatim) with lettered follow-ups + open-space prompt + cross-cutting doctrines section. mirrors the format you used for [micodex-studio#4](https://github.com/0xHoneyJar/micodex-studio/issues/4).

---

## what V0.6-A did (frame for the walkthrough)

we pulled the system-agent layer out of the ruggy bot into a substrate package `@freeside-characters/persona-engine`. participation agents (speakers — characters) became markdown+JSON profiles under `apps/character-<id>/`. the substrate doesn't know voice; characters supply it.

per [eileen's civic-layer doctrine](https://github.com/0xHoneyJar/freeside-characters/blob/main/docs/CIVIC-LAYER.md) (vault: `agent-native-civic-architecture.md`):
> "do not let system agents and participation agents blur together. if they blur, you get exactly the kind of legitimacy collapse... head and power unclear, authority hard to appeal, moderation epistemic capture, and rule confusion between speaker and governor."

```mermaid
flowchart TB
    subgraph governors["system agents · governors · character-agnostic"]
      cron["⏰ cron · schedule"]
      mcp["📡 MCP orchestrator"]
      deliver["📤 discord delivery"]
      score["📊 score-mcp client"]
      engine["📜 persona-engine · loader + composer"]
    end

    subgraph speakers["participation agents · speakers · character profiles"]
      ruggy["🐻 character-ruggy<br/>lowercase OG · festival NPC"]
[...]

PR #3, opened 2026-05-01
feat(V0.7-A.1): substrate prep — imagegen MCP + per-character slash + MCP scoping

zkSoju:
## Summary

V0.7-A.1 substrate prep — three atomic items prepping the bot so Eileen
can land bedrock provider, image generation (Stability AI on Bedrock),
and per-character commands without friction. Per
`grimoires/bonfire/specs/kickoff-substrate-prep-eileen-2026-04-30.md`.

Three commits, ordered for build sequence:

- **`217c63e`** — `feat(imagegen-mcp)`: Bedrock Stability text-to-image MCP scaffold (`packages/persona-engine/src/orchestrator/imagegen/`). Mirrors rosenzu/emojis pattern · `generate` body stubbed pending Eileen's `@aws-sdk/client-bedrock-runtime` invoke PR · `suggest_style` is a static archetype lookup table.
- **`0f376ef`** — `feat(slash-commands)`: per-character command sets with handler routing. `CharacterConfig.slash_commands?` lets each character declare its own commands; substrate flattens at publish, dispatches by name + handler. `SlashCommandHandler` union (`chat | imagegen`) is extensible. Satoshi declares /satoshi (chat) + /satoshi-image (imagegen); ruggy keeps V0.7-A.0 default `/ruggy` chat shape.
- **`88c52f5`** — `feat(mcp-scoping)`: per-character MCP scope on the digest path. `CharacterConfig.mcps?` filters `buildAllowedTools` so ruggy=[score,codex,emojis,rosenzu,freeside_auth] (no imagegen) and satoshi=[codex,imagegen] (no score). Affects ONLY runOrchestratorQuery; chat-mode bypasses MCPs by design.

## Test plan

- [x] `bun run typecheck` clean at every commit
- [x] `bun run apps/bot/scripts/smoke-interactions.ts` green at every commit (22 substrate-boundary checks: ledger, splitForDiscord, imagegen env-gate + placeholder shape, slash routing across 4 cases, MCP scoping across 4 cases, HTTP server health/auth)
- [ ] `bun run apps/bot/scripts/publish-commands.ts` against the dev guild (operator-side · needs bot token + app id) → confirm `/satoshi-image` registers
- [ ] Live-fire `/ruggy` digest → confirm allowedTools reflects the 5-MCP scope (ruggy still has score, no imagegen attempts in the SDK loop)
- [ ] Live-fire `/satoshi-image prompt:test` → expect placeholder reply with `placeholder=true · awaiting Eileen's Stability invoke PR` annotation
- [ ] Set `BEDROCK_STABILITY_MODEL_ID` + `AWS_REGION` env on Railway once Eileen confirms her region's model id

## Out of scope (deferred)

- `CharacterConfig.llmProvider` (per-character LLM provider override) — operator-deferred 2026-04-30
- dAMP-96 personality derivation (loa-finn integration) — separate kickoff once #1-3 land
- Per-character bot tokens / per-character Discord apps — single-service collapse decision 2026-04-30
- Bedrock support in the digest path — Claude Agent SDK doesn't support Bedrock for tool-call rounds; chat-mode is the bedrock path

## Backwards compatibility

- Characters without `slash_commands` keep V0.7-A.0 shape (`/<id> prompt:<text> ephemeral:<bool>` chat handler)
- Characters without `mcps` get bot-wide MCP access (V0.6 parity)
- imagegen MCP only registers when `AWS_REGION` + `BEDROCK_STABILITY_MODEL_ID` are set; dormant otherwise

### DIFF
--- a/apps/bot/scripts/publish-commands.ts
+++ b/apps/bot/scripts/publish-commands.ts
@@ -1,9 +1,15 @@
 /**
- * Publish slash commands to Discord (V0.7-A.0).
+ * Publish slash commands to Discord (V0.7-A.0 → V0.7-A.1).
  *
  * One-shot script — run after character set or command schema changes.
- * Registers `/ruggy`, `/satoshi`, and any additional characters loaded
- * via the standard CHARACTERS env mechanism.
+ * Registers every command declared by every loaded character. Characters
+ * that don't declare `slash_commands` get the V0.7-A.0 default `/<id>
+ * prompt:<text> ephemeral:<bool>` (chat handler).
+ *
+ * V0.7-A.1: characters can now declare divergent command sets in their
+ * `character.json`. Eileen's framing: "commands are diff otherwise they'd
+ * be reporting the same shit." E.g. /satoshi (chat) + /satoshi-image
+ * (imagegen handler).
  *
  * Usage:
  *   # guild-only (immediate propagation in a single guild · use during dev):
@@ -25,22 +31,15 @@
  * is the only secret in `.env` strictly required for this script.
  */
 
-import { loadCharacters } from '../src/character-loader.ts';
+import { loadCharacters, resolveSlashCommands } from '../src/character-loader.ts';
+import type { SlashCommandOption, SlashCommandSpec } from '@freeside-characters/persona-engine';
 
 const DISCORD_API_BASE = 'https://discord.com/api/v10';
 
-interface CommandOption {
+interface DiscordCommandSchema {
   name: string;
   description: string;
-  /** Discord application command option type. STRING=3, BOOLEAN=5. */
-  type: number;
-  required?: boolean;
-}
-
-interface CommandSchema {
-  name: string;
-  description: string;
-  options: CommandOption[];
+  options: SlashCommandOption[];
 }
 
 async function main(): Promise<void> {
@@ -66,7 +65,25 @@ async function main(): Promise<void> {
     process.exit(1);
   }
 
-  const commands: CommandSchema[] = characters.map((c) => buildCommand(c.id, c.displayName ?? c.id));
+  // Flatten every character's resolved (declared or defaulted) commands
+  // into one Discord-shaped registration payload. Reject duplicates loud
+  // — Discord silently overwrites by name, but a clash within our own
+  // codebase is a config error worth surfacing rather than papering over.
+  const commands: DiscordCommandSchema[] = [];
+  const ownerByName = new Map<string, string>();
+  for (const c of characters) {
+    for (const spec of resolveSlashCommands(c)) {
+      const existingOwner = ownerByName.get(spec.name);
+      if (existingOwner) {
+        console.error(
+          `publish-commands: duplicate command name /${spec.name} declared by both ${existingOwner} and ${c.id}`,
+        );
+        process.exit(1);
+      }
+      ownerByName.set(spec.name, c.id);
+      commands.push(toDiscordSchema(spec));
+    }
+  }
 
   const url = guildId
     ? `${DISCORD_API_BASE}/applications/${applicationId}/guilds/${guildId}/commands`
@@ -75,7 +92,8 @@ async function main(): Promise<void> {
   const scope = guildId ? `guild ${guildId}` : 'GLOBAL (1-hour propagation)';
   console.log(`publish-commands: registering ${commands.length} commands → ${scope}`);
   for (const cmd of commands) {
-    console.log(`  · /${cmd.name} — ${cmd.description}`);
+    const owner = ownerByName.get(cmd.name) ?? 'unknown';
+    console.log(`  · /${cmd.name} (${owner}) — ${cmd.description}`);
   }
 
   const response = await fetch(url, {
@@ -101,15 +119,11 @@ async function main(): Promise<void> {
   }
 }
 
-function buildCommand(id: string, displayName: string): CommandSchema {
-  const lower = displayName.toLowerCase();
+function toDiscordSchema(spec: SlashCommandSpec): DiscordCommandSchema {
   return {
-    name: id,
-    description: `talk to ${lower}`,
-    options: [
-      { name: 'prompt', description: `what to say to ${lower}`, type: 3, required: true },
-      { name: 'ephemeral', description: 'only you see the reply', type: 5, required: false },
-    ],
+    name: spec.name,
+    description: spec.description,
+    options: spec.options ?? [],
   };
 }
 
--- a/apps/bot/scripts/smoke-interactions.ts
+++ b/apps/bot/scripts/smoke-interactions.ts
@@ -16,7 +16,13 @@ import {
   splitForDiscord,
 } from '@freeside-characters/persona-engine';
 import { startInteractionServer } from '../src/discord-interactions/server.ts';
-import { loadCharacters } from '../src/character-loader.ts';
+import {
+  defaultSlashCommands,
+  loadCharacters,
+  resolveSlashCommandTarget,
+  resolveSlashCommands,
+} from '../src/character-loader.ts';
+import type { CharacterConfig } from '@freeside-characters/persona-engine';
 // V0.7-A.1 imagegen substrate scaffold — smoke at the substrate boundary
 // rather than through the SDK loop (the LLM-driven invocation is exercised
 // end-to-end by the dispatch handler smoke once #2 commands land).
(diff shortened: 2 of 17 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#3. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Imagegen handler bypasses anti-spam guard and circuit breaker",
      "severity": "HIGH",
      "category": "reliability",
      "description": "The new doReplyImagegen path mirrors doReplyChat in shape but the PR description claims it shares 'the 14m30s timeout, the circuit breaker'. Looking at the diff, doReplyImagegen does use TOKEN_LIFETIME_MS via timeoutAfter, but it's unclear whether the circuit breaker / anti-spam guard that wraps doReplyChat (referenced in the docstring) is applied here. If the breaker is in composeReply itself (not at dispatch boundary), imagegen calls have no rate-limiting in front of Bedrock Stability — a single user could spam /satoshi-image and incur unbounded cost. Verify the breaker is invoked at the dispatch boundary (around the Promise.race) for both handlers, not only inside composeReply.",
      "suggestion": "Audit where the circuit breaker is applied. If currently inside composeReply, lift it (or duplicate the gate) around invokeStability so imagegen shares the same per-channel/per-user rate limits. Add a dedicated cost-aware throttle for imagegen if Stability is materially more expensive than chat.",
      "confidence": 0.55
    },
    {
      "id": "F2",
      "title": "Available-commands error message has broken concatenation",
      "severity": "MEDIUM",
      "category": "correctness",
      "description": "The new fallback message at line ~155 reads: `characters.flatMap((c) => (c.slash_commands ?? []).map((s) => s.name).concat(c.slash_commands ? [] : [c.id]))`. When slash_commands is undefined, `(c.slash_commands ?? []).map(...)` returns `[]` and concat yields `[c.id]` — fine. When slash_commands IS defined, it returns the declared names and concat yields `[]` — fine. But this duplicates logic that already exists as resolveSlashCommands(c). The hand-rolled flatMap is harder to reason about and will drift from resolveSlashCommands if the default ever changes.",
      "suggestion": "Replace with `characters.flatMap((c) => resolveSlashCommands(c).map((s) => `/${s.name}`)).join(', ')` to keep a single source of truth.",
      "confidence": 0.9
    },
    {
      "id": "F3",
      "title": "Imagegen ledger entry stores raw URL, may pollute downstream chat context",
      "severity": "MEDIUM",
      "category": "design",
      "description": "doReplyImagegen appends a character ledger entry with content `[imagegen] ${result.url}`. Subsequent chat-mode invocations in the same channel will see this URL in their ledger context. For placeholder URLs (placehold.co) this leaks scaffold detail into character voice context; for real URLs, the LLM may reference the URL string verbatim in chat output, which is semantically incoherent (the URL is delivery metadata, not utterance content). Also, the prompt-as-user-entry stores the raw image prompt as if it were a chat utterance, conflating registers.",
      "suggestion": "Consider a non-content marker like `{role: 'system', content: '[satoshi generated an image]'}` or a dedicated ledger entry kind. At minimum, exclude or sanitize the URL from chat-mode prompt assembly so it isn't pasted into LLM context as utterance text.",
      "confidence": 0.7
    },
    {
      "id": "F4",
      "title": "Two timestamp calls produce slightly different timestamps for paired ledger entries",
      "severity": "LOW",
      "category": "correctness",
      "description": "userEntry uses `nowIso` (computed once), characterEntry uses `new Date().toISOString()` (computed second time). Minor inconsistency — both should likely use `nowIso` or both should be fresh. Trivial but suggests copy-paste vs intent.",
      "suggestion": "Reuse `nowIso` for both, or compute both fresh. Consistency matters for ledger ordering invariants.",
      "confidence": 0.85
    },
    {
      "id": "F5",
      "title": "isImagegenConfigured smoke test passes empty object cast as Config",
      "severity": "LOW",
      "category": "testing",
      "file": "apps/bot/scripts/smoke-interactions.ts",
      "description": "`const emptyConfig = {} as Parameters<typeof isImagegenConfigured>[0]` bypasses zod validation. The real Config object that would reach this code path goes through ConfigSchema.parse() — fields are present but undefined, not absent. The smoke verifies the runtime check works on a partial object, but doesn't exercise the actual zod-parsed shape. Low impact since isImagegenConfigured uses Boolean() coercion which handles both, but the cast pattern is brittle.",
      "suggestion": "Either use a properly-shaped Config object with required fields stubbed, or accept the simplification. Document the rationale if intentional.",
      "confidence": 0.5
    },
    {
      "id": "F6",
      "title": "MCP scoping silently drops unknown server names with no logging",
      "severity": "MEDIUM",
      "category": "observability",
      "description": "buildAllowedTools filters characterMcps against registered server names without logging dropped entries. A character config typo (`mcps: ['scor']`) or an env-gated MCP (codex with CODEX_MCP_URL unset) silently produces a narrower-than-intended scope. The character will fail tool calls without operator-visible reason. Smoke test even validates this drop-silently behavior.",
      "suggestion": "Log dropped names at INFO/WARN: `console.warn(`orchestrator: character ${id} requested mcps [${dropped}] not currently registered — dropped`)`. The smoke can still assert intersection semantics.",
      "confidence": 0.8
    },
    {
      "id": "F7",
      "title": "Discord command name validation not enforced at load/publish time",
      "severity": "MEDIUM",
      "category": "validation",
      "file": "apps/bot/src/character-loader.ts",
      "description": "SlashCommandSpec docstring says name must match `/^[-_\\p{L}\\p{N}]{1,32}$/`, description says 1-100 chars, options 0-25. Neither character-loader.ts nor publish-commands.ts validates these constraints — invalid configs surface only as 400 errors from the Discord API at publish time. The publish script does check for duplicates loud, which is good, but other constraints are not enforced. Default `c.id` is also not validated against the regex; an id with uppercase or spaces would silently produce an invalid command.",
      "suggestion": "Validate at character load (zod schema for SlashCommandSpec) so misconfigured character.json fails fast with a clear pointer rather than a Discord 400.",
      "confidence": 0.85
    },
    {
      "id": "F8",
      "title": "publish-commands.ts owner lookup uses Map but not defensively",
      "severity": "LOW",
      "category": "code-quality",
      "file": "apps/bot/scripts/publish-commands.ts",
      "description": "The duplicate-detection logic uses `ownerByName.get(spec.name)` which returns string|undefined. The `existingOwner` truthy check is fine for non-empty strings, but the subsequent log line `ownerByName.get(cmd.name) ?? 'unknown'` — owner should always be present since we just populated it. Defensive but suggests uncertainty about invariants.",
      "suggestion": "Either trust the invariant (since we just populated ownerByName for every cmd) or assert it. Minor.",
      "confidence": 0.7
    },
    {
      "id": "F9",
      "title": "Placeholder URL leaks user prompt to placehold.co",
      "severity": "LOW",
      "category": "privacy",
      "file": "packages/persona-engine/src/orchestrator/imagegen/bedrock-client.ts",
      "description": "buildPlaceholderUrl encodes the (truncated) user prompt as a query parameter to placehold.co. In scaffold mode, every /satoshi-image invocation results in a real HTTP fetch (Discord rendering the embed) to placehold.co with user content in the URL. placehold.co logs may capture these. Low impact while in dev/scaffold, but worth flagging before it ships to production users.",
      "suggestion": "Either omit the prompt from the URL (only seed and synthetic id), or guard scaffold mode behind a non-production env flag so user prompts never reach a third-party logging surface unintentionally.",
      "confidence": 0.6
    },
    {
      "id": "F10",
      "title": "Imagegen handler does not validate prompt length or content",
      "severity": "LOW",
      "category": "input-validation",
      "description": "doReplyImagegen passes `prompt` straight to invokeStability with only a `.trim()` upstream. Bedrock Stability has prompt length limits and content policy that should be checked early to give friendly errors rather than opaque AWS failures. The MCP tool surface uses zod `z.string().min(1)` but the direct dispatch handler bypasses that.",
      "suggestion": "Add a length check (e.g. > 1, < 2000) and possibly a basic prefix-blocklist in invokeStability or the dispatch handler.",
      "confidence": 0.65
    },
    {
      "id": "F11",
      "title": "Per-character MCP scoping documentation accurate; smoke covers intent ∩ registered semantics",
      "severity": "PRAISE",
      "category": "design",
      "description": "The buildAllowedTools API change is minimal (one optional parameter), preserves V0.6 behavior when omitted, and the documentation captures the 'intent ∩ registered' semantics. Smoke test covers undefined-passthrough, partial filter, empty intersection, and unknown-name drop. Clean and well-tested boundary change.",
      "suggestion": "Pair this with F6 (logging) for full operability."
    },
    {
      "id": "F12",
      "title": "Substrate scaffold pattern: type-shaped boundary lets dispatch land before Bedrock body",
      "severity": "PRAISE",
      "category": "architecture",
      "file": "packages/persona-engine/src/orchestrator/imagegen/bedrock-client.ts",
      "description": "The placeholder body that round-trips prompt/seed into a synthetic URL with placeholder=true is a sound substrate scaffolding choice — it allows MCP tool surface, dispatch handler, and smoke tests to land + run end-to-end before the real SDK call exists. Placeholder flag flowing through the type and being decorated in the reply means consumers visibly distinguish scaffold from real."
    },
    {
      "id": "F13",
      "title": "resolveSlashCommandTarget is O(N*M) but adequately scoped",
      "severity": "LOW",
      "category": "performance",
      "file": "apps/bot/src/character-loader.ts",
      "description": "resolveSlashCommandTarget iterates all characters and all their commands per dispatch. With current N≈2 and M≈2 this is trivial; if the character fleet grows, a name → {character, spec} Map built once at load time would be cheaper. Not currently a problem.",
      "suggestion": "Defer optimization until needed; possibly memoize the lookup map at module init when fleet grows.",
      "confidence": 0.9
    },
    {
      "id": "F14",
      "title": "ARCHETYPE_STYLE table mixes capitalized canonical names with lowercase fallbacks",
      "severity": "LOW",
      "category": "code-quality",
      "file": "packages/persona-engine/src/orchestrator/imagegen/server.ts",
      "description": "The static map has both 'Storyteller' and 'storyteller' (and only some lowercased). suggestStyle does `ARCHETYPE_STYLE[archetype] ?? ARCHETYPE_STYLE[archetype.trim()]` — note the second lookup is identical to the first if there's no leading whitespace. Should probably be `archetype.toLowerCase()` if the intent is case-insensitive matching, otherwise the partial lowercase entries are arbitrary.",
      "suggestion": "Either normalize the lookup with `archetype.toLowerCase()` and store only lowercase keys, or remove the redundant second lookup. Current shape implies case-insensitive intent that isn't actually implemented.",
      "confidence": 0.85
    },
    {
      "id": "F15",
      "title": "deterministicSeed uses signed-32-bit math; Math.abs(INT_MIN) edge case",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/orchestrator/imagegen/bedrock-client.ts",
      "description": "The DJB2-style hash uses `| 0` to coerce to int32, then `Math.abs(h)`. If h equals -2147483648 (INT32_MIN), Math.abs returns -2147483648 (since it has no positive int32 representation). Extremely unlikely to hit but means seed is technically nonnegative-violating. Bedrock seed param will likely accept it as int but if downstream uses z.number().int().nonnegative() this would fail.",
      "suggestion": "Use `(h >>> 0)` to get unsigned 32-bit instead of `Math.abs(h | 0)`.",
      "confidence": 0.95
    },
    {
      "id": "F16",
      "title": "SlashCommandSpec.options.type is plain `number`, not enum",
      "severity": "LOW",
      "category": "type-safety",
      "file": "packages/persona-engine/src/types.ts",
      "description": "The `type` field is typed as `number` with a doc comment listing valid values (3=STRING, 4=INTEGER, 5=BOOLEAN, 10=NUMBER). Using a numeric union or named enum would catch typos at compile time. Currently `type: 99` would type-check but fail at Discord publish.",
      "suggestion": "Define `type ApplicationCommandOptionType = 3 | 4 | 5 | 10` (or full Discord enum) and use it in the field type.",
      "confidence": 0.9
    },
    {
      "id": "F17",
      "title": "Mixed naming convention: snake_case slash_commands in TS interface",
      "severity": "LOW",
      "category": "code-quality",
      "file": "packages/persona-engine/src/types.ts",
      "description": "CharacterConfig uses camelCase throughout (displayName, personaPath, anchoredArchetypes, emojiAffinity, webhookAvatarUrl) but introduces `slash_commands` and `mcps` in snake_case/lowercase. This was likely chosen to mirror the JSON shape, but mixed conventions in a single TS interface are jarring and may cause readability friction.",
      "suggestion": "Either rename to slashCommands or document the convention shift (mirroring Discord-side JSON keys, perhaps). At minimum add a comment explaining why snake_case is used here.",
      "confidence": 0.6
    },
    {
      "id": "F18",
      "title": "buildPlaceholderUrl truncates prompt at 80 chars but smoke test asserts on substrings",
      "severity": "LOW",
      "category": "testing",
      "file": "apps/bot/scripts/smoke-interactions.ts",
      "description": "The smoke checks `stub.url.startsWith('https://placehold.co/imagegen?')` which is fine, but doesn't verify the seed round-trip or the prompt encoding. Light coverage for the placeholder URL contract; if someone changes the URL shape, only the prefix test will catch it.",
      "suggestion": "Add an assertion that `stub.url` contains `seed=${stub.seed}` and `prompt=` segments to lock the round-trip contract."
    }
  ]
}
```


## Callouts

PR #4, opened 2026-05-02
feat(score): SCORE_BEARER for routing via mcp.0xhoneyjar.xyz/score

zkSoju:
## Summary

Adds `SCORE_BEARER` env var (additive, optional). When set, score-mcp calls include `Authorization: Bearer ${SCORE_BEARER}` alongside the existing `X-MCP-Key`, enabling score traffic to route via the freeside-mcp-gateway at `mcp.0xhoneyjar.xyz/score` while preserving direct-route behavior when unset.

## Doctrine

Per **gateway-as-registry** (`~/vault/wiki/concepts/gateway-as-registry.md`, 2026-05-01): the gateway gates its own routing surface; the upstream announces its own auth requirement; the caller composes from both declarations.

Two distinct auth surfaces, two distinct keys, one composed request:

```
POST mcp.0xhoneyjar.xyz/score/mcp
Authorization: Bearer ${SCORE_BEARER}    ← gateway gate (registry policy)
X-MCP-Key: ${MCP_KEY}                    ← upstream auth (broadcast)
```

The gateway holds zero upstream secrets — caller composes the request from gateway's gate declaration + upstream's broadcast.

## Files

| file | change |
| --- | --- |
| `packages/persona-engine/src/config.ts` | `SCORE_BEARER` (additive, optional) + path-comment header |
| `packages/persona-engine/src/orchestrator/index.ts` | `Authorization` header alongside `X-MCP-Key` when `SCORE_BEARER` set |
| `packages/persona-engine/src/score/client.ts` | thread `bearer?` through `mcpInit` / `mcpToolCall`; new `authHeaders()` helper |

## Test plan

- [x] `bun run typecheck` clean (persona-engine + bot)
- [ ] post-deploy: set `SCORE_BEARER` + flip `SCORE_API_URL=https://mcp.0xhoneyjar.xyz/score` on Railway
- [ ] post-deploy: ruggy/satoshi narrative still grounded (next digest)
- [ ] backward-compat: `SCORE_BEARER` unset → V0.5+ direct-route behavior preserved (zero behavior change)

## Operator runtime ops (post-merge)

| step | where |
| --- | --- |
| set `SCORE_BEARER` on freeside-characters Railway | matches gateway's `TENANT_SCORE_API_KEY` (companion PR: 0xHoneyJar/freeside-mcp-gateway#2) |
| flip `SCORE_API_URL` on freeside-characters Railway | `https://score-api-production.up.railway.app` → `https://mcp.0xhoneyjar.xyz/score` |
| deploy | Railway auto |
| verify | next ruggy/satoshi digest stays grounded; gateway logs show `/score/mcp` traffic |

## Companion PR

- 0xHoneyJar/freeside-mcp-gateway#2 — gateway side (visibility filter · access gate · score tenant declaration)

### DIFF
--- a/packages/persona-engine/src/config.ts
+++ b/packages/persona-engine/src/config.ts
@@ -14,9 +14,17 @@ const ConfigSchema = z.object({
   LLM_PROVIDER: z.enum(['stub', 'anthropic', 'freeside', 'bedrock', 'auto']).default('auto'),
 
   // ─── score-mcp (zerker — production data path) ────────────────────────
+  // Direct path (V0.5-): SCORE_API_URL=https://score-api-production.up.railway.app, MCP_KEY set, SCORE_BEARER unset.
+  // Gateway path (V0.7+): SCORE_API_URL=https://mcp.0xhoneyjar.xyz/score, MCP_KEY + SCORE_BEARER both set.
+  // The gateway is registry-shaped: it declares each upstream's auth via the federation manifest;
+  // callers compose the request from the declaration. SCORE_BEARER passes the gateway gate;
+  // X-MCP-Key (sourced from MCP_KEY) satisfies the upstream's announced auth.
   SCORE_API_URL: z.string().url().default('https://score-api-production.up.railway.app'),
   SCORE_API_KEY: z.string().optional(),
   MCP_KEY: z.string().optional(),
+  /** Bearer for the freeside-mcp-gateway gate. Matches the gateway's
+   *  `TENANT_SCORE_API_KEY` env. Unset = direct route (no gateway gate). */
+  SCORE_BEARER: z.string().optional(),
 
   // ─── codex-mcp (gumi — mibera-codex lookup, public, no auth) ──────────
   /**
--- a/packages/persona-engine/src/orchestrator/index.ts
+++ b/packages/persona-engine/src/orchestrator/index.ts
@@ -86,7 +86,13 @@ function buildMcpServers(config: Config): Record<string, McpServerConfig> {
     servers.score = {
       type: 'http',
       url: `${config.SCORE_API_URL}/mcp`,
-      headers: { 'X-MCP-Key': config.MCP_KEY },
+      headers: {
+        'X-MCP-Key': config.MCP_KEY,
+        // SCORE_BEARER is the gateway gate; harmless when score is reached
+        // direct (upstream ignores Authorization). Set both env vars to
+        // route via mcp.0xhoneyjar.xyz/score.
+        ...(config.SCORE_BEARER ? { Authorization: `Bearer ${config.SCORE_BEARER}` } : {}),
+      },
     };
   }
 
--- a/packages/persona-engine/src/score/client.ts
+++ b/packages/persona-engine/src/score/client.ts
@@ -50,13 +50,20 @@ function parseSseEnvelope<T>(body: string): McpJsonRpcEnvelope<T> {
   return JSON.parse(json) as McpJsonRpcEnvelope<T>;
 }
 
-async function mcpInit(url: string, key: string): Promise<McpInitResult> {
+function authHeaders(key: string, bearer?: string): Record<string, string> {
+  return {
+    'X-MCP-Key': key,
+    ...(bearer ? { Authorization: `Bearer ${bearer}` } : {}),
+  };
+}
+
+async function mcpInit(url: string, key: string, bearer?: string): Promise<McpInitResult> {
   const response = await fetch(url, {
     method: 'POST',
     headers: {
       'Content-Type': 'application/json',
       Accept: 'application/json, text/event-stream',
-      'X-MCP-Key': key,
+      ...authHeaders(key, bearer),
     },
     body: JSON.stringify({
       jsonrpc: '2.0',
@@ -88,7 +95,7 @@ async function mcpInit(url: string, key: string): Promise<McpInitResult> {
     headers: {
       'Content-Type': 'application/json',
       Accept: 'application/json, text/event-stream',
-      'X-MCP-Key': key,
+      ...authHeaders(key, bearer),
       'Mcp-Session-Id': sessionId,
     },
     body: JSON.stringify({ jsonrpc: '2.0', method: 'notifications/initialized' }),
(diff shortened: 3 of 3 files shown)


zkSoju (commented):
## Summary

This PR adds a clean routing path through the MCP gateway via `SCORE_BEARER`, with a thoughtful `authHeaders` helper and operator-facing config docs. The primary opportunities are tightening config validation at startup and de-duplicating auth header logic between the orchestrator and client.

The architecture here echoes a pattern Stripe formalized in their internal RPC fabric: when the same service can be reached via two paths (direct and gatewayed), the *configuration* itself becomes the contract. Misconfiguration in such systems doesn't fail loudly — it fails subtly, often in a way that only manifests under a specific tenant or region. The discipline of validating that contract at boot time is what separates "it works on my laptop" from "it works in every region for every tenant."

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "praise-auth-helper",
      "title": "Clean header composition via authHeaders helper",
      "severity": "PRAISE",
      "category": "code-quality",
      "description": "Consolidating auth header construction into a single `authHeaders` helper avoids drift between `mcpInit`, the initialized notification, and `mcpToolCall`. The optional bearer is cleanly conditional via spread.",
      "confidence": 0.9,
      "faang_parallel": "Google's internal Stubby/gRPC clients converged on a single `CallCredentials` abstraction precisely because three call sites with three slightly different header builders is a bug factory. One source of truth for credentials.",
      "metaphor": "Like a hotel with one keycard reader at every door instead of three different lock types — guests can't get confused, and security audits take minutes instead of days.",
      "teachable_moment": "When you see the same 2-3 lines of header/auth construction repeated, that's a latent inconsistency waiting to happen during the next auth change."
    },
    {
      "id": "praise-config-docs",
      "title": "Config schema includes operator-facing routing docs",
      "severity": "PRAISE",
      "category": "documentation",
      "file": "packages/persona-engine/src/config.ts",
      "description": "The inline comments explicitly describe the V0.5/V0.7 routing matrix (which env vars to set for direct vs gateway path) and the relationship between SCORE_BEARER and the gateway's TENANT_SCORE_API_KEY.",
      "confidence": 0.9,
      "faang_parallel": "Amazon's internal config schemas (used in Apollo/AWS deployment tooling) embed routing intent inline because tribal knowledge about 'which env var goes with which' is the #1 source of pager incidents during on-call handoffs.",
      "metaphor": "These comments are the wiring diagram taped inside the breaker box — the next electrician doesn't have to guess which switch controls the kitchen."
    },
    {
      "id": "no-validation-bearer-without-key",
      "title": "No validation that gateway routing has both MCP_KEY and SCORE_BEARER",
      "severity": "LOW",
      "category": "configuration",
      "file": "packages/persona-engine/src/config.ts",
      "description": "The schema documents that gateway path requires both MCP_KEY and SCORE_BEARER, but nothing enforces this. A misconfiguration where SCORE_API_URL points to mcp.0xhoneyjar.xyz/score while SCORE_BEARER is unset will silently fail at request time. Similarly, SCORE_BEARER set without MCP_KEY would skip the score path entirely.",
      "suggestion": "Add `.superRefine()` to the zod schema: when SCORE_API_URL host matches the gateway, require both MCP_KEY and SCORE_BEARER. At minimum, emit a startup warning.",
      "confidence": 0.6,
      "faang_parallel": "Netflix's Archaius and Meta's configerator both fail-fast on cross-field invariants at boot. The principle: configuration errors should surface at deploy time, not at the first user request — ideally before traffic shifts.",
      "metaphor": "It's like a car that lets you start the engine without a steering wheel installed — you only discover the problem at the first turn.",
      "teachable_moment": "Cross-field config invariants are the cheapest bugs to catch at boot and the most expensive to debug in production. zod's `superRefine` is built for exactly this."
    },
    {
      "id": "orchestrator-comment-accuracy",
      "title": "Comment claim 'upstream ignores Authorization' is environment-dependent",
      "severity": "LOW",
      "category": "documentation",
      "description": "The comment says SCORE_BEARER is 'harmless when score is reached direct (upstream ignores Authorization)'. If upstream score-api ever begins validating Authorization (e.g., multi-tenant), an accidentally-set SCORE_BEARER pointing at the direct URL could cause auth failures rather than being harmless.",
      "suggestion": "Qualify the comment ('current upstream ignores Authorization as of <date>') OR have `buildMcpServers` omit Authorization when URL host is not the gateway. The latter is more defensive.",
      "confidence": 0.5,
      "faang_parallel": "Hyrum's Law in action: any observable behavior of your dependencies will eventually be relied upon. The mirror is also true — any *unobservable* tolerance (like ignored headers) will eventually become strict.",
      "metaphor": "Comments that assert remote behavior are weather forecasts written in permanent marker."
    },
    {
      "id": "duplicated-auth-logic-orchestrator-vs-client",
      "title": "Auth header composition duplicated between client.ts and orchestrator/index.ts",
      "severity": "LOW",
      "category": "code-quality",
      "description": "The orchestrator inlines the same `X-MCP-Key` + optional `Authorization: Bearer` pattern that client.ts now centralizes in `authHeaders`. If a third auth header is ever added, both call sites must change.",
      "suggestion": "Export `authHeaders` from client.ts (or a small shared `auth.ts` module) and reuse it in `buildMcpServers`.",
      "confidence": 0.7,
      "metaphor": "You just installed one keycard reader on the front door — but the side entrance still has the old lock. Time to rekey both."
    },
    {
      "id": "bearer-not-redacted-in-logs",
      "title": "No evidence SCORE_BEARER is treated as a secret in logging/errors",
      "severity": "SPECULATION",
      "category": "security",
      "description": "SCORE_BEARER is a credential. The diff doesn't show logging changes, but if existing error paths in mcpInit/mcpToolCall log request headers or echo config, the bearer could leak.",
      "suggestion": "Audit logger/error reporter redaction; ensure SCORE_BEARER joins MCP_KEY/SCORE_API_KEY in the redaction list.",
      "confidence": 0.3,
      "faang_parallel": "GitHub's 2018 incident — internal logs captured plaintext credentials because a new auth field wasn't added to the redaction allowlist. The lesson: redaction lists must be updated at the same commit that adds the secret."
    }
  ]
}
```


## Callouts

The `authHeaders` helper and the routing-matrix comments in `config.ts` are the kind of small choices that compound. Centralized credential composition prevents three-way drift; embedded routing docs prevent on-call confusion. Together they reduce the cognitive load on the next operator who has to reason about which env vars belong to which path — exactly the operability mindset that makes systems durable across team handoffs.

PR #8, opened 2026-05-02
feat(V0.7-A.2): compose unification + chat-tool streaming + WITNESS first-run

zkSoju:
<details open>
<summary>combined cycle: streaming + compose unification + WITNESS first-run</summary>

```mermaid
graph LR
  A[PR #8 streaming<br/>onToolUse callback] --> M[combined cycle]
  B[V0.7-A.2 unification<br/>'reply' PostType + buildPrompt + compose] --> M
  C[WITNESS first-run<br/>QA checklist generation] --> M
  M --> Q[🟡 operator-bounded<br/>dev-guild QA + Eileen + gumi]
  classDef ship fill:#1f6f3a,stroke:#0f3d20,color:#fff
  classDef gate fill:#7a5b14,stroke:#3d2d0a,color:#fff
  class A,B,C ship
  class M,Q gate
```
</details>

three bundles shipped together for one QA + bridgebuilder pass. closes the operator's "chat and digest should be the same thing" reframe at the API surface; voice fidelity preserved (per-PostType fragments untouched).

🟢 shipped & smoke-verified · 🟡 operator-bounded · 🔴 deferred to V0.7-A.3+

| bundle | what landed | gate |
|---|---|---|
| 🛠 PR #8 streaming | orchestrator streams `assistant.tool_use`; `onToolUse` callback fires per call; dispatcher PATCHes deferred message with "📊 pulling X…" progress; final via PATCH or webhook | 🟡 dev-guild |
| 🎼 V0.7-A.2 unification | 'reply' promoted to PostType (7th); `CONVERSATION_MODE_OVERRIDE` lifted to per-character persona.md fragment (byte-identical); `buildPrompt(shape)` discriminated unifier; `compose(invocation)` entrypoint | 🟡 voice fidelity |
| 🔬 WITNESS first-run | hand-rolled QA checklist at `grimoires/loa/qa/qa-cycle-V07A1+A2-2026-05-02.md` covering 7 real-interaction surfaces with COMMAND/EXPECTED/CAPTURE/TRIAGE | 🟡 dogfood feedback |

**land it** (the WITNESS checklist is the operator-facing gate · run in dev guild):
- 🟡 Surface 1 — `/ruggy zone digest stonehenge` · expect `📊 pulling…` + real digest data, no JSON leak
- 🟡 Surface 2 — `/satoshi grail of crossings` · expect `🏆 checking grail…` + Grail #4488 cite from codex
- 🟡 Surface 5 — stub digest regression · output should be byte-identical to pre-V0.7-A.2 baseline
- 🟡 Surface 7 — gumi blind-judge strip-the-name (≥80% per character)
- 🟡 Eileen async review of fragment lift (`<!-- @FRAGMENT: reply -->` byte-identical to historical const)

**STOP merge if**: tool-call JSON still leaks (PR #8 didn't fix) · digest voice regresses · gumi <80%

→ kickoff brief: `grimoires/loa/specs/build-compose-unification-v07a2.md`
→ WITNESS proposal: `grimoires/loa/proposals/qa-real-interaction-construct.md`
→ pattern reference: [ruggy-v2 onToolUse](https://github.com/0xHoneyJar/ruggy-v2/blob/main/src/agent.ts) + [moltbot thinking-frame](https://github.com/0xHoneyJar/ruggy-moltbot/blob/main/src/handlers/chat.ts)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#8. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "buildPromptPair silently coerces 'reply' PostType to 'digest'",
      "severity": "MEDIUM",
      "category": "correctness",
      "description": "The backward-compat shim accepts PostType='reply' (now valid in the union) and silently coerces to 'digest' rather than throwing or warning. A misrouted call (e.g., a future caller passing 'reply' to the cron path) would silently produce a digest-shaped prompt with no signal to the developer. The comment acknowledges 'type discipline is the right long-term fix' but the runtime behavior masks bugs in the meantime.",
      "suggestion": "Either (a) throw with a clear error 'buildPromptPair received reply postType — use buildReplyPromptPair or buildPrompt({kind: reply}) instead', or (b) log a warning + use the CronPostType narrowing on the args.postType field directly (TypeScript would reject 'reply' at call sites). Given CronPostType already exists in post-types.ts, narrowing args.postType to CronPostType is the cleanest fix and could land in V0.7-A.2.",
      "confidence": 0.7
    },
    {
      "id": "F2",
      "title": "postTypeFitsData('reply', digest) returns true unconditionally",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/compose/post-types.ts",
      "description": "The new 'reply' case in postTypeFitsData returns true with a comment that 'reply' is on-demand and shouldn't reach this function. Returning true defensively means a buggy cron scheduler that mis-classifies a slot as 'reply' would proceed to compose+broadcast a digest using the reply fragment, instead of being filtered out. Returning false would be the safer fail-closed default.",
      "suggestion": "Return false for 'reply' in postTypeFitsData — cron paths that ask 'does this fit?' should never green-light an on-demand-only PostType. Or throw to make the misuse loud.",
      "confidence": 0.55
    },
    {
      "id": "F3",
      "title": "ToolUseEvent input field carries opaque LLM-supplied data into onToolUse callback",
      "severity": "LOW",
      "category": "security",
      "description": "ToolUseEvent.input is typed `unknown` and passed to caller-supplied callbacks without size or shape validation. The dispatcher's onToolUse only reads event.name + event.id, which is fine, but downstream consumers (trajectory loggers, future PATCH formatters) might serialize event.input and expose it. Tool args from the LLM could be arbitrarily large or contain user-controlled content (from chat prompts) that gets logged or rendered.",
      "suggestion": "Document explicitly that event.input is untrusted LLM output. Consider redacting/truncating in trajectory logs. Not blocking for V0.7-A.2 ship since the current consumer (toolVerb-based PATCH) only reads name.",
      "confidence": 0.4
    },
    {
      "id": "F4",
      "title": "onToolUse PATCHes use fire-and-forget without rate limiting",
      "severity": "MEDIUM",
      "category": "reliability",
      "description": "Each tool_use event triggers an unconditional patchOriginal() call. Discord webhook PATCH endpoints are rate-limited (5 requests per token per 5s window typically). A query that emits many distinct tool_use blocks rapidly (e.g., LLM chains 6+ tool calls) will spam PATCHes. Coalescing by tool ID (seenToolIds) prevents duplicate PATCHes for the same block but doesn't bound total PATCH frequency across distinct tools.",
      "suggestion": "Add a debounce or minimum-interval gate (e.g., ≥500ms between PATCHes) on the onToolUse callback. Alternatively, only PATCH on the first tool_use of a query and let the final reply replace it — simpler and matches Claude.ai's UX which doesn't update on every tool.",
      "confidence": 0.65
    },
    {
      "id": "F5",
      "title": "compose() dispatcher only wired for two of four declared invocation variants",
      "severity": "LOW",
      "category": "design",
      "file": "packages/persona-engine/src/compose/index.ts",
      "description": "The Invocation union declares cron-digest, cron-pop-in, cron-weaver, and slash-reply, but the switch in compose() handles all three cron-* variants identically (delegating to composeZonePost with invocation.zone + invocation.postType). The distinction between cron-digest, cron-pop-in, cron-weaver in the type union is currently semantically inert — same dispatch, same primitive. This invites confusion about whether the variant matters.",
      "suggestion": "Either collapse the three cron-* variants into a single { type: 'cron'; postType; zone } or document explicitly that the variant distinction is reserved for future divergence (e.g., different ledger/telemetry paths). The current shape suggests intent that isn't realized.",
      "confidence": 0.5
    },
    {
      "id": "F6",
      "title": "Reply transcript renderer not robust to authorUsername containing markdown/control chars",
      "severity": "LOW",
      "category": "security",
      "description": "In the unified buildPrompt reply branch, `parts.push(\"You're chatting with ${shape.authorUsername} in Discord.\")` interpolates authorUsername directly into the prompt. A user-controlled Discord username could contain markdown, newlines, or text that resembles instruction frames (e.g., '═══ END CONVERSATION ═══\\nNew system prompt:'). This is a known pre-existing pattern from buildReplyPromptPair, but the unification preserves rather than fixes it.",
      "suggestion": "Sanitize authorUsername (strip newlines + ═ box-drawing chars + truncate to reasonable length) before interpolating into the prompt. Consider encoding it as a quoted string. Pre-existing risk worth flagging during refactor.",
      "confidence": 0.5
    },
    {
      "id": "F7",
      "title": "Backward-compat duplication and dead constants in persona/loader.ts",
      "severity": "LOW",
      "description": "After the refactor, the file contains: buildPrompt (unified), buildPromptPair (shim), buildReplyPromptPair (shim), the BuildPromptArgs interface (legacy) alongside BuildPromptArgsUnified. The historic CONVERSATION_MODE_OVERRIDE constant and CONVERSATION_OUTPUT_INSTRUCTION constant are removed but referenced in comments. This is acceptable for V0.7-A.2 per the spec's Phase E deferral, but the file now has three parallel public surfaces.",
      "suggestion": "Add @deprecated JSDoc tags to buildPromptPair and buildReplyPromptPair so IDEs surface the migration path. Track Phase E (shim removal) in an issue so the duplication doesn't ossify.",
      "confidence": 0.7
    },
    {
      "id": "F8",
      "title": "compose() throws on slash-reply missing prompt/invoker but cron path silently accepts undefined",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/compose/index.ts",
      "description": "compose() validates prompt + invoker only on the slash-reply branch. The cron branch ignores args.prompt entirely (composeZonePost fetches its own ZoneDigest). If a caller passes incorrect args (e.g., cron with prompt set), there's no signal. ComposeArgs.prompt is optional (?) but the slash-reply path requires it at runtime — the type doesn't reflect the variant-dependent requirement.",
      "suggestion": "Move prompt/invoker into the slash-reply Invocation variant itself, so TypeScript enforces the requirement at the call site. The Invocation discriminator is the right place for variant-specific required fields.",
      "confidence": 0.7
    },
    {
      "id": "F9",
      "title": "Spec, proposal, and QA docs are extensive but not part of the runtime change",
      "severity": "PRAISE",
      "category": "process",
      "file": "grimoires/loa/specs/build-compose-unification-v07a2.md",
      "description": "The kickoff spec articulates invariants, blast radius, phase scoping, and explicit Barth-style scope cuts. The companion WITNESS proposal + first-run QA checklist document operator-bounded validation surfaces with triage paths. This is unusually thorough planning documentation that pairs intent with the diff.",
      "suggestion": "No change. Worth noting that the volume of docs (800+ lines) outweighs the code change (~400 lines net) — reviewers should be aware that most diff bytes are planning material, not behavior.",
      "confidence": 0.9
    },
    {
      "id": "F10",
      "title": "Tool-verb mapping is hardcoded and silently falls back for unknown tools",
      "severity": "LOW",
      "file": "apps/bot/src/discord-interactions/tool-verb.ts",
      "description": "TOOL_VERBS hardcodes ~20 tool names. New MCP tools (or renamed existing ones) won't have human-friendly verbs and will fall through to the generic '🔧 invoking server::tool' pattern. There's no test asserting the mapping stays in sync with allowedTools or the actual MCP server tool catalogs. A typo in a tool name (e.g., mcp__codex__lookup_grail vs mcp__codex__get_grail) would silently produce the fallback without any signal.",
      "suggestion": "Add a CI check or test that loads the MCP servers' tool catalogs and asserts every callable tool name has an entry in TOOL_VERBS (or is in an explicit ignore list). Alternatively, log a warning the first time fallback is hit per process.",
      "confidence": 0.5
    },
    {
      "id": "F11",
      "title": "Streaming refactor relies on assistant.message.content shape not asserted",
      "severity": "LOW",
      "category": "correctness",
      "description": "The new tool_use extraction iterates `message.message.content` and matches on `block.type === 'tool_use'`, then accesses block.name, block.id, block.input. This assumes the SDK message shape — if the SDK ships a minor version that changes the content block schema, the loop silently produces no events (no toolUses appended, no warning). Given onToolUse is a behavioral feature, silent regression to non-progressive UX would be hard to notice.",
      "suggestion": "Add a smoke or contract test that exercises the orchestrator with a mocked SDK stream containing assistant+tool_use to assert toolUses populates. Or log when an unexpected block.type is encountered to surface schema drift.",
      "confidence": 0.55
    },
    {
      "id": "F12",
      "title": "Comment on toolUses ordering contains contradictory phrasing",
      "severity": "LOW",
      "category": "documentation",
      "description": "The OrchestratorResponse field comment reads: 'Tool calls observed during the run (most recent first never — append order).' The phrasing 'most recent first never' is grammatically broken and confusing for readers trying to understand the ordering guarantee.",
      "suggestion": "Reword to: 'Tool calls observed during the run, in stream order (oldest first).'",
      "confidence": 0.95
    },
    {
      "id": "F13",
      "title": "Unified buildPrompt double-builds movementGuidance for reply path",
      "severity": "LOW",
      "category": "code quality",
      "description": "The unified buildPrompt computes movementGuidance once, then the reply branch substitutes it into the system half. But for reply mode the constant is '(not applicable in conversation mode)' — substituting it into placeholders that may exist in voice/codex anchor blocks is fine, but the user-half rendering for reply mode skips substitution entirely. If a future reply fragment author embeds {{MOVEMENT_GUIDANCE}}, the literal would leak. (Pre-existing risk inherited from the prior reply builder.)",
      "suggestion": "For consistency, run the substitution chain over the reply user-half too, or document explicitly which placeholders are guaranteed in each mode.",
      "confidence": 0.4
    },
    {
      "id": "F14",
      "title": "Deferred shim cleanup is the right tradeoff",
      "severity": "PRAISE",
      "category": "design",
      "description": "The decision to keep buildPromptPair and buildReplyPromptPair as thin delegating shims rather than migrating all call sites in one PR keeps the blast radius narrow. The unified buildPrompt is the new canonical entry; consumers can migrate at their own pace. This is a sound application of strangler-fig refactoring.",
      "suggestion": "No change. The shim layer is well-justified.",
      "confidence": 0.85
    },
    {
      "id": "F15",
      "title": "Fragment lift content is byte-identical between ruggy and satoshi personas",
      "severity": "MEDIUM",
      "category": "design",
      "description": "Both ruggy and satoshi persona.md files received the same identical 49-line `<!-- @FRAGMENT: reply -->` block. The spec frames this as voice-tuning preservation (same content, different per-character context wraps it). However, having the same fragment text duplicated in every character file means future edits must be made in N places. The previous CONVERSATION_MODE_OVERRIDE constant was a single source. The lift to per-character fragments anticipates per-character variation that doesn't yet exist.",
      "suggestion": "Either (a) accept the duplication knowing voice-tuning will diverge per character (consistent with the spec's intent), or (b) keep a shared constant + load it as a default fragment when persona.md doesn't override. Worth an explicit decision rather than implicit duplication.",
      "confidence": 0.6
    }
  ]
}
```


## Callouts

zkSoju:
## Post-Merge Pipeline Results

| Phase | Status | Details |
|-------|--------|--------|
| classify | ✅ completed | cycle |
| semver | ✅ completed | 0.9.1 → 0.10.0 (minor) |
| changelog | ✅ completed |  |
| gt_regen | ❌ failed |  |
| rtfm | ✅ completed | 1 gap(s) |
| tag | ✅ completed | v0.10.0 |
| release | ✅ completed |  |
| lore_promote | ⊘ skipped | disabled in config |
| notify | ⏳ in_progress |  |

_Started: 2026-05-02T03:50:24Z_

PR #18, opened 2026-05-02
feat(substrate): Effect.Schema MCP contracts · emojis migrated · 3 test classes

zkSoju:
## Summary

Substrate strengthening per kickoff seed (`~/bonfire/grimoires/bonfire/specs/freeside-mcp-organization-kickoff-2026-05-01.md`). Closes the broken-chat regression class via contract discipline + drift detection. **Worked example**: emojis MCP migrated end-to-end as the proof-of-pattern that scales to rosenzu/freeside_auth/imagegen in follow-up sessions.

- **Effect.Schema substrate** (`_schema/`): `McpToolContract` + `McpServerContract` types, decode helpers, `assertZodParity<T>()` for compile-time Effect↔Zod structural parity. Aligned with `freeside-mcp-gateway` (`effect ^3.21.2` single package per Effect 3.x consolidation).
- **emojis migration** (`emojis/{schema,server,server.test}.ts`): contracts are the single source of truth; `server.ts` derives all 6 SDK `tool()` registrations from `schema.ts`; mood/kind literal tuples parity-checked against `registry.ts`.
- **3 test classes (49 tests · 0 fail · 113 expect() calls)**:
  - `emojis/server.test.ts` (24): per-MCP boundary contract — input round-trip, invalid-input rejection, output round-trip
  - `apps/bot/src/tests/surface-completeness.test.ts` (16): every `character.mcps` name resolves under full env; CHAT_MODE × LLM_PROVIDER matrix locked
  - `apps/bot/src/tests/persona-tool-drift.test.ts` (9): every `mcp__<server>__<tool>` reference in persona.md maps to a registered tool a character is allowed to call
- **`docs/MCP-FEDERATION.md`**: 3-flavor strategy (substrate-internal · federated tenant · per-character scope), 4-axis ownership table, promotion path, capability-request flow.

Cross-repo drafts (in `grimoires/loa/proposals/`, awaits operator coordination push per seed §9.2-9.4):
- `registering-mcp-tenant-skill-draft.md` → `0xHoneyJar/construct-freeside`
- `capability-request-template-draft.md` → `0xHoneyJar/score-mibera` + `0xHoneyJar/construct-mibera-codex`

## Real drift caught + fixed

[...]

PR #19, opened 2026-05-02
feat(expression): character-voice loading + errors + silence + bedrock-first auto

zkSoju:
## Summary

Session 04 (kickoff-2026-05-01) — three V1 expression moments + the concurrent §4.0 cost micro-fix. Substrate routes; character expression fills.

- **§4.0 BEDROCK FLIP** — `agent-gateway.ts:resolveProvider` + `reply.ts:resolveChatProvider` auto-rule prefer `bedrock` when AWS env is present → `anthropic` (dev fallback) → `stub` → `freeside`. Boot log reads `llm: bedrock-direct (LLM_PROVIDER=auto resolved · …)`. Production already mitigated via `ANTHROPIC_API_KEY` removal; this codifies the rule. `.env.example` reordered to lead with bedrock.
- **§4.1 tool-mood-map** (`packages/persona-engine/src/expression/tool-mood-map.ts`) — Effect.Schema'd prefix→mood-pool routing for the six MCPs. `emojis` + `imagegen` route to empty pools (intentional silence).
- **§4.2 dispatch.onToolUse → composeToolUseStatusForCharacter** — ruggy fires an animated emoji from his catalog (whole-pool union random pick for variance); satoshi fires gumi-locked plain-text `"the ledger advances — one moment."`. Replaces the generic `tool-verb.ts` marker (deleted as dead code).
- **§4.3 error-register** (`expression/error-register.ts`) — per-character × per-error-class templates. Ruggy lowercase/slangy (`"cables got crossed — nothing came back. try again?"`); satoshi sentence-case/gnomic (`"The signal is unclear this window. Retry on the next."`). `formatErrorReply` routes through it. Strip-the-name + corporate-bot-banned tests pass.
- **§4.4 silence-register** (`apps/character-ruggy/silence-register.md` · 6 ruggy templates · cross-references satoshi's canonical pattern · plus `expression/silence-register.ts` substrate runtime + `composer.ts` LLM-bypass on flat windows: `events<100` + no climbed + no spotlight).
- **§4.5 timing** (`docs/EXPRESSION-TIMING.md` rhythm guide + `MIN_TOOL_PATCH_INTERVAL_MS` raised 500→800 ms per YANAGI cadence — jittery → alive zone).

### Numbers

- 21 files changed (+1912 / −134)
- 134 tests pass (33 new across `provider-resolution.test.ts` + 4 expression test files); 0 regressions in PR #18 surface-completeness / persona-tool-drift
- `bun run typecheck` green across both workspaces

### Deviations (surfaced per `feedback_spec_deviation_pattern`)

See NOTES.md Decision Log entries dated 2026-05-02 / `session-04-expression-layer`:

1. **Mood vocabulary** — used existing register-moods (`eyes`, `flex`, `noted`, `cool`, `dapper`, `peace`, `wave`) instead of seed §3 metaphor moods (`digging`, `pattern-watching`, `lore-checking`, `reading-the-room`, `who-is-this`). Capability-request to extend mood vocab is a follow-up if KEEPER pass surfaces register-fit issues.
2. **Pool union pick** — single-mood-pick (kickoff §3 wording) silently nulled when `noted` mood landed for ruggy (zero ruggy-kind `noted` emojis). Switched to whole-pool union → random pick from union; 4–7 ruggy emojis per category preserves variance and keeps the patch reliable.
3. **§4.4 wiring split** — substrate composer bypass shipped this session; ruggy persona.md cross-reference deferred pending gumi sync (per kickoff §9.2).
4. **`tool-verb.ts` deleted** as dead code; comments in dispatch.ts retain a one-line archaeological reference.
5. **`resolveProvider` exported** from `agent-gateway.ts` so the bedrock-first auto-rule has a unit-test surface (parallels existing `resolveChatProvider` export).

### Out of scope (per kickoff §6)

- §4.6 KEEPER user-truth listening pass — scheduled as V1.5 per `feedback_character_play_vs_friction` (listen first, project second).
- Throttle / circuit-breaker user-visible state.
- Imagegen-in-progress visibility.
- Removing the `anthropic` provider path; new providers; bedrock auth itself.
- Modifying satoshi's locked register (no custom emoji).
- New animated emoji creation (work with the catalog as-is).

## Test plan

- [x] `bun test` → 134 pass / 0 fail
- [x] `bun run typecheck` → clean across `packages/persona-engine` + `apps/bot`
- [x] Boot-log smoke check via `provider-resolution.test.ts` matrix (7 rows × digest path + 7 rows × chat path)
- [ ] **Manual dev-guild QA** (operator-run · post-merge):
  - `/ruggy <multi-tool prompt>` → animated emojis from catalog appear during loading; verify variance across multiple runs
  - `/satoshi <prompt>` → `"the ledger advances — one moment."` (no custom emoji); plain-text register holds
  - Force a compose error path → verify in-character register fires (ruggy: `"cables got crossed"`; satoshi: `"The signal is unclear …"`)
  - Trigger a flat-window digest (low-event zone) → verify ruggy posts a silence-register template instead of LLM elaboration
  - Watch timing: cadence should feel alive not jittery between tool patches (≥800ms gate)
- [ ] **ES coordination** — verify AWS Bedrock console activity reflects bot traffic post-merge; Jani's anthropic key should NOT increment for default deploys (kickoff §9.1)
- [ ] **Gumi coordination** — review `apps/character-ruggy/silence-register.md`; decide whether to promote into `persona.md` as an explicit "Quiet weeks" section parallel to satoshi's

## References

- Kickoff seed: `~/bonfire/grimoires/bonfire/specs/freeside-character-expression-layer-kickoff-2026-05-01.md`
- Parent doctrine: `~/vault/wiki/concepts/multi-axis-daemon-architecture.md` (§14 amendment + axis-3 capability-equipping)
- Substrate baseline: PR #18 (Effect.Schema MCP contracts) merged 2026-05-02 (`901503d`) + bridgebuilder follow-up `9cee5bf`


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#19. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "BEDROCK_TEXT_MODEL_ID default uses non-existent model alias",
      "severity": "HIGH",
      "category": "correctness",
      "file": ".env.example:43",
      "description": "The example sets `BEDROCK_TEXT_MODEL_ID=us.anthropic.claude-opus-4-7` and `ANTHROPIC_MODEL=claude-opus-4-7`. AWS Bedrock model IDs for Claude Opus follow the pattern `anthropic.claude-opus-4-...-v1:0` (e.g. `us.anthropic.claude-opus-4-1-20250805-v1:0` for Opus 4.1) — `us.anthropic.claude-opus-4-7` is not a valid Bedrock inference-profile ID and `claude-opus-4-7` is not a valid Anthropic model name (current published variants are `claude-opus-4-20250514`, `claude-opus-4-1-20250805`). If operators copy the example verbatim, both providers will fail with model-not-found at runtime.",
      "suggestion": "Replace with the actual model IDs in use (e.g. `us.anthropic.claude-opus-4-1-20250805-v1:0` and `claude-opus-4-1-20250805`), or leave the values blank with a comment pointing operators to a doc that lists currently-supported model IDs.",
      "confidence": 0.85
    },
    {
      "id": "F2",
      "title": "Auto-rule documentation contradicts implementation about ANTHROPIC_API_KEY precedence",
      "severity": "MEDIUM",
      "category": "documentation",
      "file": ".env.example:22-31",
      "description": "The .env.example comment says: 'when both bedrock env AND ANTHROPIC_API_KEY are set, LLM_PROVIDER=auto resolves to bedrock'. Then below: 'To opt into anthropic-direct in dev, EITHER: set LLM_PROVIDER=anthropic explicitly, OR unset AWS_BEARER_TOKEN_BEDROCK + BEDROCK_API_KEY in your shell.' This is correct. But the further-down comment at line 56 says 'Production should leave this UNSET so bedrock wins by elimination even if `auto` somehow misroutes' — which contradicts the codified preference order (bedrock wins by preference, not just elimination). The 'somehow misroutes' language plants doubt in a deterministic rule.",
      "suggestion": "Tighten the prose: bedrock wins by both preference AND elimination when production env is correct; ANTHROPIC_API_KEY removal is defense-in-depth, not the primary mechanism. Remove 'somehow misroutes' phrasing.",
      "confidence": 0.6
    },
    {
      "id": "F3",
      "title": "Schema.NonEmptyArray decode may strip readonly correctly but type assertion in test relies on runtime defaults",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/expression/silence-register.ts:42-46",
      "description": "`SilenceRegistrySchema = Schema.Record({ key: Schema.String, value: Schema.NonEmptyArray(TemplateSchema) })`. If a future contributor adds an empty array to TEMPLATES_RAW (e.g. for a placeholder character), `Schema.decodeUnknownSync` throws at module load — which is desired loud-fail behavior. However this means importing this module has a load-time side effect that can crash the bot on startup with a parse error. The error message would be Effect's default which may not be especially legible to operators. Consider wrapping with a more contextual error.",
      "suggestion": "Add a try/catch around the decode call (or use Effect's Schema.decodeUnknownEither) that re-throws with a contextual prefix like 'silence-register: invalid TEMPLATES_RAW shape — every character must have ≥1 template'. Same pattern applies to error-register.ts and tool-mood-map.ts.",
      "confidence": 0.4
    },
    {
      "id": "F4",
      "title": "EmojiMood literal list duplicated between schema and registry",
      "severity": "LOW",
      "file": "packages/persona-engine/src/expression/tool-mood-map.ts:50-78",
      "description": "The 27 mood literal strings are hardcoded in `Schema.Literal(...)` here, but they also live in `../orchestrator/emojis/registry.ts` (referenced by `ALL_MOODS` and the `EmojiMood` type). If `registry.ts` adds or removes a mood, this schema silently drifts — a runtime decode failure occurs only if someone uses the new mood here. There's no compile-time link.",
      "suggestion": "Either (a) export the mood literal array from registry.ts and use `Schema.Literal(...ALL_MOODS)` here, or (b) add a TS-level `assertEqual<EmojiMood, Schema.Schema.Type<typeof MoodSchemaInner>>()` parity check so drift fails compilation.",
      "confidence": 0.75
    },
    {
      "id": "F5",
      "title": "Spec deviation: union-pick across pool moods rather than pick-mood-then-emoji",
      "severity": "REFRAME",
      "category": "design",
      "file": "packages/persona-engine/src/expression/loading-status.ts:71-85",
      "description": "The kickoff §3 spec says 'pick one mood at random, then look up an emoji from the character's catalog.' Implementation unions all moods in the pool and picks one emoji from the union. The deviation is documented in NOTES.md and the module docstring with a sound rationale (single-mood-pick can return null when a mood has zero matching-kind emojis). This is a reasonable substrate-correctness call — flagging as REFRAME for visibility, not as a bug. The trade-off: union-pick weights emojis by frequency-of-membership across moods (an emoji in 2 of 3 pool moods has 2x odds in the union vs single-mood-pick where the mood is picked uniformly first). For 3 moods × 4-7 emojis the bias is small but real.",
      "suggestion": "Acceptable as-is given the deviation is logged. If KEEPER pass surfaces uneven emoji distribution, a two-step weighted pick (uniform mood selection, fall back to next mood if pool empty) would correct the bias without losing the patch-emission guarantee.",
      "confidence": 0.7
    },
    {
      "id": "F6",
      "title": "isFlatWindow predicate ignores climbed-only-when-rank-changes-shape-exists",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/expression/silence-register.ts:101-107",
      "description": "`rawStats.rank_changes.climbed.length > 0` will throw if `rank_changes` is undefined. The fixture in tests defines it for both v1 and v2 shapes, but if any historical RawStats variant lacks `rank_changes`, the composer's flat-window check throws and breaks the whole compose path on what should be a robustness boundary. Schema layer probably enforces this, but defensive null-handling here is cheap.",
      "suggestion": "Use `rawStats.rank_changes?.climbed?.length > 0` or assert the shape upstream. Confirm the RawStats schema makes rank_changes required across all schema_versions.",
      "confidence": 0.5
    },
    {
      "id": "F7",
      "title": "Silence-register wasted LLM call on flat windows",
      "severity": "LOW",
      "category": "performance",
      "file": "packages/persona-engine/src/compose/composer.ts:61-90",
      "description": "Comment acknowledges the trade-off: digest fetch + LLM invocation are parallelized, then on flat-window the LLM result is discarded. For bedrock-priced production, every flat-window digest spends an Opus call's worth of tokens for nothing. Comment claims 'flat windows are the minority case' but cron-driven digests across 4 zones × N characters can rack up. The fix would be to fetch digest first, check flatness, then conditionally invoke — adding latency to the common path.",
      "suggestion": "Quantify before optimizing. Add a metric/log counter for flat-window detections so operator can see the actual rate. If flat windows are >10% of digests, switch to sequential fetch-then-invoke. Acceptable trade-off as documented for now.",
      "confidence": 0.5
    },
    {
      "id": "F8",
      "title": "displayName fallback inconsistent — composeErrorReply gets the resolved name, but unknown-character fallback path may leak character.id",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/expression/error-register.ts:135-148",
      "description": "`composeErrorReply` always returns `**${displayName}**\\n\\n${body}`. When `getErrorTemplate` returns null (unknown character), the fallback body 'something broke. try again?' is used. Display name passes through, so for an unknown character the test passes a manually-supplied 'Stranger' — fine. But in dispatch.ts the call site uses `character.displayName ?? character.id`. If a new character is added to character-loader without a displayName AND without an error-register entry, the user sees `**character-id**\\n\\nsomething broke. try again?`. Low-impact but worth noting.",
      "suggestion": "Acceptable; ensure character-loader requires displayName or warns on missing.",
      "confidence": 0.3
    },
    {
      "id": "F9",
      "title": "PRAISE: Provider-resolution matrix test is well-structured",
      "severity": "PRAISE",
      "category": "testing",
      "file": "apps/bot/src/tests/provider-resolution.test.ts",
      "description": "Table-driven matrix exercising both `resolveProvider` (digest path) and `resolveChatProvider` (chat path) against the same fixture rows is exactly the right shape to lock the auto-rule against drift. The explicit-provider tests cover the 'throws' edge cases. The bareConfig fixture isolates the unit under test cleanly.",
      "suggestion": "No action needed.",
      "confidence": 0.95
    },
    {
      "id": "F10",
      "title": "PRAISE: Substrate-quiet-on-unmapped-tool fallback matches civic-layer doctrine",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/persona-engine/src/expression/tool-mood-map.ts:160-170",
      "description": "Returning null for unmapped tools (rather than a generic `🔧 invoking ...`) is the correct call: substrate-shaped quiet during a character moment is preferable to substrate-shaped noise. The three-state semantics (null = unknown/skip, [] = intentional silence, [...moods] = pick) are clean.",
      "suggestion": "No action needed.",
      "confidence": 0.95
    },
    {
      "id": "F11",
      "title": "PRAISE: Strip-the-name test for error-register voice register",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/persona-engine/src/expression/error-register.test.ts:78-105",
      "description": "Asserting that ruggy templates start lowercase and satoshi templates start uppercase, plus that they don't share template bodies, operationalizes the Alexander craft 'strip the name' test in code. The CORPORATE_BOT_BANNED list is a good substrate-level guardrail against doctrine drift.",
      "suggestion": "Consider adding a max-length / no-emoji-in-satoshi-templates assertion to lock the gnomic register further if voice drift becomes a concern.",
      "confidence": 0.9
    },
    {
      "id": "F12",
      "title": "Test asserts rendered emoji format matches `<a?:name:id>` but render path may produce non-animated for some entries",
      "severity": "LOW",
      "category": "testing",
      "file": "packages/persona-engine/src/expression/loading-status.test.ts:44-52",
      "description": "Tests for satoshi assert the plain-text register `'the ledger advances — one moment.'` exactly. If anyone tweaks the template string (typo fix, em-dash variation), this fails — which is desired test rigidity for a locked register. However the variance test (`30 invocations yield > 1 distinct emoji`) is probabilistic; with bad luck it could flake. Assuming the union has ≥4 emojis with non-uniform variance, the false-failure rate is low but nonzero. Worth seeding randomness or asserting against `>= 2` over a much larger N if flakes appear.",
      "suggestion": "Acceptable for a non-critical test. If flakes surface, inject a seedable RNG into pickRandomMood / loading-status, or expand to 100 trials.",
      "confidence": 0.4
    },
    {
      "id": "F13",
      "title": "Inconsistency: dispatch tool patch comments reference YANAGI but timing doc is unenforced ceiling",
      "severity": "LOW",
      "category": "documentation",
      "file": "docs/EXPRESSION-TIMING.md:11-15",
      "description": "The doc table shows '14m30s' as token lifetime ceiling and 'no enforced cap' for tool-call patch interval ceiling. The 4-second 'ABANDONED' threshold is named but unenforced. This is fine but the doc explicitly notes 'orchestrator round-trips can take 2-15 seconds' — so the 4s ceiling is aspirational, not real. Worth being explicit that the only enforced limit is the 800ms floor.",
      "suggestion": "Add a one-line summary in the table: 'enforced floor: 800ms · ceiling: orchestrator-bounded, not substrate-enforced'.",
      "confidence": 0.4
    },
    {
      "id": "F14",
      "title": "Hardcoded MIN_TOOL_PATCH_INTERVAL_MS and FLAT_WINDOW_EVENT_THRESHOLD not env-tunable",
      "severity": "LOW",
      "file": "apps/bot/src/discord-interactions/dispatch.ts:272-278",
      "description": "The 800ms gate and 100-event flat-window threshold are compile-time constants. The KANSEI doc says KEEPER pass may want to revisit after user feedback — when that happens, tweaking requires a code change + redeploy rather than env adjustment. For values explicitly described as 'hypothesis pending confirmation,' env-tunability would lower the cost of iteration.",
      "suggestion": "Optional: lift to Config (e.g. `MIN_TOOL_PATCH_INTERVAL_MS`, `FLAT_WINDOW_EVENT_THRESHOLD`) with current values as defaults. Trade-off: more config knobs vs faster iteration on KANSEI tuning.",
      "confidence": 0.4
    },
    {
      "id": "F15",
      "title": "PR commit removes tool-verb.ts but no greptest verifies all references gone",
      "severity": "LOW",
      "category": "correctness",
      "file": "apps/bot/src/discord-interactions/tool-verb.ts",
      "description": "File deleted; dispatch.ts updated to use composeToolUseStatusForCharacter. If any test or other module imports './tool-verb', tsc would catch it — but the diff doesn't show search results. Trust tsc but verify no stale grep references in docs/comments.",
      "suggestion": "Run `rg toolVerb apps/ packages/` to confirm zero references remain. If clean, no action.",
      "confidence": 0.3
    }
  ]
}
```


## Callouts

Issue #20, opened 2026-05-02
WITNESS QA: codex MCP integration · live trial in dev guild (gated on construct-mibera-codex#69)


zkSoju:
## Summary

WITNESS QA checklist for testing the codex MCP integration end-to-end through this bot. Bot wiring is complete (`apps/character-satoshi/character.json` has `codex` in `mcps`, `packages/persona-engine/src/orchestrator/index.ts:138-149` registers the codex MCP server when `CODEX_MCP_URL` is set), but two upstream items must land first.

### 1. construct-mibera-codex Dockerfile fix (PR open · awaiting merge)

**Status today**: prod codex MCP at `https://mcp.0xhoneyjar.xyz/codex/mcp` returns `serverInfo.version: "1.1.0"` even though source on main is v1.4.0 (PR #65 merged 2026-05-02T22:30:53Z). Latest Railway deploy attempt (`828bff67` at 2026-05-02T23:26Z) **failed** because `node-llama-cpp@3.18.1` postinstall needs `git`/`make`/`glibc` — alpine image lacks all three.

Fix: https://github.com/0xHoneyJar/construct-mibera-codex/pull/69 — Dockerfile `node:20-alpine` → `node:20-slim` + builder-stage `apt-get install git make python3 g++ ca-certificates`.

Until PR #69 lands AND Railway redeploys successfully, deployed `tools/list` is missing `search_codex` (intent layer). The 8 v1.1.0 lookup_* tools work today; intent search does not.

### 2. CODEX_MCP_URL env var on this bot's prod deploy

Per `packages/persona-engine/src/orchestrator/index.ts:146`:
```ts
if (config.CODEX_MCP_URL) {
  servers.codex = { url: \`\${config.CODEX_MCP_URL}/mcp\`, ... };
}
```
And per `orchestrator/index.ts:168` comment (verbatim): *"CODEX_MCP_URL) are unset in production — H4 from the kickoff brief"*.

Reviewer notes for sprint-3 confirm AC-3.1 + AC-3.2 are `⏸ ACCEPTED-DEFERRED` because *"requires live deployment + CODEX_MCP_URL"*.

**Operator action on Railway dashboard for `prod-ruggy` service**:
```
service: prod-ruggy
env var: CODEX_MCP_URL=https://mcp.0xhoneyjar.xyz/codex
save → auto-redeploy (~30s)
```

## Full WITNESS QA checklist

[...]

PR #21, opened 2026-05-03
feat(v07a3): env-aware image composition + voice anti-hallucination

zkSoju:
## What V0.7-A.3 unlocks (capability landscape)

Two gaps surfaced in micodex-09a dogfood (2026-05-02 PM, issue [#20](https://github.com/0xHoneyJar/freeside-characters/issues/20)) close in this PR:

**1. Image composition gap** (§1-§10):
The codex MCP returns grail image URLs (`https://assets.0xhoneyjar.xyz/Mibera/grails/black-hole.png`); the composer never surfaces them. URL-paste fallback breaks because Discord automod filters `assets.0xhoneyjar.xyz` on community servers (operator's own server confirmed). **Fix mirrors imagegen V0.11.3 / PR #15**: download bytes server-side, attach via `webhook.send({files: [...]})`. Bytes are the contract; URL never appears in the message.

**2. Voice anti-hallucination gap** (§11):
On abstract prompts (`/ruggy "show me transformation"`), characters MIX canonical refs (✓ `@g4488 Hermes`, `@g876 Black Hole`) with **invented categories** (✗ "Death-tarot grail", "Tower", "alchemical panels"). Substrate is queried (`tool_uses=2`) but voice composer ignores literal hits in favor of training-data occult-iconography. **Fix**: persona instructions ("ALWAYS cite literal `@g<id>` ref before creative interpretation; NEVER invent grail categories") + composer-side `grail-ref-guard` validator (V1: warning footer; V1.5: strip).

## Spec + sprint references

- Combined PRD+SDD: `grimoires/loa/specs/build-env-aware-image-composition-v07a3.md` (`external_work_plan: true`)
- Sprint plan: `grimoires/loa/sprint-v07a3.md`
- Beads epic: `bd-ypi` · tasks `bd-ypi.1` through `bd-ypi.10` (all closed)
- WITNESS QA: `grimoires/loa/qa/qa-cycle-V07A3-2026-05-02.md`

## What ships (3 commits)

| commit | scope | LOC delta |
|---|---|---|
| `1c4c5a2` | phase A — composeWithImage substrate + composeReplyWithEnrichment + dispatch wire | +651 / -11 |
| `4fb6698` | phase B — grail-ref-guard validator + persona anti-hallucination rules | +430 / -8 |
| `c9cebc4` | phase C — WITNESS QA cycle file | +264 / -1 |

**File-level summary**:

NEW:
- `packages/persona-engine/src/deliver/embed-with-image.ts` (138 LOC) — `composeWithImage` helper + types
- `packages/persona-engine/src/deliver/embed-with-image.test.ts` (8 tests)
- `packages/persona-engine/src/deliver/grail-ref-guard.ts` (138 LOC) — `validateGrailRefs` + `appendGrailRefGuardFooter`
- `packages/persona-engine/src/deliver/grail-ref-guard.test.ts` (15 tests)
- `grimoires/loa/qa/qa-cycle-V07A3-2026-05-02.md` (260 lines · 6 verification surfaces + 3 showcase)

MODIFIED:
- `packages/persona-engine/src/deliver/webhook.ts` (+14 LOC · `sendChatReplyViaWebhook(files?)` arg)
- `packages/persona-engine/src/orchestrator/index.ts` (+85 LOC · `ToolResultEvent` type + `tool_result` block capture from SDK user messages)
- `packages/persona-engine/src/compose/reply.ts` (+213 LOC · `composeReplyWithEnrichment` sibling + grail-ref-guard wiring)
- `packages/persona-engine/src/index.ts` (+13 LOC · public exports)
- `apps/bot/src/discord-interactions/dispatch.ts` (+25 LOC · swap composeReply→composeReplyWithEnrichment, files-on-first-chunk)
- `apps/character-satoshi/persona.md` (+30 LOC · CODEX GROUNDING section, sentence-case gnomic register)
- `apps/character-ruggy/persona.md` (+27 LOC · CODEX GROUNDING section, lowercase OG register)

## Acceptance criteria (sprint plan)

- [x] AC-1: `composeWithImage(text, [grailToolResult])` returns `{content, files: [Buffer]}` with image bytes (Tasks 1.1 + 1.5 — 8 tests)
- [x] AC-2: webhook delivery accepts files arg → attaches as discord.js `AttachmentBuilder` (Task 1.2)
- [x] AC-3: composeReply contract unchanged · `composeReplyWithEnrichment` is additive surface (Task 1.3, additive option C per spec §4.1)
- [x] AC-4: per-character MCP scoping holds · attach fires only for grail tool results (Task 1.4 — codex tools only filter)
- [x] AC-5: `grail-ref-guard` flags `@g99999` (invalid) but allows `@g876` (canonical) (Tasks 1.6 + 1.9 — 15 tests)
- [x] AC-6: persona.md updates land canonical-citation rule for ruggy + satoshi (Task 1.7)
- [x] AC-7: post-processing footer appears when `invalid.length > 0` (Task 1.8)
- [x] AC-8: `bun run typecheck` clean across workspace
- [x] AC-9: all unit tests pass (111 persona-engine + 46 bot = 157 / 0 fail / 553 expects)
- [x] AC-10: WITNESS regression covers S0-S6 + SC1-SC3 (Task 1.10)
- [x] AC-11: PR opened against `freeside-characters:main`

## Operator-bounded (KEEPER pass · post-merge per spec §7)

- [ ] **G-6 KEEPER PASS** — `/satoshi prompt:"the dark grail"` in 2nd server with `assets.0xhoneyjar.xyz` blocklisted; bytes-attachment SHOULD persist where pre-A.3 URL-paste was deleted by automod. **This is the doctrine-validation gate** for `~/vault/wiki/concepts/environment-aware-composition.md`.
- [ ] G1 inheritance (already green): `/satoshi-image prompt:"test"` works in prod via PR #15 V0.11.3 — Surface 6 in WITNESS QA verifies no regression.

## Doctrine validation

This cycle validates `~/vault/wiki/concepts/environment-aware-composition.md` (the doctrine candidate from spec §10). Three-layer mutability frame:

| Layer | Mutability | Who Controls |
|---|---|---|
| metadata FIELD (`image`, `image_url`) | stable | substrate authority (codex) |
| URL SHAPE (`assets.0xhoneyjar.xyz/...`) | operator-mutable | operator (CDN choice, paths) |
| URL AS RENDERED in target environment | environment-mutable | end-user environment (Discord automod, Slack URL preview) |

If KEEPER G-6 succeeds in 2+ servers (one with blocklist), candidate promotes → vault.

## Test plan (verification)

- [x] `bun run typecheck` green across workspace (`packages/persona-engine` + `apps/bot`)
- [x] `bun test packages/persona-engine/src/deliver/embed-with-image.test.ts` → 8/8 pass
- [x] `bun test packages/persona-engine/src/deliver/grail-ref-guard.test.ts` → 15/15 pass
- [x] `bun run --cwd packages/persona-engine test` → 111/111 pass (no regressions in prior 88)
- [x] `bun run --cwd apps/bot test` → 46/46 pass
- [ ] WITNESS QA Surfaces 1-6 in dev guild (operator runs)
- [ ] WITNESS SC3 / G-6 KEEPER pass in 2nd server with blocklist (operator runs)

## Deviations from spec (with rationale)

1. **`undici` import → global `fetch`**: spec §4.2 reference impl says `import { fetch } from 'undici'`; this codebase has no undici dep and uses global fetch consistently (Bun 1.3 + Node 22+ ship undici-based fetch globally). Surgical choice — no new dep, same runtime behavior, supports `AbortSignal.timeout`. Documented in `embed-with-image.ts` header comment.

2. **Canonical 43 grail IDs hardcoded as 7-id subset**: this repo doesn't carry a registry of all 43 IDs (those live in `construct-mibera-codex` upstream). V1 hardcodes the 7 IDs explicitly named in spec/sprint/MEMORY (876, 4488, 235, 6458, 4221, 1606, 6805). Other canonical refs fall to `invalid` (= warning footer) UNLESS passed via `sessionRefs` from active tool results. V1.5 roadmap (documented in `grail-ref-guard.ts`): replace with startup-time fetch from `mcp__codex__list_archetypes`. Conservative-but-honest: behavior is observable, non-blocking, doesn't ship false canonical claims.

3. **gumi voice-authority sync**: persona.md changes (CODEX GROUNDING section in both ruggy + satoshi) are register-matched (sentence-case gnomic for satoshi, lowercase OG for ruggy) but bypass gumi's canonical authoring authority. Queued in WITNESS QA operator-bounded list — operator pings gumi async with diff for review; revert option preserved if voice register drifts.

## Bridgebuilder review

Per spec §9 #5 — recommend running `/bridgebuilder-review` before merge per established cadence (PR #69, #70, #71, #68 all went through). Suggested focus areas:
- composer logic in `composeReplyWithEnrichment` (envelope shape variance handling)
- grail-ref-guard regex correctness (`@g(\d+)` and `#(\d+)` patterns)
- tool envelope shape coverage in `extractGrailCandidates` + `pickFirstGrailFromEnvelope`
- whether the V1 hardcoded 7-id canonical set is the right conservative default vs full 43

## Bridgebuilder fixes applied (2026-05-02)

Bridgebuilder review at https://github.com/0xHoneyJar/freeside-characters/pull/21 surfaced 11 findings. 5 addressed in this PR, 6 deferred with rationale.

| Finding | Severity | Commit | Resolution |
|---|---|---|---|
| **F5** SSRF: image fetch has no URL allowlist | HIGH security | `b85ec12` | `ALLOWED_SCHEMES = {'https:'}` + `ALLOWED_IMAGE_HOSTS = {'assets.0xhoneyjar.xyz'}` allowlist via `isAllowedImageUrl()` gate before `fetch()` + `redirect: 'error'` so undici throws on 3xx redirect-to-internal pivots. 9 new tests (allowlist unit cases + SSRF graceful-degrade). |
| **F6** No size cap on fetched bytes | MEDIUM security | `b85ec12` | `MAX_ATTACHMENT_BYTES = 8 * 1024 * 1024` (8MB; under Discord's 25MB limit per spec §2 invariant 6). Pre-fetch Content-Length check + post-arrayBuffer belt-and-braces. 2 new tests. |
| **F1** REF_HASH_PATTERN overly broad | MEDIUM correctness | `bc06c1f` | Drop `#<id>` form entirely; only `@g<id>` recognized. The `@g` prefix IS the disambiguator. Persona instructions in both `apps/character-{ruggy,satoshi}/persona.md` updated to require `@g<id>` form + forbid bare `#876`. 3 new tests (issue-ref shapes ignored). |
| **F4** User-visible footer leaks internal jargon | MEDIUM ux/design | `bc06c1f` | `GRAIL_REF_GUARD_FOOTER` removed from user-visible Discord text — broke persona illusion (KEEPER/V1.5/corpus signal jargon). `appendGrailRefGuardFooter` renamed to `inspectGrailRefs`; text returned UNCHANGED. Validation surfaces via existing `console.warn` + structured `grailRefValidation` field on `EnrichedReplyResult`. 4 new tests (telemetry-only contract). |
| **F3** V1 hardcoded 7-of-43 conservative set | MEDIUM design | `bc06c1f` | Kept as-is. Now that the user-visible footer is gone (F4), false positives are observable telemetry signal (helpful for V1.5 corpus expansion) rather than user-visible noise. V1.5 roadmap unchanged: hydrate canonical 43 from `mcp__codex__list_archetypes` at startup. |

### Bridgebuilder triage notes (deferred)

| Finding | Severity | Defer rationale |
|---|---|---|
| **F2** session-ref regex asymmetry | LOW | V1.5 cleanup; doesn't affect correctness in V1 (sessionRefs strip handles both `@g876` and `876` shapes today). |
| **F7** Content-Type from URL extension | LOW | Discord renders by sniffing bytes; `inferExtension` mismatch is cosmetic at most. V1.5: trust `Content-Type` response header when present. |
| **F8** PATCH fallback drops attachments silently | LOW | DESIGN intent per spec §2 graceful-degrade philosophy: text reaches the user even when bytes don't. Not a bug. |
| **F9** search_codex top-1 ordering | LOW | Depends on codex MCP envelope contract; defer until contract is documented in `construct-mibera-codex` upstream. |
| **F10** Footer chunking order | LOW | Moot: footer removed in F4. |
| **F11** Non-text content blocks dropped | LOW | Existing `console.warn` log addition (orchestrator/index.ts already covers tool_use/tool_result paths) is sufficient signal. Defer richer surfacing to V1.5. |

### Verify gates (post-triage)

- [x] `bun run typecheck` green across workspace
- [x] `bun run --cwd packages/persona-engine test` → 128/128 pass (was 111 baseline · +17 new tests this triage)
- [x] `bun run --cwd apps/bot test` → 46/46 pass (no regressions)
- [x] All 3 commits pushed to `feat/v07a3-env-aware-image-composition`
- [x] NOTES.md Decision Log + Session Continuity rows added (3 new Decision Log entries)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#21. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "REF_HASH_PATTERN may match unrelated #digit tokens",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/deliver/grail-ref-guard.ts",
      "description": "REF_HASH_PATTERN = /#(\\d+)/g matches any '#' followed by digits anywhere in the text. This will catch issue refs (#123 in 'fixed in #123'), markdown anchors, channel mentions in some forms, ordinal lists like '#1', and other non-grail uses. Many of these will fall to 'invalid' and trigger the warning footer for entirely benign text. The @g pattern is well-targeted; the # pattern is overly broad given common Discord/markdown conventions.",
      "suggestion": "Either require a word boundary + grail-shaped digit length (e.g. /\\B#(\\d{3,})\\b/) or drop the # form entirely and require @g<id>. If both forms must be supported, consider only validating # refs when context suggests a grail (e.g., near 'grail', 'codex', or following a known noun pattern). At minimum, document that #digit triggers the guard so persona authors avoid that idiom.",
      "confidence": 0.8
    },
    {
      "id": "F2",
      "title": "Session-ref regex permits malformed refs",
      "severity": "LOW",
      "category": "correctness",
      "description": "In collectRefsFromValue, the regex /^@?g?\\d+$/ accepts strings like 'g123' (no @) or '@123' (no g). These won't match REF_AT_PATTERN /@g(\\d+)/ in the validator anyway, so storing them is harmless but adds noise. More importantly, the sessionSet construction in validateGrailRefs strips '^[@#]g?' which would normalize '@123' to '123' but '123' itself is a valid id form — so refs collected as bare numeric ids from .id fields work, but only because of the strip. The asymmetry between collection regex and validator strip is fragile.",
      "suggestion": "Tighten the collection regex to /^(@g\\d+|\\d+)$/ to match the two canonical session-ref shapes the validator's strip handles, and add a unit test exercising the collection→validation roundtrip.",
      "confidence": 0.5
    },
    {
      "id": "F3",
      "title": "Hardcoded canonical grail set will produce false-positive warnings",
      "severity": "MEDIUM",
      "category": "design",
      "file": "packages/persona-engine/src/deliver/grail-ref-guard.ts",
      "description": "CANONICAL_GRAIL_IDS contains only 7 ids out of the documented 43 canonical grails. Any time the LLM cites a real canonical grail not in this list (e.g., the other 5 zodiac signs, most planets, most ancestors) without the codex tool firing in-session, the footer will be appended on a perfectly valid reference. The author flags this as 'intentionally conservative' but the cost is incorrect operator signal — the warning footer will fire frequently for legitimate citations, training operators to ignore it before V1.5 lands.",
      "suggestion": "Either (a) include the full 43-id canonical list (the spec claims they're enumerable), (b) gate the footer behind a higher-confidence signal (e.g., only when the LLM cited a ref AND no tool result exists for it), or (c) make this a counter rather than a user-visible footer until V1.5. As-is the footer leaks engineering jargon ('KEEPER: contains unverified ref(s) · captured for V1.5 corpus signal') to end users which is a UX concern.",
      "confidence": 0.75
    },
    {
      "id": "F4",
      "title": "User-visible footer leaks internal jargon",
      "severity": "MEDIUM",
      "category": "ux",
      "file": "packages/persona-engine/src/deliver/grail-ref-guard.ts",
      "description": "GRAIL_REF_GUARD_FOOTER = '_[KEEPER: contains unverified ref(s) · captured for V1.5 corpus signal]_' is appended to the reply text that users see in Discord. This exposes internal versioning ('V1.5'), internal terminology ('KEEPER', 'corpus signal'), and a debugging concept ('unverified ref') directly to end users. Combined with F3's likely false positives, regular users will see this footer attached to character voice frequently, breaking the persona illusion.",
      "suggestion": "Move the warning to operator-only telemetry (console.warn already exists at the call site) and do not mutate the user-facing text in V1. If a user-visible signal is needed, design it as in-voice ('the codex doesn't return that') rather than as engineering metadata.",
      "confidence": 0.85
    },
    {
      "id": "F5",
      "title": "Image attachment fetch has no SSRF / URL allowlist",
      "severity": "HIGH",
      "category": "security",
      "file": "packages/persona-engine/src/deliver/embed-with-image.ts",
      "description": "fetchAttachment calls fetch(url) on whatever URL strings appear in the parsed tool result envelopes. There is no scheme check (http/https), no host allowlist (e.g., assets.0xhoneyjar.xyz), and no IP filtering. A compromised or misbehaving MCP tool could return file://, data:, or internal-network URLs (e.g., http://169.254.169.254/ for cloud metadata, http://localhost:*/ for adjacent services). Bun/undici fetch follows redirects by default, so even an https://attacker.example/ URL could redirect to internal addresses. The bytes are then attached to a Discord webhook — exfiltration channel.",
      "suggestion": "Add a strict allowlist on the URL host (and scheme=https) before fetch. Disable redirect-following or restrict redirects to the same allowlist. The codex tool's image hosts should be enumerable. Log + drop disallowed URLs.",
      "confidence": 0.9
    },
    {
      "id": "F6",
      "title": "No size cap on fetched attachment bytes",
      "severity": "MEDIUM",
      "category": "security",
      "file": "packages/persona-engine/src/deliver/embed-with-image.ts",
      "description": "fetchAttachment reads res.arrayBuffer() in full with no size limit. A malicious or misbehaving image host could return a multi-GB stream, exhausting memory before the 5s timeout fires (bytes can flow fast). Discord's webhook attachment cap is ~25MB but nothing in this code enforces it; oversized payloads will fail the webhook send AFTER buffering everything.",
      "suggestion": "Stream-read with a size cap (e.g., 10MB) by reading res.body chunk-by-chunk and aborting when the threshold is exceeded. Alternatively, check Content-Length header (with the understanding that it can be absent or wrong) before calling arrayBuffer.",
      "confidence": 0.75
    },
    {
      "id": "F7",
      "title": "Content-Type is inferred from URL extension, not response header",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/deliver/embed-with-image.ts",
      "description": "extToContentType derives the MIME type purely from the URL's filename extension. The response's Content-Type header is ignored. URLs without an extension (or with unusual ones) silently default to image/png even when the server returns image/jpeg. While Discord generally infers from bytes for display, the attachment metadata becomes inconsistent.",
      "suggestion": "Prefer res.headers.get('content-type') when present and image/*; fall back to extension inference only when the header is absent or non-image.",
      "confidence": 0.6
    },
    {
      "id": "F8",
      "title": "PATCH fallback path drops attachments silently",
      "severity": "LOW",
      "category": "ux",
      "description": "When the webhook delivery throws and the code falls back to deliverViaInteraction, the attachedFiles are not passed and the image is silently dropped. Users see only text. This may be acceptable behavior, but it's worth confirming this is intentional — and the comment notes it for ephemeral but not for the catch-fallback case.",
      "suggestion": "Either propagate attachments to the PATCH path (if discord-interactions interaction PATCH supports multipart) or log a warning when fallback drops attachments so operators can see this happened.",
      "confidence": 0.5
    },
    {
      "id": "F9",
      "title": "search_codex top-1 selection assumes ordering",
      "severity": "LOW",
      "category": "correctness",
      "description": "pickFirstGrailFromEnvelope takes results[0] for search_codex per spec §2.2. The MCP tool's ordering contract isn't visible in this PR — if it returns lexicographic, by-relevance, or unspecified order, 'top-1 has image' may not match user intent. Also, if results[0] has no image but results[1] does, the candidate is dropped even though a valid image exists in the envelope.",
      "suggestion": "Consider scanning all results for the first one with an image field rather than only inspecting results[0]. Or document explicitly that the codex tool guarantees results are sorted by intended display priority.",
      "confidence": 0.5
    },
    {
      "id": "F10",
      "title": "Footer added before re-chunking but after first-chunk attachment decision",
      "severity": "LOW",
      "category": "correctness",
      "description": "composeReplyWithEnrichment re-chunks after appending the footer (good), but the footer is appended to the LAST chunk's content area implicitly via splitForDiscord. The webhook delivery path attaches the image only to chunk 0. If the footer pushes content into a new final chunk, the attached chunk and the footer chunk are now well separated in the channel. This is fine functionally but worth noting that voice→image→...→footer is the rendered order, which may surprise readers.",
      "suggestion": "If the intent is voice|image and footer-as-aside, this is fine. If footer should ride near the citation, consider inlining it differently. Document the chosen order in the design.",
      "confidence": 0.4
    },
    {
      "id": "F11",
      "title": "extractToolResultText silently drops non-text content blocks",
      "severity": "LOW",
      "category": "observability",
      "description": "extractToolResultText only collects parts with type==='text'. ImageBlockParam, ToolUseBlockParam variants, and any future block types are ignored. For MCP tools returning image blocks (which could happen for image-returning tools), the parsed envelope will be empty/missing data and won't be available downstream.",
      "suggestion": "If non-text blocks appear in production, log at minimum that they were observed and dropped. Consider extracting image URLs from image blocks too if they show up.",
      "confidence": 0.4
    },
    {
      "id": "F12",
      "title": "Good additive design preserving existing composeReply contract",
      "severity": "PRAISE",
      "category": "design",
      "description": "composeReplyWithEnrichment is implemented as a sibling that wraps composeReply rather than mutating it, with the existing onToolResult callback chained rather than replaced. Existing callers (digest, weaver, callout) are unaffected; new callers opt in. This is a clean way to introduce optional enrichment without ripple-edit risk.",
      "suggestion": "n/a — keep the pattern.",
      "confidence": 0.9
    },
    {
      "id": "F13",
      "title": "Strong test coverage for new modules",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/persona-engine/src/deliver/embed-with-image.test.ts",
      "description": "embed-with-image.test.ts and grail-ref-guard.test.ts both cover happy path, failure modes (404, network throw, empty results), edge cases (sessionRefs override, dedup, hash vs at form), and clamp behavior. Tests are clear and exercise real branching.",
      "suggestion": "n/a.",
      "confidence": 0.9
    },
    {
      "id": "F14",
      "title": "Graceful degradation on fetch failure is the right default",
      "severity": "PRAISE",
      "category": "reliability",
      "file": "packages/persona-engine/src/deliver/embed-with-image.ts",
      "description": "Fetch failures (timeout, 404, network error) silently fall through to text-only payload rather than blocking the reply. The 5s AbortSignal.timeout and try/catch around fetch is appropriately defensive. Voice delivery is the priority; image is a bonus.",
      "suggestion": "n/a, though see F5/F6 for security hardening that complements this.",
      "confidence": 0.9
    }
  ]
}
```


## Callouts

PR #22, opened 2026-05-03
fix(v07a3): search_codex envelope shape + URL strip from reply text (hotfix)

zkSoju:
## V0.7-A.3 hotfix on top of #21

PR #21 merged (squash `dd9f088`) but operator's runtime testing 2026-05-02 PM surfaced two bugs the lab tests didn't catch. Both fixed in this single PR. Telemetry-first per `feedback_telemetry_first_when_hypothesis_fails` — went back to operator's prod log + the codex MCP source before generating the second hypothesis.

## B1 — `search_codex` envelope shape silently dropped every hit

**Symptom**: operator's "the dark grail" intent flow returned without any image attached even though `search_codex` ran. attached=0 on the wire.

**Root cause** (verified against codex MCP source):

The codex MCP `search_codex` handler returns `SearchHit[]` (top-level JSON array, no `{ results: }` wrapper). Sources:
- `/tmp/codex-probe/src/lookups/search.ts:204` — `export function searchCodex(opts: SearchOptions): SearchHit[]`
- `/tmp/codex-probe/src/server.ts` `search_codex` handler — `return { content: [{ type: "text", text: JSON.stringify(hits, null, 2) }] }`

PR #21 expected `{ results: [...] }` per spec §2.2 wording — the spec was wrong about the envelope shape. Old code:
```ts
const obj = parsed as { results?: unknown };
const results = Array.isArray(obj.results) ? obj.results : null;
```
Always saw `null` because `parsed` is the array itself, not a wrapper.

**Fix** (`packages/persona-engine/src/compose/reply.ts`):
1. `pickFirstGrailFromEnvelope` now treats `parsed` as `SearchHit[]` directly when `toolName === mcp__codex__search_codex`
2. **Scan-all-results** (bridgebuilder F9 deferred) — top hit may be a non-grail entry (core-lore, etc.); walk hits in order until one with `image` field surfaces. Top-1-only would silently drop grails when the MCP returns mixed-collection hits.
3. Telemetry: `console.warn` when search returns hits but ZERO have `image` field — operators diagnose envelope-shape regression vs non-grail-only result sets from trajectory tail.

**Tests** (`packages/persona-engine/src/compose/reply.envelope.test.ts`, +236 lines, 12 cases):
- search_codex top-1 grail with image returns the candidate
- empty array → null
- non-grail top-1 + grail-with-image second hit → returns grail (scan-all)
- all hits non-grail (no image fields) → null gracefully
- `image_url` alt key works
- legacy `{results:[...]}` shape returns null (defensive — documents the shape change)
- lookup_grail flat envelope still works (PR #21 path preserved)
- lookup_grail nested `{result:{...}}` still works
- lookup_grail with no image returns null
- null safety: null / undefined / string parsed all return null

## B2 — LLM pastes image URL inline → automod deletes message

**Symptom**: operator's Call A (lookup_grail × 2) had `attached=1` (bytes attached via webhook) but the Discord message DISAPPEARED. Same blocklist failure as the original V0.7-A.3 design problem.

**Root cause**: the LLM saw `image: "https://assets.0xhoneyjar.xyz/Mibera/grails/black-hole.png"` in the lookup_grail tool result envelope and naturally pasted that URL inline in voice prose. Discord automod still triggered on the URL even though bytes were attached. Bytes-on-the-wire contract defeated by URL-paste residual.

### B2a — composer-side strip (defense-in-depth, primary)

`packages/persona-engine/src/deliver/strip-image-urls.ts` (new, +113 lines):

`stripAttachedImageUrls(text, attachedUrls)` removes 3 forms:
- markdown image: `alt` → entire construct removed
- markdown link: `[text](url)` → entire construct removed (image attached separately makes link redundant)
- plain URL paste: bare URL string removed

Plus whitespace cleanup: collapse multi-space, drop orphan punctuation adjacent to removed URL, collapse 3+ newlines to 2 (preserves paragraph breaks). **Only strips URLs in the supplied list** — non-attached references preserved.

Wired in `composeReplyWithEnrichment`:
- Derives attached URL list from `grailCandidates` (slice to actual `payload.files.length` so failed-fetch candidates don't false-strip)
- Re-chunks stripped text via `splitForDiscord` (chunk boundaries shift)
- Updates `payload.content`, `result.content`, `result.chunks` consistently
- Telemetry: `[embed-with-image] stripped N inline image URL(s) from reply text (character=...)` — operators confirm when LLM was leaking URLs and the safety net caught it

**Tests** (`strip-image-urls.test.ts`, +191 lines, 21 cases): plain URL inline / start / end · markdown image alt+empty+mid-paragraph · markdown link with text+inline-sentence · preservation: non-attached URL preserved + empty list no-op · multi-paragraph structure preserved · multiple URL forms together · whitespace cleanup edge cases · `extractAttachedUrls` helper coverage

### B2b — persona instruction reinforcement

Both `apps/character-{ruggy,satoshi}/persona.md` updated with a new bullet under the existing CODEX GROUNDING section. Register-matched per character:

**ruggy** (lowercase / OG / festival-NPC):
> never paste image urls in your reply text. when codex returns `image: '...'` in a tool result, the substrate attaches the image bytes directly... reference grails by `@g<id>` (e.g. `@g876`) only — the image follows automatically. URLs in voice text get caught by Discord automod and your whole reply disappears. ngl that's not the vibe.

**satoshi** (sentence-case / gnomic-cypherpunk):
> Never paste image URLs in your reply text. When codex returns `image: '...'` in a tool result, the substrate attaches the image bytes to your reply directly... Reference grails by `@g<id>` (e.g. `@g876`) only — the image follows automatically. URLs inline in voice trigger Discord automod and the entire reply disappears. The cypherpunk observes the channel; the channel's filters are also part of the protocol.

### Why both layers

Single-layer options considered:
- **persona-only** — fragile (LLM ignores rule under load · register may rationalize URL as in-character "show the link")
- **composer-only strip** — silent fix, no learning signal back to LLM

Both shipped: persona addresses the LLM-side habit; composer strip is the belt-and-braces with telemetry that surfaces when the persona instruction failed and the safety net caught it.

## Test additions

Total: +33 new tests across 2 files. **All 161/161 persona-engine tests pass** (was 128 baseline). **All 46/46 apps/bot tests pass** (no regressions).

```
B1:  packages/persona-engine/src/compose/reply.envelope.test.ts       (12 tests)
B2a: packages/persona-engine/src/deliver/strip-image-urls.test.ts     (21 tests)
```

## Files changed

| File | Δ | What |
|---|---|---|
| `packages/persona-engine/src/compose/reply.ts` | +85/-15 | B1: `pickFirstGrailFromEnvelope` rewrite for `search_codex` direct-array + scan-all + telemetry. B2a: wire strip+rechunk in `composeReplyWithEnrichment` |
| `packages/persona-engine/src/compose/reply.envelope.test.ts` | +236 | B1 tests (new) |
| `packages/persona-engine/src/deliver/strip-image-urls.ts` | +113 | B2a helper (new) |
| `packages/persona-engine/src/deliver/strip-image-urls.test.ts` | +191 | B2a tests (new) |
| `packages/persona-engine/src/index.ts` | +10 | export `stripAttachedImageUrls` + `extractAttachedUrls` |
| `apps/character-ruggy/persona.md` | +8 | B2b instruction (lowercase register) |
| `apps/character-satoshi/persona.md` | +10 | B2b instruction (sentence-case register) |
| `grimoires/loa/NOTES.md` | +2 | Decision Log: B1 envelope-shape fix + B2 dual-layer rationale |

## Verify gates

- [x] `bun run typecheck` GREEN (workspace)
- [x] `bun run --cwd packages/persona-engine test` GREEN (161/161, +33 new, no regressions)
- [x] `bun run --cwd apps/bot test` GREEN (46/46, no regressions)

## Doctrine validation gate G-6 (operator-bounded)

This hotfix RESETS PR #21's G-6 KEEPER pass on the 2nd server with `assets.0xhoneyjar.xyz` blocklisted. Operator should re-test:
- Call A: `lookup_grail` × 2 → bytes attached + URL absent from voice text ✓
- Call B: `search_codex` "the dark grail" → SearchHit[] parsed correctly + grail image attached ✓
- Both: no inline `https://assets.0xhoneyjar.xyz/...` in displayed message text

## Out of scope (deferred from prior bridgebuilder)

F2 (session-ref regex asymmetry · V1.5 cleanup), F7 (Content-Type from URL ext · low impact), F8 (PATCH fallback drops attachments · DESIGN-intent), F10 (footer chunking moot post-F4), F11 (non-text content blocks dropped · log addition sufficient). **F9 (search_codex top-1 ordering) lands here** as part of B1 scan-all-results — the bug being fixed *required* solving F9 to be correct.


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#22. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "search_codex envelope shape correction with scan-all fallback",
      "severity": "PRAISE",
      "category": "correctness",
      "description": "The fix correctly addresses two bugs at once: (1) the envelope shape was `{results: [...]}` but the actual codex MCP returns a top-level `SearchHit[]` array, and (2) top-1-only matching dropped grails when a non-grail hit ranked higher. The scan-all-with-image-field approach is the right minimal correction. Clear inline comments document the upstream contract source.",
      "suggestion": "Keep this pattern. Consider adding an integration test that pins the actual codex MCP response shape via a fixture file to detect future upstream regressions."
    },
    {
      "id": "F2",
      "title": "Comprehensive test coverage for envelope variants and edge cases",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/persona-engine/src/compose/reply.envelope.test.ts",
      "description": "Tests cover the new direct-array shape, the legacy wrapper (now-wrong) shape as a defensive regression marker, lookup_grail flat + nested envelopes, scan-all behavior with non-grail top hit, image_url alt key, and null-safety. The legacy-shape test that documents 'this used to work, now returns null' is particularly useful for future readers."
    },
    {
      "id": "F3",
      "title": "Markdown link removal drops link text entirely — may strip semantic content",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/deliver/strip-image-urls.ts",
      "description": "The markdown-link branch (`[text](url)`) removes the entire `[text](url)` construct including the visible text. For markdown images this is correct (alt text is invisible), but for `[the void](url)` the visible text 'the void' is part of the voice prose and gets deleted. Test on line 70-74 confirms this: `[the void](URL) is the dark grail.` becomes `is the dark grail.` — losing the noun phrase.",
      "suggestion": "Consider preserving the link text for markdown links: replace `\\[([^\\]]*)\\]\\(${escaped}\\)` with `$1` instead of empty string. Images can still be fully removed since alt text is non-visible. The current behavior may be intentional (the LLM's link text is usually redundant with surrounding prose), but it warrants a comment explaining the trade-off.",
      "confidence": 0.6
    },
    {
      "id": "F4",
      "title": "extractAttachedUrls slice may misalign with files when candidates are deduplicated downstream",
      "severity": "LOW",
      "category": "correctness",
      "description": "Line ~302: `extractAttachedUrls(grailCandidates).slice(0, payload.files.length)` assumes the first N candidates correspond to the N attached files. If composeWithImage deduplicates candidates by URL/ref or skips some due to download failures, the URLs stripped may not match the URLs actually attached. The strip would then leave attached URLs in text and/or strip URLs that weren't attached.",
      "suggestion": "Have composeWithImage return the list of URLs actually attached (e.g. `payload.attachedUrls`) and use that directly rather than reconstructing from candidates. Failing that, document the invariant that composeWithImage attaches candidates in order with no dedup/skip.",
      "confidence": 0.55
    },
    {
      "id": "F5",
      "title": "Punctuation cleanup regex may damage non-orphan punctuation",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/deliver/strip-image-urls.ts",
      "description": "Line ~89: `out.replace(/\\s+([.,;:!?])/g, '$1')` runs unconditionally over all whitespace+punctuation pairs, not just gaps adjacent to removed URLs. If the original LLM text contained ' .' or ' ,' (e.g. an em-dash construction or stylistic spacing), this will silently mutate it. For Ruggy's lowercase-vibes voice this could matter (e.g. `'word , word'` → `'word, word'`).",
      "suggestion": "Either scope the punctuation cleanup to only run when a URL was actually removed (track a boolean), or accept the global cleanup as a benign normalization and document it. The current docstring says 'orphan punctuation immediately adjacent to a removed URL' but the regex is global.",
      "confidence": 0.5
    },
    {
      "id": "F6",
      "title": "Plain URL replacement does not handle URLs followed by punctuation as part of URL",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/deliver/strip-image-urls.ts",
      "description": "The plain URL replacement (`new RegExp(escaped, 'g')`) is exact-match. If the LLM pastes `https://assets.0xhoneyjar.xyz/.../black-hole.png?v=2` (with query string) or `https://assets.0xhoneyjar.xyz/.../black-hole.png.` (URL adjacent to a period that auto-detection might consider part of URL), the strip will fail to match. Discord/Markdown URL parsers often include trailing punctuation differently.",
      "suggestion": "Consider whether trailing query strings or fragments are possible from the codex envelope. If image URLs are stable canonical paths (no query), current behavior is fine — add a comment confirming this. Otherwise extend the pattern to tolerate optional `?...` and `#...` suffixes.",
      "confidence": 0.4
    },
    {
      "id": "F7",
      "title": "Telemetry warning may be noisy for legitimate non-grail searches",
      "severity": "LOW",
      "category": "operability",
      "description": "The `[embed-with-image] search_codex returned N hits but none had image field` warning fires whenever search_codex returns non-grail results (e.g. core-lore-only queries). This may be expected behavior, not a regression. The warning could create noise that obscures real envelope-shape regressions.",
      "suggestion": "Consider downgrading to debug log, or only warn when the search query was grail-intent (if that signal is available). Alternatively keep at warn level but add 'this is normal for non-grail searches' to the message.",
      "confidence": 0.5
    },
    {
      "id": "F8",
      "title": "imageBearing counter is computed but only used for branch logic",
      "severity": "LOW",
      "category": "code-quality",
      "description": "The `imageBearing` variable in `pickFirstGrailFromEnvelope` is incremented in the loop but only checked once at the end (`if (imageBearing === 0)`). Since the early return inside the loop triggers when image is found, by the time we reach the warning line `imageBearing` is always 0 (we'd have returned early if any candidate had an image). The counter is dead code in its current form.",
      "suggestion": "Remove `imageBearing` and just emit the warning unconditionally after the loop, since reaching the post-loop code already implies no image-bearing grail was found. Or, if the intent was to count hits-with-image-but-not-coercible-as-grail, the counter's increment condition needs refinement.",
      "confidence": 0.85
    },
    {
      "id": "F9",
      "title": "Re-chunking after strip: chunk array reference equality lost",
      "severity": "PRAISE",
      "category": "correctness",
      "description": "Correctly recognizes that text mutation invalidates the original chunks and re-runs splitForDiscord. This avoids the subtle bug of returning stripped content with stale pre-strip chunk boundaries that downstream Discord delivery would use."
    },
    {
      "id": "F10",
      "title": "payload.content mutation alongside finalContent — dual sources of truth",
      "severity": "MEDIUM",
      "category": "code-quality",
      "description": "The hotfix mutates `payload.content = stripped` after composeWithImage returns. This couples the caller to internal payload structure and creates two sources of truth (returned `content` field + payload.content) that must stay in sync. If composeWithImage is later updated to use `content` differently or if another field also encodes the text, a regression is easy.",
      "suggestion": "Either pass the stripped text into composeWithImage as input (preferred — strip before image attachment), or have composeWithImage accept a post-process callback. Mutating its return value out-of-band is fragile. At minimum add a comment in composeWithImage docstring noting that `payload.content` is mutated by callers.",
      "confidence": 0.6
    }
  ]
}
```


## Callouts

PR #28, opened 2026-05-04
fix(compose): translate @g<id> → Mibera #<id> (CMP-boundary finding 2)

zkSoju:
## context

Operator surfaced 4 friction points in WITNESS-shape Discord QA on 2026-05-04. All collapsed to one missing translation layer · crystallized as new vault doctrine `[[chat-medium-presentation-boundary]]`. This PR addresses **finding 2 · @g→display translation**.

## change

Add `translateGrailRefsForChat(text)` post-process to compose/reply.ts. Applied at the END of `composeReplyWithEnrichment` (after grail-ref-guard inspect + stripAttachedImageUrls) so substrate refs persist in `grailRefValidation` and telemetry · only user-visible `content` + `chunks` get translated.

## why

- Operator framing 2026-05-04: "actual NFT number cleanly and the name of the collection, obviously within brackets or within the code area"
- `@g<id>` IS load-bearing for grail-ref-guard hallucination detection — must persist substrate-side
- Translation lives at the chat-medium-presentation-boundary per doctrine
- NFT-style display matches the broader Mibera ecosystem vocabulary

## tests

27 pass / 0 fail · 15 new tests for `translateGrailRefsForChat` + 12 existing envelope tests still green:
- single ref · multiple refs · idempotence · non-grail-shapes-untouched · empty + edge cases

## non-goals (deferred)

- Codex MCP enrichment (`Black Hole · Mibera #876`) → V1.5 / cycle B composition · adds latency for diminishing returns
- Templating literal substitutions (`{{ZONE_ID}}` → `stonehenge`) → separate finding · not addressed here

## refs

- vault doctrine (NEW): `[[chat-medium-presentation-boundary]]` · finding 2
- WITNESS QA evidence: 2026-05-04 6:14PM PT operator dogfood
- composes with `[[discord-native-register]]` · `[[metadata-as-integration-contract]]`

## status

DRAFT — awaiting bridgebuilder review per operator standing rule (agent quota reset after 7:50pm PT).

note: parallel PRs in flight on same package:
- PR #27 `fix/cmp-boundary-webp-prefer` (T3 · finding 4 · grail-cache.ts only · no overlap)
- T3' agent · `fix/cmp-boundary-attachment-citation-order` (finding 3 · TOUCHES same compose/reply.ts file · expect rebase one direction)

PR #27, opened 2026-05-04
fix(deliver): prefer .webp variants (CMP-boundary finding 4)

zkSoju:
## context

Operator surfaced 4 friction points in WITNESS-shape Discord QA on 2026-05-04. All collapsed to one missing translation layer · crystallized as new vault doctrine `[[chat-medium-presentation-boundary]]`. This PR addresses **finding 4 · webp prefer**.

## change

Flip `CANONICAL_GRAIL_URLS` from `.png` → `.webp` for all 7 V1 grails. Pre-flight verified all 7 webp variants exist at canonical paths · sizes 64KB–361KB (~1.2MB total cache vs ~50MB+ as PNGs).

## why

- V07A4 cap-mistune tail concerns (5MB cap blocked 6/7 grails as PNG) → dissolves with webp
- Boot prefetch faster · cache holds less memory
- Discord renders WebP natively · zero visual change to user
- Composes with [[metadata-as-integration-contract]] (canonical URL stable · variant resolved at presentation-time)
- Bridges to Cycle B asset-pipeline AssetService.fetchOptimal architecture

## tests

53/53 pass · typecheck clean. Updated:
- `grail-cache.test.ts` canonical-list assertion
- `embed-with-image.test.ts` canonical-CDN fixtures + filename + contentType

## refs

- vault doctrine (NEW): `[[chat-medium-presentation-boundary]]` · finding 4
- V07A4 hotfix history: PR #26 (cap mistune) + 03ca840 (mercury → satoshi-as-hermes)
- Cycle B asset-pipeline kickoff: `~/bonfire/grimoires/bonfire/specs/freeside-asset-pipeline-substrate-extraction-kickoff-2026-05-03.md`

## status

DRAFT — awaiting bridgebuilder review per operator standing rule (agent quota reset after 7:50pm PT).

PR #29, opened 2026-05-04
fix(compose): attach by voice citation order (CMP-boundary finding 3)

zkSoju:
## context

Closes the last of the 4 CMP-boundary findings surfaced in 2026-05-04 WITNESS QA. PRs #27 (webp · finding 4) and #28 (@g translation · finding 2) merged earlier · channel-id env vars set (finding 1) · this PR addresses **finding 3 · attachment-citation-order**.

## the bug

Operator dogfood (2026-05-03 6:14PM PT screenshot):
```
prompt:    /satoshi prompt:"the dark grail"
voice:     "no 'dark' grail straight up, but closest read is Black Hole (@g876)"
attached:  Mibera/grails/fire.png  ← WRONG (warm-orange figure with yellow orb)
expected:  Mibera/grails/black-hole.png  ← the void
```

Root cause: `search_codex` returned `[Fire 0.88, Black Hole 0.92]` ranked by relevance · `composeWithImage` attached `candidates[0]` (Fire). Voice authoritatively cited Black Hole. Image-voice misalignment.

## change

Add `orderAttachmentsByCitation(text, candidates)` helper · apply BEFORE `composeWithImage` so `candidates[0]` matches the first @g ref in voice text.

```typescript
const grailCandidates = orderAttachmentsByCitation(
  inspected.text,
  extractGrailCandidates(captured),
);
const initialPayload = await composeWithImage(inspected.text, grailCandidates);
```

## tests

40/40 compose-suite tests pass · 13 new tests for `orderAttachmentsByCitation` including EXACT reproduction of the 2026-05-03 operator-dogfood bug case.

## non-goals

Multi-attach (maxAttachments>1) stays parked in V0.7-A.5 cycle. This PR extracts ONE helper from the V0.7-A.5 SDD §4.2 spec.

## refs

- vault doctrine: `[[chat-medium-presentation-boundary]]` finding 3
- WITNESS QA evidence: 2026-05-04 6:14PM PT operator dogfood
- V0.7-A.5 SDD §4.2 (still parked · this is the standalone extraction)
- composes with PRs #27 (webp · finding 4) + #28 (@g translation · finding 2) + T1 env config (finding 1) · CMP-boundary fully closed

### DIFF
--- a/packages/persona-engine/src/compose/order-attachments-by-citation.test.ts
+++ b/packages/persona-engine/src/compose/order-attachments-by-citation.test.ts
@@ -116,6 +116,38 @@ describe('orderAttachmentsByCitation · uncited candidates go last', () => {
   });
 });
 
+describe('orderAttachmentsByCitation · digit-prefix collision (bridgebuilder PR #29 MED)', () => {
+  test('@g876 candidate is UNCITED when text only contains @g8761 (longer id with same prefix)', () => {
+    // Bridgebuilder finding `ref-substring-collision`: bare `text.indexOf('@g876')`
+    // would match position 12 inside '@g8761' (same prefix). Negative-lookahead
+    // `@g876(?!\d)` correctly says "not cited" because the next char IS a digit.
+    const G8761: CodexGrailResult = {
+      ref: '@g8761',
+      name: 'Some Long-Id Grail',
+      image: 'https://assets.0xhoneyjar.xyz/Mibera/grails/some-long.webp',
+    };
+    const text = 'voice cites @g8761 only';
+    const result = orderAttachmentsByCitation(text, [BLACK_HOLE, G8761]);
+    // G8761 cited at pos 12 → first · BLACK_HOLE uncited → second
+    // (pre-fix would put BLACK_HOLE first because indexOf('@g876') = 12 ties G8761
+    // and stable-sort preserves input order [BLACK_HOLE, G8761])
+    expect(result.map((c) => c.ref)).toEqual(['@g8761', '@g876']);
+  });
+
+  test('both prefix-overlapping refs cited · each matches its own occurrence', () => {
+    const G8761: CodexGrailResult = {
+      ref: '@g8761',
+      name: 'Long-Id',
+      image: 'https://assets.0xhoneyjar.xyz/Mibera/grails/x.webp',
+    };
+    // text cites @g876 at pos 5, @g8761 at pos 18
+    const text = 'cite @g876 then @g8761 too';
+    const result = orderAttachmentsByCitation(text, [G8761, BLACK_HOLE]);
+    // BLACK_HOLE first (cited earlier at pos 5) · G8761 second (pos 18)
+    expect(result.map((c) => c.ref)).toEqual(['@g876', '@g8761']);
+  });
+});
+
 describe('orderAttachmentsByCitation · edge cases', () => {
   test('empty candidates returns empty array', () => {
     const result = orderAttachmentsByCitation('any text', []);
--- a/packages/persona-engine/src/compose/order-attachments-by-citation.test.ts
+++ b/packages/persona-engine/src/compose/order-attachments-by-citation.test.ts
@@ -1,0 +1,148 @@
+/**
+ * orderAttachmentsByCitation tests · CMP-boundary 2026-05-04 finding 3.
+ *
+ * Verifies that grail candidates are sorted by their first-citation position
+ * in voice text BEFORE composeWithImage attaches. With default maxAttachments=1,
+ * candidates[0] wins · ordering matters or the user sees image-text mismatch.
+ *
+ * Operator dogfood evidence (2026-05-03 6:14PM PT):
+ *   prompt: /satoshi prompt:"the dark grail"
+ *   search_codex returned: [Fire 0.88, Black Hole 0.92]
+ *   voice text cited: Black Hole (@g876)
+ *   wrong outcome (pre-fix): Fire image attached (candidates[0] = Fire)
+ *   fix: orderAttachmentsByCitation moves Black Hole to slot[0] because
+ *        @g876 appears before @g6458 in the voice text (only @g876 cited).
+ *
+ * Per [[chat-medium-presentation-boundary]] doctrine: substrate-correct
+ * candidate retrieval (search_codex by relevance) needs presentation-time
+ * reordering to align with what the LLM actually voiced.
+ */
+
+import { describe, test, expect } from 'bun:test';
+import { orderAttachmentsByCitation } from './reply.ts';
+import type { CodexGrailResult } from '../deliver/embed-with-image.ts';
+
+const FIRE: CodexGrailResult = {
+  ref: '@g6458',
+  name: 'Fire',
+  image: 'https://assets.0xhoneyjar.xyz/Mibera/grails/fire.webp',
+};
+const BLACK_HOLE: CodexGrailResult = {
+  ref: '@g876',
+  name: 'Black Hole',
+  image: 'https://assets.0xhoneyjar.xyz/Mibera/grails/black-hole.webp',
+};
+const HERMES: CodexGrailResult = {
+  ref: '@g4488',
+  name: 'Satoshi-as-Hermes',
+  image: 'https://assets.0xhoneyjar.xyz/Mibera/grails/satoshi-as-hermes.webp',
+};
+const SCORPIO: CodexGrailResult = {
+  ref: '@g235',
+  name: 'Scorpio',
+  image: 'https://assets.0xhoneyjar.xyz/Mibera/grails/scorpio.webp',
+};
+
+describe('orderAttachmentsByCitation · the operator-dogfood bug case', () => {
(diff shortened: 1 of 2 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#29. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "ref-substring-collision",
      "title": "indexOf on bare ref can match longer ref as prefix",
      "severity": "MEDIUM",
      "category": "correctness",
      "description": "orderAttachmentsByCitation uses text.indexOf(ref) where ref is shaped like '@g876'. If the voice text cites '@g8761' (or any longer id starting with 876), '@g876' will match as a prefix substring at the same position, producing a false-positive citation. Given grail ids appear to be variable-length numeric strings, this collision is plausible and would silently misorder attachments — exactly the failure mode this fix is intended to prevent.",
      "suggestion": "Match with a boundary-aware regex like `new RegExp(escapedRef + '(?!\\\\d)')` and use the resulting match index, or scan all `@g\\d+` occurrences in the text once and look up by exact ref. Add a test where candidates include @g876 and the voice text only cites @g8761 to lock this in.",
      "confidence": 0.85
    },
    {
      "id": "type-narrowing-of-ref",
      "title": "ref typed as string in CodexGrailResult but defensively checked",
      "severity": "LOW",
      "category": "code-quality",
      "description": "The function checks `typeof ref !== 'string' || ref.length === 0` even though the test file types `CodexGrailResult.ref` as optional string. The defensive check is fine, but if `ref` is declared as `string | undefined` in the type, `typeof ref !== 'string'` is the only needed guard; the empty-string check is dead unless callers pass empty strings.",
      "suggestion": "Either drop the empty-string clause or document the case it covers. Minor.",
      "confidence": 0.5
    },
    {
      "id": "test-coverage-praise",
      "title": "Strong test coverage including operator-dogfood reproduction",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/persona-engine/src/compose/order-attachments-by-citation.test.ts",
      "description": "The test file pins down the exact production failure (Fire vs Black Hole), covers stable-sort semantics for uncited entries, immutability of input, repeated-ref handling, and missing-ref edge cases. The bug-reproduction test directly references the operator dogfood incident, making the regression intent unambiguous.",
      "suggestion": "Consider adding a substring-collision test (see ref-substring-collision finding) to harden the suite further."
    },
    {
      "id": "missing-call-site-test",
      "title": "No integration test verifying composeReplyWithEnrichment uses the ordered candidates",
      "severity": "LOW",
      "category": "testing",
      "description": "The unit tests cover orderAttachmentsByCitation in isolation, but there is no test in this diff verifying that composeReplyWithEnrichment actually wires the ordered output into composeWithImage and that the final attached image matches the voice citation. A regression where someone inadvertently passes the unordered list back to composeWithImage would not be caught.",
      "suggestion": "Add an integration-style test (or a smoke test) that drives composeReplyWithEnrichment with a captured tool envelope returning [Fire, Black Hole] and voice text citing only Black Hole, asserting the attached URL is Black Hole's image.",
      "confidence": 0.6
    },
    {
      "id": "future-dated-comments",
      "title": "Comments and tests reference 2026-05-03/2026-05-04 dates",
      "severity": "SPECULATION",
      "category": "documentation",
      "description": "Comments and test docstrings cite dates in 2026 ('CMP-boundary 2026-05-04', 'operator dogfood 2026-05-03 6:14PM PT'). If these are intentional internal calendar references they're fine; if they're typos for 2025 they may confuse later readers grepping for incident dates.",
      "suggestion": "Confirm dates are intentional; if not, correct.",
      "confidence": 0.3
    }
  ]
}
```


## Callouts

zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#29. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "praise-substring-collision-fix",
      "title": "Negative-lookahead correctly prevents @g876/@g8761 prefix collision",
      "severity": "PRAISE",
      "category": "correctness",
      "description": "The `findRefPosition` helper uses a negative-lookahead `(?!\\d)` to ensure `@g876` does not match inside `@g8761`. This is a real edge case in a grail-ref namespace where IDs share digit prefixes, and the test suite explicitly covers both directions (cited longer-id only, and both cited).",
      "confidence": 0.95
    },
    {
      "id": "praise-stable-sort-and-immutability",
      "title": "Stable sort with explicit idx tiebreaker and non-mutating contract",
      "severity": "PRAISE",
      "category": "code-quality",
      "description": "The sort comparator explicitly returns `a.idx - b.idx` on position ties rather than relying on JS engine sort stability assumptions, and the function returns a new array rather than mutating input. Both contracts are tested.",
      "confidence": 0.9
    },
    {
      "id": "regex-escape-on-ref-with-at",
      "title": "Regex escape pattern omits `@` and `-` but ref starts with `@` which is not regex-special — fine, but list is incomplete",
      "severity": "LOW",
      "category": "robustness",
      "description": "The escape pattern `/[.*+?^${}()|[\\]\\\\]/g` is the standard set for RegExp escaping and is correct. However, if a future ref format introduces hyphens inside character classes or other contexts, this is fine since none of those chars are special outside character classes. Noting only because the comment claims defensive escaping; the escape set is complete for non-class regex contexts.",
      "suggestion": "No change required. Optional: use a shared `escapeRegExp` utility if one exists in the codebase to centralize this pattern.",
      "confidence": 0.4
    },
    {
      "id": "compile-regex-per-call",
      "title": "RegExp constructed per candidate per call",
      "severity": "LOW",
      "category": "performance",
      "description": "`findRefPosition` constructs a new `RegExp` for every candidate on every call. With typical N≤4 candidates this is negligible, but if multi-attach cycles (V0.7-A.5) increase N or this is called in hot paths, consider building a single combined regex or caching.",
      "suggestion": "Acceptable for current scale. Revisit if multi-attach maxAttachments grows or this gets called in tight loops.",
      "confidence": 0.5
    },
    {
      "id": "ref-shape-assumption",
      "title": "Negative-lookahead `(?!\\d)` assumes ref ends in digits only",
      "severity": "LOW",
      "category": "correctness",
      "description": "The collision guard `(?!\\d)` only blocks digit continuation. If ref shape ever changes to include alphanumeric suffixes (e.g. `@g876a`), `@g876` would still incorrectly match `@g876a`. Current ref shape `@g<digits>` makes this fine today, but the assumption is implicit.",
      "suggestion": "Either document the ref-shape assumption explicitly, or use `(?![\\w])` / a word-boundary `\\b` approach to be format-agnostic. Add a test if ref grammar is expected to evolve.",
      "confidence": 0.55
    },
    {
      "id": "no-test-for-multiple-occurrences-of-same-ref",
      "title": "First-occurrence semantics tested only via Fire-before-BlackHole case",
      "severity": "LOW",
      "category": "testing",
      "file": "packages/persona-engine/src/compose/order-attachments-by-citation.test.ts",
      "description": "The 'same ref repeated' test only exercises one repeated ref (@g876). A test where the same ref appears multiple times AND another candidate is cited between repetitions would more directly verify first-occurrence semantics. Minor gap; current coverage is reasonable.",
      "suggestion": "Optional: add a test like `'@g876 then @g6458 then @g876'` to assert ordering uses position 0, not position of last match.",
      "confidence": 0.4
    },
    {
      "id": "case-sensitivity-not-covered",
      "title": "Regex is case-sensitive; voice-text ref casing not asserted",
      "severity": "SPECULATION",
      "category": "robustness",
      "description": "If the LLM ever capitalizes `@G876` in voice text, the citation will be missed and the candidate will sort to end. Likely not an issue if the prompt enforces lowercase `@g<id>` shape, but the assumption isn't tested.",
      "suggestion": "Either add a lowercase invariant test on voice text, or make the regex case-insensitive (`'i'` flag). Verify with the actual LLM output corpus.",
      "confidence": 0.3
    }
  ]
}
```


## Callouts

zkSoju:
## bridgebuilder pass-2 disposition

✅ Pass-2 verdict: 0 HIGH · 0 MEDIUM · 4 LOW · 1 SPECULATION · 2 PRAISE.

The MED `ref-substring-collision` from pass-1 is CLOSED — explicit PRAISE on the negative-lookahead fix + stable-sort-and-immutability contract.

### LOWs deferred (rationale)

- **regex-escape-on-ref-with-at**: escape pattern is complete for current ref shape (`@g<digits>`). Optional cleanup if a shared `escapeRegExp` utility is added codebase-wide.
- **compile-regex-per-call**: negligible at N≤4 candidates per call. Revisit when V0.7-A.5 multi-attach raises N.
- **ref-shape-assumption** (`(?!\d)` only blocks digit continuation): current ref grammar `@g<digits>` makes this safe. Document assumption in code comment if ref format evolves.
- **no-test-for-multiple-occurrences-of-same-ref**: existing `same ref repeated` test covers first-occurrence semantics; additional case is incremental.

### SPECULATION deferred

- **case-sensitivity-not-covered**: persona prompt at `apps/character-{ruggy,satoshi}/persona.md` enforces lowercase `@g<id>` shape. If LLM drift surfaces uppercase, add `'i'` flag.

All deferrals are non-blocking. Merging.

PR #32, opened 2026-05-04
fix(cmp-boundary): emoji shortcode translation + ruggy webhook username (findings 5+6)

zkSoju:
## context

Operator surfaced 2 more CMP-boundary leaks 2026-05-04 8:36 PT during post-cycle dogfood (`/ruggy "what's xabbu's score"`):

1. \`:ruggy_salute:\` rendered as plain text in Discord · should be silently dropped (hallucinated · not in registry)
2. webhookUsername \"Ruggy\" should be lowercase \"ruggy\"

### emoji translator (`compose/reply.ts`)

New \`translateEmojiShortcodes(text)\` post-process at chat-medium boundary:
- \`:name:\` → \`<:name:id>\` (static) or \`<a:name:id>\` (animated) via existing \`renderEmoji\` helper
- hallucinated \`ruggy_*\`/\`mibera_*\` shortcodes silently dropped
- non-custom shortcodes left alone (preserves \`12:30\` time strings, ascii art)

### webhook

\`character-ruggy/character.json\` · webhookUsername \"Ruggy\" → \"ruggy\"

## tests

14 new emoji-translator tests including exact operator-dogfood hallucination case · 56 total compose-suite tests pass · typecheck clean.

## doctrine

Doctrine \`[[chat-medium-presentation-boundary]]\` now has 6 instances closed across 5 PRs in 24h. Drift signature §2 should grow to include emoji shortcode leak after this PR merges.

## status

Ready for bridgebuilder + merge.

### DIFF
--- a/apps/character-ruggy/character.json
+++ b/apps/character-ruggy/character.json
@@ -12,7 +12,7 @@
   "mcps": ["score", "codex", "emojis", "rosenzu", "freeside_auth"],
   "tool_invocation_style": "Use score for zone-stat questions, digest-shaped factual queries, AND per-wallet OG-factor scorecards (`get_wallet_scorecard` accepts wallet address OR Mibera ID — call when someone asks about a specific wallet's score, dimensions, top factors, or rank). Use codex when someone names an archetype, grail, or factor and you want to ground the reference. Use rosenzu when you describe spatial transitions or zone-to-zone movement, or when you want a fresh read on the room. Use emojis composition for vibe-flag punctuation. Use freeside_auth in BOTH directions: `resolve_wallet` (wallet → handle) when wallet identity surfaces, AND `resolve_handle_to_wallet` / `resolve_mibera_id_to_wallet` (handle/mibera-id → wallet) when someone asks about a NAMED user — chain those reverse-resolvers into `get_wallet_scorecard` so 'what's xabbu's score' or 'MIBERA-XXXX OG breakdown' both surface clean. Default to text; tools augment.",
   "webhookAvatarUrl": "https://raw.githubusercontent.com/0xHoneyJar/freeside-characters/staging/apps/character-ruggy/avatar.png",
-  "webhookUsername": "Ruggy",
+  "webhookUsername": "ruggy",
   "webhookAvatarTarget": "https://assets.0xhoneyjar.xyz/freeside-characters/ruggy/avatar.png — canonical target per SDD §0.2 per-world URL contract; bridge URL above points at the staging branch's avatar.png (operator-uploaded 2026-04-30); swap to canonical when assets.0xhoneyjar.xyz CDN cycle reaches /freeside-characters/",
   "doc": {
     "creativeDirection": "creative-direction.md"
--- a/packages/persona-engine/src/compose/reply.ts
+++ b/packages/persona-engine/src/compose/reply.ts
@@ -444,10 +444,18 @@ export function translateGrailRefsForChat(text: string): string {
  * `ruggy_` prefix) gets dropped.
  */
 const KNOWN_EMOJI_PREFIXES: ReadonlySet<string> = new Set(
-  EMOJIS.map((e) => {
-    const match = e.name.match(/^([a-z]+_)/);
-    return match ? match[1]! : null;
-  }).filter((p): p is string => p !== null),
+  EMOJIS.flatMap((e) => {
+    const name = e.name.toLowerCase();
+    // Underscore-shape (e.g. `ruggy_aaa` → `ruggy_`)
+    const underscoreMatch = name.match(/^([a-z0-9]+_)/);
+    if (underscoreMatch) return [underscoreMatch[1]!];
+    // Bridgebuilder PR #32 pass-3 MED `F1-prefix-derivation`: single-word
+    // entries (e.g. `mibera`, `yetinaut`) contribute `<name>_` as a potential
+    // hallucination prefix · so `:mibera_xyz:` drops even when no `mibera_*`
+    // entry exists yet. Bound by registered names · still data-driven.
+    if (/^[a-z][a-z0-9]*$/.test(name)) return [`${name}_`];
+    return [];
+  }),
 );
 
 export function translateEmojiShortcodes(text: string): string {
@@ -463,8 +471,15 @@ export function translateEmojiShortcodes(text: string): string {
     // Miss + matches a known custom-emoji prefix from registry →
     // custom-emoji-shaped hallucination · drop (also drop the preceding
     // space if one was matched · no whitespace artifact).
+    const lower = name.toLowerCase();
     for (const prefix of KNOWN_EMOJI_PREFIXES) {
-      if (name.toLowerCase().startsWith(prefix)) {
+      if (lower.startsWith(prefix)) {
+        // Bridgebuilder PR #32 pass-3 MED `F5-silent-drop-no-observability`:
+        // emit a debug log per dropped hallucination so operators can see
+        // when LLM emoji drift recurs (signal for persona drift OR registry
+        // gap worth filling). Cardinality bounded by LLM creativity ·
+        // single console.warn line per drop · low log volume in normal use.
+        console.warn(`[emoji-translate] dropped hallucinated shortcode :${name}: (matched prefix '${prefix}' but no registry entry)`);
         return '';
       }
     }
--- a/packages/persona-engine/src/compose/reply.ts
+++ b/packages/persona-engine/src/compose/reply.ts
@@ -52,7 +52,7 @@ import {
   type GrailRefValidation,
 } from '../deliver/grail-ref-guard.ts';
 import { stripAttachedImageUrls } from '../deliver/strip-image-urls.ts';
-import { findByName, renderEmoji } from '../orchestrator/emojis/registry.ts';
+import { EMOJIS, findByName, renderEmoji } from '../orchestrator/emojis/registry.ts';
 import {
   appendToLedger,
   getLedgerSnapshot,
@@ -427,14 +427,30 @@ export function translateGrailRefsForChat(text: string): string {
  * `12:30` collisions; underscore + alphanumeric allowed thereafter to
  * match Discord's emoji name grammar.
  */
+/**
+ * Module-load: derive known custom-emoji prefixes from the registry itself.
+ *
+ * Bridgebuilder PR #32 pass-2 MED `F1-underscore-heuristic-false-positives`:
+ * the previous heuristic ("contains underscore → drop on miss") was
+ * over-eager · it would silently consume legitimate non-emoji shortcodes
+ * like `:my_var:`, `:file_path:`, technical jargon. The fix is
+ * registry-as-source-of-truth: collect prefixes from registered names
+ * (`ruggy_*` and `mibera_*` today · auto-extends as registry grows for
+ * new personas like yetinaut_*, honeybee_*, etc) and only drop on miss
+ * when the shortcode shape matches a KNOWN prefix.
+ *
+ * False positives now bounded by data the operator controls (registry).
+ * `:my_var:` passes through · `:ruggy_salute:` (hallucinated but matches
+ * `ruggy_` prefix) gets dropped.
+ */
+const KNOWN_EMOJI_PREFIXES: ReadonlySet<string> = new Set(
+  EMOJIS.map((e) => {
+    const match = e.name.match(/^([a-z]+_)/);
+    return match ? match[1]! : null;
+  }).filter((p): p is string => p !== null),
+);
+
 export function translateEmojiShortcodes(text: string): string {
-  // Bridgebuilder PR #32 pass-1 MED `F2-prefix-list-hardcoded`: shifted from
-  // hardcoded `ruggy_`/`mibera_` allowlist to underscore-heuristic so new
-  // personas (purupuru-yetinaut · honeyport-bee · future) inherit hallucination-
-  // drop without code change. Custom emoji names conventionally include
-  // underscore (`ruggy_smoke` · `mibera_kiii`); single-word shortcodes
-  // (`:hello:`, `:colon:`, ascii art) lack underscore and pass through.
-  //
   // Bridgebuilder PR #32 pass-1 LOW `F1-whitespace-artifact`: consume one
   // preceding space when dropping a hallucinated shortcode so we don't leak
   // double-spaces / trailing whitespace into Discord output.
@@ -444,12 +460,16 @@ export function translateEmojiShortcodes(text: string): string {
       // Hit · render Discord format · keep the leading space.
       return `${leadingSpace}${renderEmoji(entry)}`;
     }
-    // Miss + underscore in name → custom-emoji-shaped hallucination · drop
-    // (also drop the preceding space if one was matched · no whitespace artifact).
-    if (name.includes('_')) {
-      return '';
+    // Miss + matches a known custom-emoji prefix from registry →
+    // custom-emoji-shaped hallucination · drop (also drop the preceding
+    // space if one was matched · no whitespace artifact).
+    for (const prefix of KNOWN_EMOJI_PREFIXES) {
+      if (name.toLowerCase().startsWith(prefix)) {
+        return '';
+      }
     }
-    // Miss + no underscore → not custom-emoji-shaped · pass through unchanged.
+    // Miss + not a known-prefix shape → not a custom-emoji-shaped hallucination ·
+    // pass through unchanged (covers `:my_var:`, `:file_path:`, technical jargon).
     return match;
   });
 }
--- a/packages/persona-engine/src/compose/reply.ts
+++ b/packages/persona-engine/src/compose/reply.ts
@@ -52,6 +52,7 @@ import {
   type GrailRefValidation,
 } from '../deliver/grail-ref-guard.ts';
 import { stripAttachedImageUrls } from '../deliver/strip-image-urls.ts';
+import { findByName, renderEmoji } from '../orchestrator/emojis/registry.ts';
 import {
   appendToLedger,
   getLedgerSnapshot,
@@ -338,7 +339,18 @@ export async function composeReplyWithEnrichment(
   // cleanly and the name of the collection" (2026-05-04 /smol session).
   // Substrate refs persist UNCHANGED in `grailRefValidation` + telemetry —
   // only `content` + `chunks` get the chat-medium translation.
-  const displayContent = translateGrailRefsForChat(finalContent);
+  const grailTranslated = translateGrailRefsForChat(finalContent);
+  // CMP-boundary 2026-05-04 finding 5 (vault doctrine
+  // `[[chat-medium-presentation-boundary]]`): translate raw `:name:` emoji
+  // shortcodes into Discord render format `<:name:id>` (or `<a:name:id>`
+  // for animated). LLM commonly emits raw shortcode form (matches persona
+  // doc + meme convention) but Discord renders raw `:name:` as plain text.
+  // Operator dogfood `/ruggy "what's xabbu's score"` 2026-05-04 8:36 PT
+  // surfaced this — `:ruggy_salute:` rendered as text.
+  // Side effect: hallucinated names (`ruggy_salute` not in registry) are
+  // silently dropped per persona's `fall back to bear emoji` convention,
+  // preventing fake-emoji text leaks.
+  const displayContent = translateEmojiShortcodes(grailTranslated);
   const displayChunks =
     displayContent === finalContent
       ? finalChunks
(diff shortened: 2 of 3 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#32. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Whitespace artifact left behind when shortcode is dropped",
      "severity": "LOW",
      "category": "correctness",
      "description": "When a hallucinated shortcode like `:ruggy_salute:` is silently dropped, the surrounding spaces remain. The test `'dude\\'s on the map :ruggy_salute:'` asserts the result is `'dude\\'s on the map '` (trailing space), and `'honey moment :mibera_sparkle: yes'` becomes `'honey moment  yes'` (double space). This produces visible artifacts in chat output — trailing whitespace and runs of double spaces — that the operator will see in Discord.",
      "suggestion": "Either trim the whitespace adjacent to the dropped match (consume one leading space) or post-process to collapse double spaces and trim trailing whitespace before sending. Simpler: extend the regex to optionally consume one preceding space, e.g. ` ?:([a-zA-Z][a-zA-Z0-9_]*):`, and only consume it on drop.",
      "confidence": 0.85
    },
    {
      "id": "F2",
      "title": "Hallucination drop list is hardcoded to two prefixes",
      "severity": "MEDIUM",
      "description": "The drop heuristic is hardcoded to `ruggy_` and `mibera_` prefixes. As new personas/worlds are added (the registry already holds 43 emojis from THJ guild), this list will silently fail to drop hallucinations for new prefixes, leaking raw `:foo_bar:` text. The function lives in shared compose code but encodes character-specific knowledge.",
      "suggestion": "Derive the set of known custom-emoji prefixes from the registry itself (e.g. unique prefixes among registered names) or maintain the prefix list alongside the registry as a single source of truth. Alternatively, drop ANY unknown shortcode that contains an underscore (typical custom-emoji marker) while keeping single-word shortcodes intact.",
      "confidence": 0.7
    },
    {
      "id": "F3",
      "title": "Translation runs after grail-ref guard but isn't covered by validation telemetry",
      "severity": "LOW",
      "category": "observability",
      "description": "Hallucinated shortcodes are dropped silently without any counter, log, or telemetry signal. Per the comment, this is intentional (mirrors the bear-emoji fallback), but it makes regression of LLM emoji hallucination invisible. If hallucination rate spikes, there's no signal.",
      "suggestion": "Emit a debug-level log or increment a counter when a custom-prefix shortcode is dropped, including the persona/character context. Useful for tracking which hallucinated names recur and could be added to the registry.",
      "confidence": 0.6
    },
    {
      "id": "F4",
      "title": "Regex does not handle adjacent shortcodes correctly in all cases",
      "severity": "LOW",
      "category": "correctness",
      "description": "The pattern `:([a-zA-Z][a-zA-Z0-9_]*):` is greedy on the inner capture but anchored on colons. For text like `:a:b:` it would match `:a:` first (correct), but for chained tokens with no spaces (`:ruggy_dab::ruggy_cheers:`) the match works because of non-overlapping greedy left-to-right. However, no test case covers adjacent shortcodes — worth a unit test.",
      "suggestion": "Add a test for adjacent shortcodes (`:ruggy_cheers::ruggy_dab:`) to lock in behavior.",
      "confidence": 0.4
    },
    {
      "id": "F5",
      "title": "webhookUsername case change has no test or migration note",
      "severity": "LOW",
      "category": "consistency",
      "file": "apps/character-ruggy/character.json",
      "description": "Changing `webhookUsername` from `Ruggy` to `ruggy` is an externally visible identity change in Discord. The PR title labels this as 'finding 6' but there is no test, snapshot, or operator-visible note in the diff explaining why lowercase is canonical (vs. capitalized as other personas may use). Risk of inconsistency across personas if this is character-specific styling rather than a global convention.",
      "suggestion": "Document the rationale (e.g. 'persona name is lowercase per voice doc' or 'matches handle convention') in a comment field or commit message, and confirm other character.json files in the repo follow the same convention or are intentionally different.",
      "confidence": 0.5
    },
    {
      "id": "F6",
      "title": "Good test coverage for happy path, hallucination drop, and edge cases",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/persona-engine/src/compose/translate-emoji-shortcodes.test.ts",
      "description": "Tests cover real registry entries (static + animated), hallucination drop case with exact operator-dogfood reproduction, non-custom shortcodes left alone (time `12:30:00`, generic `:hello:`), uppercase variant pass-through, position variations (start/end/middle), and mixed scenarios. Clear comments tying tests back to the operator screenshot give strong regression value.",
      "suggestion": "No change required."
    },
    {
      "id": "F7",
      "title": "Function correctly preserves substrate truth at presentation boundary",
      "severity": "PRAISE",
      "category": "design",
      "description": "The translation runs only on `displayContent`/`displayChunks` while substrate (`finalContent`, `grailRefValidation`, telemetry) remains UNCHANGED — correctly applying the chat-medium-presentation-boundary doctrine consistent with `translateGrailRefsForChat`.",
      "suggestion": "No change required."
    },
    {
      "id": "F8",
      "title": "Test relies on registry containing specific emoji names",
      "severity": "LOW",
      "category": "testing",
      "file": "packages/persona-engine/src/compose/translate-emoji-shortcodes.test.ts",
      "description": "Tests assert specific registry contents (`ruggy_cheers` static, `ruggy_dab` animated). If the upstream registry is regenerated and either entry's animated flag flips or the entry is removed, tests will break in non-obvious ways. The test asserts via regex on `\\d+` for the ID which is fine, but the existence and animated-flag assumptions are implicit.",
      "suggestion": "Either mock the registry imports for unit isolation, or add an explicit precondition assertion at the top of the suite that documents/verifies the registry shape the tests depend on.",
      "confidence": 0.6
    }
  ]
}
```


## Callouts

zkSoju (commented):
## Summary

This PR closes the Discord render boundary on emoji shortcodes with surgical precision: substrate stays canonical, presentation gets translated. The hallucination-drop heuristic is the right idea but slightly over-eager, and a few whitespace/test-naming details deserve polish before merge.

The architectural instinct here — translating only `displayContent`/`displayChunks` while leaving substrate refs untouched — is exactly the kind of presentation-vs-domain separation that ages well. The remaining concerns are tactical: the underscore heuristic is a load-bearing regex, and load-bearing regexes deserve fence posts.

## Architectural Meditation

The pattern at play is what Google's monorepo culture calls *"render-time translation"* — the substrate (LLM output, audit log, replay tape) remains in a canonical, tool-agnostic form, and platform-specific encoding happens at the edge. This is how Gmail keeps message bodies in MIME while rendering HTML, and how Stripe keeps amounts in minor units while formatting per-locale at display. Get this boundary right once, and every future platform (Slack, Telegram, web) plugs in without contaminating the core.

The hallucination-drop logic, however, is doing two jobs in one regex: *detecting* a likely-fake shortcode and *cleaning* its whitespace footprint. When a single expression carries two responsibilities, edge cases multiply combinatorially — which is exactly what F1/F2/F3 surface.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1-underscore-heuristic-false-positives",
      "title": "Underscore heuristic silently drops legitimate non-emoji shortcodes",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/compose/reply.ts:447",
      "description": "Any `:foo_bar:` pattern is consumed even when not an emoji — `:my_var:`, file refs, technical jargon. The previous prefix-list (ruggy_/mibera_) was over-restrictive but precise; this heuristic is broader and corrupts legitimate text. Tests don't cover non-emoji underscore-colon patterns.",
      "suggestion": "Derive the drop predicate from the registry itself: collect known prefixes (ruggy_, mibera_, ...) at module load and only drop on miss when the shortcode shape is `<knownPrefix><word>`. This keeps false-positive surface bounded by data you control.",
      "confidence": 0.6,
      "faang_parallel": "Slack's message formatter faced this exact tension with `:emoji:` syntax colliding with code blocks containing colons. They resolved it by deriving allowed shortcode shapes from the workspace's emoji registry rather than syntactic heuristics — registry-as-source-of-truth.",
      "metaphor": "It's a spam filter that blocks every email with the word 'invoice' — catches the bad ones, but also your accountant.",
      "teachable_moment": "When a heuristic is broader than the data it claims to recognize, prefer deriving the heuristic from the data. Registries are cheap; false positives in user-visible text are not."
    },
    {
      "id": "F2-leading-space-only-one",
      "title": "Leading-space consumption only handles single space",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/compose/reply.ts:443",
      "description": "`( ?)` consumes one space. `text  :ruggy_salute:` (double space after punctuation) or `text\\n:ruggy_salute:` leaves a whitespace artifact. Start-of-text drops also leave a trailing space.",
      "suggestion": "Match `(\\s*)` for leading whitespace, or run a final whitespace-collapse pass on segments where a drop occurred. Add tests for double-space, newline-adjacent, and start-of-text scenarios.",
      "confidence": 0.7,
      "metaphor": "Cleaning up after a guest who spilled wine — you wiped the obvious puddle but missed the splash on the wall.",
      "teachable_moment": "String surgery should always have a normalization pass downstream. The cut and the suture are separate concerns."
    },
    {
      "id": "F3-trailing-space-not-handled",
      "title": "Trailing whitespace artifact when dropped shortcode has no preceding space",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/compose/reply.ts:443",
      "description": "Input `:ruggy_salute: text` matches with empty leadingSpace, leaves ` text` with leading space. Tests only cover trailing-position drops.",
      "suggestion": "When leadingSpace is empty, optionally consume one trailing space; or add a final `.replace(/\\s+/g, ' ').trim()` pass on display strings.",
      "confidence": 0.75,
      "connection": "Pairs with F2 — both dissolve under a single normalization-pass solution."
    },
    {
      "id": "F4-test-comment-mismatch",
      "title": "Test title references ruggy_smoke but body uses ruggy_cheers",
      "severity": "LOW",
      "category": "documentation",
      "file": "packages/persona-engine/src/compose/translate-emoji-shortcodes.test.ts:25",
      "description": "Test title says `ruggy_smoke translates...` but the body asserts on `ruggy_cheers`. Misleading for future maintainers grepping by emoji name.",
      "suggestion": "Rename the test to match the body: `static emoji ruggy_cheers translates to <:name:id> form`.",
      "confidence": 0.95,
      "teachable_moment": "Test names are the API documentation of behavior. A mismatch is a broken doc-link in your safety net."
    },
    {
      "id": "F5-webhook-username-case-change-rationale",
      "title": "Webhook username case change lacks documented rationale",
      "severity": "LOW",
      "category": "documentation",
      "file": "apps/character-ruggy/character.json:15",
      "description": "`Ruggy` → `ruggy` is bundled with the emoji fix but unjustified inline. Likely persona-tone alignment, but unverifiable without context.",
      "suggestion": "Add a one-line note in the PR description or commit body: 'lowercase matches ruggy's persona voice convention.' Future agents reviewing webhook config drift will thank you.",
      "confidence": 0.4,
      "teachable_moment": "Decision Trail: every casing/naming change should leave a breadcrumb. The cost is one sentence; the benefit is preventing a future revert PR."
    },
    {
      "id": "F6-doctrine-alignment",
      "title": "Clear separation of substrate vs presentation data",
      "severity": "PRAISE",
      "category": "architecture",
      "file": "packages/persona-engine/src/compose/reply.ts:342",
      "description": "Translation applied only to displayContent/displayChunks; substrate refs and telemetry untouched. The comment explicitly names the boundary doctrine.",
      "faang_parallel": "Mirrors Stripe's API design: amounts stored in minor units (substrate), formatted per-locale at the edge (presentation). One source of truth, many renderings.",
      "metaphor": "The screenplay stays in English; the subtitles get translated. You can always re-dub."
    },
    {
      "id": "F7-test-coverage-completeness",
      "title": "Comprehensive test suite covers happy path, hallucinations, and edge cases",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/persona-engine/src/compose/translate-emoji-shortcodes.test.ts:1",
      "description": "Static + animated registry hits, hallucinated drops with operator-dogfood regression anchor, time-string false-positive avoidance, position variants, repetition, mixed scenarios.",
      "suggestion": "Add the F2/F3 edge cases (double-space, start-of-text drop, newline-adjacent) to close the matrix.",
      "teachable_moment": "Preserving the operator-dogfood reproduction as a named test is exemplary — bug provenance encoded as executable history."
    },
    {
      "id": "F8-uppercase-test-correctness",
      "title": "Uppercase-prefix test depends on findByName case-sensitivity",
      "severity": "SPECULATION",
      "category": "correctness",
      "file": "packages/persona-engine/src/compose/translate-emoji-shortcodes.test.ts:79",
      "description": "`:Ruggy_smoke:` → empty assumes case-sensitive findByName + underscore drop. The contract is implicit.",
      "suggestion": "Add a direct unit test for findByName case behavior, or add a comment on the test explaining the dependency chain.",
      "confidence": 0.5
    }
  ]
}
```


## Callouts

The substrate/presentation split (F6) and the regression-anchor test discipline (F7) are the kind of details that mark mature codebases. The instinct to drop hallucinated shortcodes rather than render `:ruggy_salute:` literally to users shows real empathy for the operator experience — most teams ship the literal and patch later.

Tighten F1's heuristic, normalize whitespace once (F2/F3), fix the test name (F4), and this is a clean merge.

Verdict: **REQUEST_CHANGES** — primarily for F1, which is a user-visible correctness risk in any text containing underscore-bearing colon syntax (code snippets, technical chat).

zkSoju (commented):
## Summary

This PR plugs two operational paper cuts: hallucinated emoji shortcodes leaking into Discord output, and a casing inconsistency in Ruggy's webhook identity. The registry-derived prefix approach is genuinely well-designed, and the test matrix is thoughtful. The main gaps are around observability (silent drops blind operators to LLM drift) and a few edge cases in the regex that will surface as the registry grows beyond the current `<prefix>_<name>` shape.

The architectural pattern here — treating the LLM as an untrusted producer and the registry as the trust boundary — is exactly the right mental model. It mirrors how Google's safe-HTML libraries treat template output: the model can suggest, but only the registry can authorize. The remaining work is making that trust boundary observable and resilient to schema drift.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Prefix derivation regex misses single-word emoji names",
      "severity": "MEDIUM",
      "category": "correctness",
      "description": "KNOWN_EMOJI_PREFIXES is built via `e.name.match(/^([a-z]+_)/)`, capturing only names containing an underscore. Single-word entries (e.g. `mibera`, `yetinaut`) contribute no prefix, so a hallucinated `:mibera_xyz:` would only be dropped if some OTHER `mibera_*` entry happens to exist. The invariant 'every persona has at least one underscore-shaped emoji' is implicit and untested.",
      "suggestion": "Add a unit test snapshotting KNOWN_EMOJI_PREFIXES contents so registry-shape changes surface loudly. Optionally fall back to including full names without underscores as exact-match prefixes.",
      "confidence": 0.7,
      "faang_parallel": "Meta's GraphQL schema validators learned this lesson: implicit invariants ('all types have a description') silently regress when new types arrive. They now codify them as schema lint rules. Same principle applies here.",
      "metaphor": "It's like a bouncer who only checks IDs that have a hyphen in the last name. Most guests are fine — until someone named 'Madonna' shows up.",
      "teachable_moment": "When a derivation depends on data shape, either lock the shape with a test or make the derivation shape-agnostic. 'It works because the data happens to look this way' is a future bug with a delivery date."
    },
    {
      "id": "F2",
      "title": "Lowercase-only prefix derivation will miss mixed-case registry entries",
      "severity": "LOW",
      "category": "correctness",
      "description": "The regex `/^([a-z]+_)/` requires lowercase. Discord allows mixed-case emoji names; if the registry ever contains `Ruggy_smoke`, no prefix is derived. The match site does `name.toLowerCase().startsWith(prefix)`, which is asymmetric — input is normalized but the prefix set is not.",
      "suggestion": "Normalize at construction: `e.name.toLowerCase().match(/^([a-z0-9]+_)/)`. Symmetric normalization is cheaper than documentation.",
      "confidence": 0.5,
      "metaphor": "Two clerks, one only reading lowercase, one translating everything to lowercase first. They mostly agree — until they don't."
    },
    {
      "id": "F3",
      "title": "Leading-space consumption only handles single space",
      "severity": "LOW",
      "category": "correctness",
      "description": "`( ?):...` consumes one literal space. Double spaces, tabs, or newlines before a hallucinated shortcode leave whitespace artifacts. Start-of-string case `:ruggy_fake: hi` becomes ` hi` with a leading space.",
      "suggestion": "Expand to `(\\s?)` or add a final `.replace(/[ \\t]{2,}/g, ' ').trim()` pass. Add tests for the start-of-string and double-space cases.",
      "confidence": 0.75,
      "faang_parallel": "Twitter's text rendering pipeline famously had a class of bugs where entity removal left orphaned whitespace — they ended up adding a normalization post-pass for exactly this reason.",
      "metaphor": "Pulling a book off a shelf leaves a gap. You either push the others over, or accept the visual seam."
    },
    {
      "id": "F4",
      "title": "Test for `:Ruggy_smoke:` couples to undocumented findByName case-sensitivity",
      "severity": "LOW",
      "category": "testing",
      "file": "packages/persona-engine/src/compose/translate-emoji-shortcodes.test.ts",
      "description": "The mixed-case test passes via a chain: regex allows capital, findByName misses (case-sensitive), prefix check normalizes via toLowerCase. If findByName ever becomes case-insensitive, this test silently flips meaning — the emoji would render rather than drop.",
      "suggestion": "Add an inline comment documenting the dependency on findByName case-sensitivity, or assert it directly in a separate test.",
      "confidence": 0.4,
      "teachable_moment": "Tests that pass for the wrong reason are worse than missing tests. A comment naming the load-bearing assumption turns an invisible coupling into a visible one."
    },
    {
      "id": "F5",
      "title": "Silent dropping of hallucinated emoji removes operator visibility into LLM drift",
      "severity": "MEDIUM",
      "category": "observability",
      "description": "Dropped shortcodes vanish without a log, counter, or trace. Operators can't detect when the LLM starts inventing new emoji names — which is signal for prompt drift, model regression, or a registry gap worth filling.",
      "suggestion": "Emit a structured debug log or increment a counter on drop: `logger.debug({ event: 'emoji.hallucinated', name }, ...)`. Cardinality is bounded by LLM creativity, but a sampled or rate-limited log is safe.",
      "confidence": 0.7,
      "faang_parallel": "Google's SRE book calls this 'silent data corruption' — the worst kind of failure because it doesn't page anyone. Their answer is the 'four golden signals' applied even to filters: count what you drop.",
      "metaphor": "A spam filter that throws mail away without telling you is indistinguishable from a broken mail server. The log entry is the receipt.",
      "teachable_moment": "Every silent transformation in a pipeline is a future debugging session you can't have. Cheap counters at boundaries pay for themselves the first time something looks weird in production."
    },
    {
      "id": "F6",
      "title": "Registry-derived prefix set computed at module load",
      "severity": "LOW",
      "category": "correctness",
      "description": "KNOWN_EMOJI_PREFIXES is computed once at import. If EMOJIS is ever refreshed at runtime (e.g. from a Discord guild emoji sync), the prefix set goes stale. The doc comment 'auto-extends as registry grows' suggests dynamic intent.",
      "suggestion": "Either confirm EMOJIS is build-time constant and tighten the comment, or memoize lazily keyed on the registry reference.",
      "confidence": 0.4,
      "metaphor": "A printed phone book in a city with new arrivals every week."
    },
    {
      "id": "F7",
      "title": "Webhook username case change — verify no downstream case-sensitive consumers",
      "severity": "LOW",
      "category": "compatibility",
      "file": "apps/character-ruggy/character.json",
      "description": "`Ruggy` → `ruggy` is reasonable for persona consistency, but Discord displays the literal string and any mention-parsing, message-history dedup, or log-aggregation keys could be case-sensitive.",
      "suggestion": "Document the rationale in the commit body and grep for case-sensitive consumers of webhookUsername. Add to a CHANGELOG if user-facing.",
      "confidence": 0.4,
      "teachable_moment": "Identity strings are interfaces. Changing them is an API change even when the type system doesn't notice."
    },
    {
      "id": "F8",
      "title": "Excellent test coverage of hallucination + false-positive matrix",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/persona-engine/src/compose/translate-emoji-shortcodes.test.ts",
      "description": "Coverage spans registry hits (static + animated), hallucination drops, false-positive pass-through (`:my_var:`, `:file_path:`), time-shape collisions (`12:30:00`), and mixed scenarios. The mixed real-and-hallucinated test is a strong regression guard.",
      "faang_parallel": "This is the equivalence-class testing approach Hillel Wayne champions — every category of input has a representative, including the adversarial-but-benign cases. Stripe's idempotency-key tests follow the same shape."
    },
    {
      "id": "F9",
      "title": "Registry-as-source-of-truth avoids brittle hardcoded prefix list",
      "severity": "PRAISE",
      "category": "design",
      "description": "Deriving KNOWN_EMOJI_PREFIXES from EMOJIS rather than `['ruggy_', 'mibera_']` is correct — it auto-extends with new personas and bounds the false-positive surface to operator-controlled data.",
      "metaphor": "The difference between a dictionary that updates itself and one printed in 1987."
    }
  ]
}
```


## Callouts

The trust boundary here is well-placed: LLM output is filtered against a registry the operator owns, not against a regex of 'plausible' emoji names. That's the same posture Google's safe-HTML libraries take — production must be authorized by trusted data, not pattern-matched. The test file's coverage of the false-positive surface (`:my_var:`, `12:30:00`) shows the author thought adversarially about what *real* text looks like, which is rarer than it should be.

The remaining polish — observability on drops, shape-invariant tests for the registry — is the kind of work that pays for itself the first time the LLM starts emitting `:ruggy_yeet:` and someone asks "is this new?"

Verdict: `COMMENT`

zkSoju:
## bridgebuilder disposition · 3 passes

✅ pass-1 MED `F2-prefix-hardcoded` CLOSED (commit 22168ec · underscore-heuristic, then refined)
✅ pass-2 MED `F1-underscore-false-positives` CLOSED (registry-derived prefixes, 3b01be9)
✅ pass-3 MED `F1-prefix-derivation` CLOSED (single-word prefix derivation, d6f0e36)
✅ pass-3 MED `F5-silent-drop-observability` CLOSED (console.warn on drop, d6f0e36)

### LOW + SPECULATION deferred (rationale)

- **F2 lowercase-prefix-derivation**: registry data is lowercase by Discord convention · symmetric `name.toLowerCase()` at match site · acceptable
- **F3 leading-space single-only**: edge case · double-space + tab + newline would leave artifact · operator-paced final pass-through normalization can land separately
- **F4 test-couples-to-findByName-case-sensitivity**: implicit coupling · documented in test comment · would need separate test if findByName changes
- **F6 prefix-set-stale-if-runtime-refresh**: registry currently static (module-load) · if dynamic refresh ships, recompute at refresh

3 passes is enough. Operator's actual friction (`:ruggy_salute:` from dogfood) closed at pass-1. Subsequent passes hardened edge cases and observability. Merging.

PR #48, opened 2026-05-05
feat(cmp-boundary): voice-discipline transforms · cycle-r sprint-1

zkSoju:
## Summary

- ships `stripVoiceDisciplineDrift(text, opts)` post-process at the chat-medium presentation boundary · strips em-dashes · en-dashes · short asterisk roleplay · default-on closing signoffs (digest exempt) · code-block-safe · idempotent · zero opt-out per architect lock A4
- per-persona regression guard `cmp-boundary.test.ts` for ruggy + satoshi + mongolian-stub · CI lint script enforces every `apps/character-*/` ships its own
- persona file audits: ruggy 293 → 0 em-dashes · satoshi 105 → 0 em-dashes · VOICE DISCIPLINE LOCK section added with affirmative-blueprint cadence rules per `[[negative-constraint-echo]]`

Closes voice-discipline drift class at the chat-medium presentation boundary (cmp-boundary §9 staged this session). Preserves satoshi's gumi-locked `*performed-silence*` exception via discriminator (≤30 chars + no period = strip · longer-or-with-period = preserve).

## Trigger

> *"Do not under any circumstances use em dashes — Instant AI feel"* — operator + Eileen Discord 2026-05-04 4:45 PM PT
> *"shorter dialogues, one on ones don't work in this way"* — same source
> *"way too verbose. ruggy never spoke with full on blocks ... this is discord. you are aware of that right."* — jani 2026-04-29 ruggy V0.5-C pushback (commit 44f3a7f · proof point #2 of class)

## Test plan

- [x] 29 cases in `sanitize.test.ts` — em-dash · en-dash · asterisk · closing · code-block · idempotency · edge · escape-regression · composition (29 pass · 0 fail)
- [x] 12 cases in `apps/character-ruggy/cmp-boundary.test.ts` — voice-discipline + ID-leakage class regression (12 pass)
- [x] 14 cases in `apps/character-satoshi/cmp-boundary.test.ts` — incl. 3 explicit performed-silence preservation cases (14 pass)
- [x] 2 cases in `apps/character-mongolian/cmp-boundary.test.ts` — cycle-3 forward-dep stub w/ `@future` marker (2 pass)
- [x] CI lint `scripts/lint-cmp-boundary-tests.sh` — 3/3 character apps verified
- [x] Full repo suite `bun test` — 377 pass · 0 fail · 927 expect() calls (zero existing-test regression)
- [ ] **Operator-bounded post-merge gate**: KEEPER 100-message Discord production sample sweep — em-dash count = 0 in real LLM-generated output · ≥80% one-on-one replies ≤3 sentences (G-1 KPI)

## Files changed

**Modified:** `packages/persona-engine/src/deliver/sanitize.ts` · `embed.ts` · `apps/character-ruggy/persona.md` · `apps/character-satoshi/persona.md`
**New:** `packages/persona-engine/src/deliver/sanitize.test.ts` · `apps/character-{ruggy,satoshi,mongolian}/cmp-boundary.test.ts` · `scripts/lint-cmp-boundary-tests.sh`

## Cycle context

- **Cycle R**: cmp-boundary-architecture (4 sprints · 35 beads tasks)
- **Sprint 1**: R1 voice discipline (this PR · FR-1 + FR-5)
- **Sprint 2**: R2 MediumCapabilities Schema (NEW \`freeside-mediums\` repo · in flight)
- **Sprint 3**: R3 adapter refactor + CLI proof
- **Sprint 4**: R4 NFT-binding seam (gates on `mibera-as-NPC` cycle-3 fill)

## Architect locks honored

- **A4** universal voice discipline · zero opt-out · code-block-safe · idempotent ✓
- **A6** `@future cycle-3` grep-traceable marker (mongolian stub) ✓

## Doctrine

- `~/vault/wiki/concepts/chat-medium-presentation-boundary.md` §9 (voice-discipline drift class · staged this session)
- `~/vault/wiki/concepts/discord-native-register.md` 2026-05-04 PM amendment (em-dash universal · one-on-one cadence affirmative blueprint)

## Quality gates

- ✅ /implement (sprint-1 reviewer.md)
- ✅ /review-sprint APPROVED with 7 noted non-blocking concerns (engineer-feedback.md)
- ✅ /audit-sprint APPROVED - LETS FUCKING GO (auditor-sprint-feedback.md · OWASP Top 10 review · ReDoS-resistant 6/6 regex)
- ⏳ Operator-bounded KEEPER 100-message sweep post-merge

## Known limitations (documented · non-blocking)

1. CI workflow file not created — repo has no `.github/` directory · lint script invocable from CLI; operator-paced CI wiring
2. CLAUDE.md persona-file canonical-sync convention requires downstream sync to codex grails (operator-paced)
3. Bulk perl em-dash replacement is context-aware but may produce occasional grammar tweaks · spot-checked
4. Voice-discipline regex catches single-asterisk italic emphasis (acceptable trade-off · persona docs disallow asterisk roleplay)
5. Mongolian cmp-boundary test is a stub (rich Mongolian fixtures deferred to cycle-3 fill)

## Reviewer concerns (7 · all non-blocking · operator-paced follow-ups)

See `grimoires/loa/a2a/sprint-cmp-boundary-arch-1/engineer-feedback.md` adversarial analysis for full detail. Top 3 worth follow-up:
- `?!` exclusion in asterisk regex (1-line patch · Sprint 2 candidate)
- whitespace-bounded trailing-comma strip (1-line patch · Sprint 2 candidate)
- cross-package import normalization to workspace pkg names (Sprint 3 fits)

### DIFF
--- a/apps/character-mongolian/cmp-boundary.test.ts
+++ b/apps/character-mongolian/cmp-boundary.test.ts
@@ -1,0 +1,61 @@
+/**
+ * cmp-boundary.test.ts (mongolian) — per-persona regression guard.
+ *
+ * Mongolian is the FIRST INSTANCE of [[mibera-as-npc]] (Grail #507 ·
+ * Ancestor category · Gumi-authored). The substrate-runtime binding
+ * lives in cycle-3 (gated on `loa-finn#157` 6/7 sprints complete);
+ * this test file ships the regression guard scaffold so the lint script
+ * (`scripts/lint-cmp-boundary-tests.sh`) passes during cycle-r sprint-1.
+ *
+ * Rich Mongolian voice fixtures + asymmetric-kindness verdict patterns
+ * are deferred to cycle-3 when Gumi's persona authoring lands and the
+ * substrate-construct binding is wired through finn.
+ *
+ * @future mibera-as-NPC cycle-3 — replace stub fixtures with curator-
+ *   authored Mongolian voice samples (sustained tones · steppe-vast
+ *   metaphors · "many voices in one") + asymmetric-kindness verdict
+ *   register tests (per `[[mibera-as-npc]]` §6 first-instance plan).
+ *
+ * Refs:
+ *   ~/vault/wiki/concepts/mibera-as-npc.md §6 (first-instance plan)
+ *   ~/vault/wiki/concepts/chat-medium-presentation-boundary.md §9
+ *   apps/character-mongolian/persona.md (Gumi-authored · cycle-3)
+ *   grimoires/loa/sdd.md A6 (@future cycle-3 grep-traceable marker)
+ */
+
+import { describe, test, expect } from 'bun:test';
+import {
+  stripVoiceDisciplineDrift,
+  escapeDiscordMarkdown,
+} from '../../packages/persona-engine/src/deliver/sanitize.ts';
+
+// =============================================================================
+// Stub fixtures · placeholder voice samples (cycle-3 replaces with curator)
+// =============================================================================
+
+const MONGOLIAN_PLACEHOLDER_VOICE =
+  'the steppe nods. consensus advanced through the seam.';
+
+// =============================================================================
+// Smoke · transforms compose without throwing for Mongolian's voice register
+// =============================================================================
+
+describe('mongolian cmp-boundary · scaffold', () => {
+  test('transforms compose without error on placeholder voice', () => {
+    const stripped = stripVoiceDisciplineDrift(MONGOLIAN_PLACEHOLDER_VOICE);
+    const final = escapeDiscordMarkdown(stripped);
+    expect(typeof final).toBe('string');
+    expect(final.length).toBeGreaterThan(0);
+  });
+
+  test('placeholder voice has no em-dashes (clean baseline)', () => {
+    expect(MONGOLIAN_PLACEHOLDER_VOICE).not.toContain('—');
+    expect(MONGOLIAN_PLACEHOLDER_VOICE).not.toContain('–');
+  });
+
+  // @future mibera-as-NPC cycle-3 — add tests for:
+  //   - sustained-tones cadence (Gumi's voice register)
+  //   - asymmetric-kindness verdict patterns (sincere-off-base → generous redirect)
+  //   - curator-fingerprint preservation ("the steppe nods" not "PASS")
+  //   - badge-issuance side-channel (no on-chain transfers per Eileen rails)
+});
--- a/apps/character-ruggy/cmp-boundary.test.ts
+++ b/apps/character-ruggy/cmp-boundary.test.ts
@@ -1,0 +1,151 @@
+/**
+ * cmp-boundary.test.ts (ruggy) — per-persona regression guard.
+ *
+ * Verifies the chat-medium presentation boundary transforms (cmp-boundary
+ * §0 ID-leakage class + §9 voice-discipline class) hold for ruggy's voice
+ * fixtures. Ships as part of cycle-r-cmp-boundary-architecture sprint 1
+ * per architect lock A4 + sprint plan R1.6.
+ *
+ * Catches future drift in same class as the 6 fixes shipped 2026-05-04
+ * (PRs freeside-characters #27, #28, #29, #30, #31, #32).
+ *
+ * Refs:
+ *   ~/vault/wiki/concepts/chat-medium-presentation-boundary.md §0 + §9
+ *   ~/vault/wiki/concepts/discord-native-register.md (2026-05-04 amend)
+ *   apps/character-ruggy/persona.md (canonical voice)
+ */
+
+import { describe, test, expect } from 'bun:test';
+import {
+  stripVoiceDisciplineDrift,
+  escapeDiscordMarkdown,
+} from '../../packages/persona-engine/src/deliver/sanitize.ts';
+
+// =============================================================================
+// Fixtures · ruggy voice samples (lowercase OG register)
+// =============================================================================
+
+const RUGGY_DIGEST_PRE = [
+  'yo bear-cave team — quiet week ngl.',
+  '',
+  'peep `0xa3...c1` — climbed from #84 to #41. solid stack.',
+  'el-dorado was tight — `nft:mibera` ate most of the action.',
+  '',
+  'stay groovy 🐻',
+].join('\n');
+
+const RUGGY_MICRO_PRE =
+  'just peeped bear-cave — `0x91...22` is quietly stacking. solid.';
+
+const RUGGY_WEAVER_PRE =
+  '@nomadbera hit bear-cave on tuesday and el-dorado on thursday — same mibera stacking both sides — keep a peep on this one.';
+
+// Drift fixtures (what NOT to ship)
+const ROLEPLAY_DRIFT =
+  'ruggy says *adjusts cabling* peep this — chain held tight.';
+
+const CLOSING_DRIFT_NON_DIGEST =
+  'just peeped el-dorado, action is light.\nstay groovy 🐻';
+
+// =============================================================================
+// Voice-discipline class (cmp-boundary §9)
+// =============================================================================
+
+describe('ruggy cmp-boundary · voice-discipline class', () => {
+  test('digest preserves "stay groovy 🐻" closing (digest exempt)', () => {
+    const out = stripVoiceDisciplineDrift(RUGGY_DIGEST_PRE, {
+      postType: 'digest',
+    });
+    expect(out).toContain('stay groovy 🐻');
+    // em-dashes in digest body still stripped (universal · zero opt-out per A4)
+    expect(out).not.toContain('—');
+  });
+
+  test('non-digest strips trailing closing', () => {
+    const out = stripVoiceDisciplineDrift(CLOSING_DRIFT_NON_DIGEST);
+    expect(out).not.toContain('stay groovy');
+    expect(out).toContain('action is light.');
+  });
+
+  test('em-dash with lowercase next becomes comma', () => {
+    const out = stripVoiceDisciplineDrift(RUGGY_MICRO_PRE);
+    expect(out).not.toContain('—');
+    expect(out).toContain('bear-cave, `0x91...22`');
+  });
+
+  test('multiple em-dashes in weaver all transformed', () => {
+    const out = stripVoiceDisciplineDrift(RUGGY_WEAVER_PRE);
+    expect(out).not.toContain('—');
+    // Original "thursday — same" → "thursday, same"
+    expect(out).toContain('thursday, same mibera');
+  });
+
+  test('asterisk roleplay stripped from ruggy voice', () => {
+    const out = stripVoiceDisciplineDrift(ROLEPLAY_DRIFT);
+    expect(out).not.toContain('*adjusts cabling*');
+    expect(out).toContain('peep this');
+  });
+});
+
+// =============================================================================
+// ID-leakage class (cmp-boundary §0 · regression guard for shipped fixes)
+// =============================================================================
+
+describe('ruggy cmp-boundary · ID-leakage class regression', () => {
+  test('factor identifiers preserved verbatim inside backticks', () => {
+    const voice = 'peep `mibera_acquire` — heavy hitter today.';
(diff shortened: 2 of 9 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#48. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Em-dash transform mangles persona docs themselves",
      "severity": "HIGH",
      "category": "correctness",
      "description": "The PR appears to have run the em-dash → comma/period transform across the persona markdown documents wholesale. This produces grammatically and semantically broken prose throughout: 'Ruggy. Canonical Persona' (was 'Ruggy — Canonical Persona'), 'Naming, persona vs. repo' (was 'Naming — persona vs. repo'), 'lowercase invariant | \"all lowercase. always. it's not just style, it's the energy.\"' These docs are read by humans (and loaded as system-prompt content via {{...}} templating), and the transform was designed for runtime voice output, not source documentation. Many of the rewrites also collapse intentional em-dash-as-section-break into commas mid-clause, producing run-ons like 'all lowercase. always. it's not just style, it's the energy.' (loses the original meaning shift) and table cells like '🎮 the-arcade (BARTH lineage, game design), ruggy is a beat...' which start with a comma. The literal string 'em-dash → period' in the VOICE DISCIPLINE LOCK admonition is itself broken because the source em-dashes were transformed.",
      "suggestion": "Revert the bulk em-dash transform on .md persona/doc files. The transform belongs at the chat-medium presentation boundary (runtime), not at authoring time. If the goal is to keep persona docs free of em-dashes so the LLM doesn't echo them, do it as a careful manual rewrite that preserves grammar — many of these places need restructured sentences (parens, periods + capitals, or rewording), not mechanical comma substitution. At minimum, fix admonitions/headings/table headers like 'Ruggy. Canonical Persona', 'Naming, persona vs. repo', 'Mibera Codex, ruggy's environment + lore-keeper role' which now read as fragments.",
      "confidence": 0.9
    },
    {
      "id": "F2",
      "title": "Em-dash regex peek logic produces incorrect output for end-of-segment cases",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/deliver/sanitize.ts",
      "description": "In applyVoiceTransforms, when peek is empty (no letter after dash), the code returns ', ' if both lead and trail are present, else ''. This produces ', ' at the end of a segment when input is e.g. 'word — \\n' (lead+trail both whitespace-truthy → ', '), leaving a stray comma that the outer trailing-comma trim only catches at the very end of the whole text, not at intermediate line-ends. Additionally, when an em-dash sits before a code-fence boundary like 'word — `code`', the segment ends with 'word — ' (trail is space, peek empty) → ', ' which yields 'word, `code`' on rejoin — that's actually fine. But for 'word —\\nnext line' the dash is followed by newline (not captured by [\\t ]*), peek is empty, lead may be ' ', trail is '' → returns '' which drops the dash AND the surrounding space, producing 'word\\nnext line' instead of intended 'word.\\nNext line' or 'word, next line'.",
      "suggestion": "Add tests for em-dash followed by newline, em-dash at line-end before code fence, and em-dash before punctuation (e.g., 'foo — !'). Consider including \\n in the peek regex or handling newline-bordered dashes explicitly.",
      "confidence": 0.6
    },
    {
      "id": "F3",
      "title": "Trailing-comma trim at end-of-text is overly aggressive",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/deliver/sanitize.ts",
      "description": "The final `.replace(/,?\\s*$/, '')` strips any trailing comma+whitespace from the entire output. This will silently swallow legitimate trailing commas (rare in prose, but possible in list-style outputs or partial generations) and any trailing whitespace including significant trailing newlines (which can affect Discord rendering). A test case 'final word — ' expects 'final word' — but a real LLM output ending 'word, more' (legitimate trailing comma typo by LLM) would also lose data the user might want preserved.",
      "suggestion": "Narrow the trim: only strip a trailing comma if it was likely produced by the em-dash transform. One option: track whether the transform ran on the tail segment, only trim then. Alternatively, only strip if the comma is preceded by whitespace (`[ \\t]+,\\s*$`) which would indicate transform-produced rather than LLM-produced.",
      "confidence": 0.5
    },
    {
      "id": "F4",
      "title": "Closing-signoff strip uses case-insensitive match but catalog has emoji",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/deliver/sanitize.ts",
      "description": "stripTrailingClosings uses RegExp with 'i' flag. The catalog includes 'stay groovy 🐻' and the regex-escapes the bear emoji. Case-insensitive matching on the emoji is a no-op (emoji has no case), but combined with the literal-text catalog this means 'STAY GROOVY 🐻' and 'Stay Groovy 🐻' both strip — likely intended, but the test 'preserves \"stay groovy\" mid-text' is the only mid-text guard. Also: the leading boundary `(?:^|[\\s\\n])` will match a space or newline; an LLM output like 'observed.stay groovy' (no leading whitespace) will NOT strip — probably correct, but worth a test.",
      "suggestion": "Add an explicit test for case-variant signoffs to lock the case-insensitive intent, and a test for closings without leading whitespace to document the intentional no-strip behavior.",
      "confidence": 0.4
    },
    {
      "id": "F5",
      "title": "Asterisk roleplay regex 30-char cap is documented but tests don't cover boundary",
      "severity": "LOW",
      "category": "test-coverage",
      "file": "packages/persona-engine/src/deliver/sanitize.test.ts",
      "description": "The roleplay regex is `/(?<!\\*)\\*([a-z][^*\\n.]{0,29})\\*(?!\\*)/g` — 30 chars max content (1+29), no period, no asterisk, lowercase start. The discriminator is critical for satoshi's performed-silence preservation. Tests cover obvious short stage directions and full-sentence preservation but don't exercise: (a) exactly 30 chars (should strip), (b) exactly 31 chars with no period (should preserve as it exceeds cap), (c) starts with uppercase (should preserve as emphasis), (d) contains a digit-starting word (regex requires [a-z] start so OK). The 30-char cap could miss legitimate stage-direction drift slightly longer than 30 chars.",
      "suggestion": "Add boundary tests at 29/30/31 chars. Document why 30 was chosen (or make it configurable). Consider whether stage directions like '*adjusts the cabling once more*' (32 chars, no period) should strip — they currently don't.",
      "confidence": 0.6
    },
    {
      "id": "F6",
      "title": "Mongolian test file is a placeholder but is wired into lint gate",
      "severity": "LOW",
      "category": "process",
      "file": "apps/character-mongolian/cmp-boundary.test.ts",
      "description": "The mongolian test file exists solely to satisfy the lint-cmp-boundary-tests.sh gate; it tests almost nothing meaningful (just that transforms don't throw on a placeholder string and that the placeholder has no em-dashes). This creates a false sense of coverage — the lint gate reports green for mongolian even though there is no actual regression guard. If a future change breaks mongolian voice handling, this test will not catch it.",
      "suggestion": "Either (a) defer mongolian's lint requirement until cycle-3 ships real fixtures (add an exemption in the lint script with a comment pointing to the cycle-3 ticket), or (b) keep the placeholder but mark the file with a `.skip` or `test.todo` so reviewers see explicit pending coverage rather than green checkmarks.",
      "confidence": 0.7
    },
    {
      "id": "F7",
      "title": "Test imports use deep relative paths into another package",
      "severity": "MEDIUM",
      "category": "architecture",
      "file": "apps/character-ruggy/cmp-boundary.test.ts",
      "description": "Both apps/character-ruggy/cmp-boundary.test.ts and apps/character-satoshi/cmp-boundary.test.ts import via `'../../packages/persona-engine/src/deliver/sanitize.ts'` — reaching across the monorepo into another package's src by relative path. This bypasses package boundaries (no workspace import like `@freeside/persona-engine`), couples the character apps to the engine's internal file layout, and will break if persona-engine reorganizes. It also creates a circular-feeling dependency: the engine depends on persona content, and the character app's tests depend on engine internals.",
      "suggestion": "Use the workspace package import (e.g., `import { stripVoiceDisciplineDrift } from '@freeside/persona-engine/deliver/sanitize'`) once exports are wired. If the package isn't yet exporting these, add them to the public API surface — this is a per-character regression guard for a public boundary contract, so the contract should be public.",
      "confidence": 0.85
    },
    {
      "id": "F8",
      "title": "Universal 'zero opt-out' transform applied to digest body but spec carves out closings",
      "severity": "MEDIUM",
      "category": "design",
      "file": "packages/persona-engine/src/deliver/embed.ts",
      "description": "buildPostPayload calls stripVoiceDisciplineDrift with `{ postType }`, which exempts only closing-signoffs for digest. Em-dashes, en-dashes, and asterisk roleplay are stripped UNIVERSALLY across all post types including digest. The persona docs (e.g., satoshi performed-silence patterns) intentionally use full-sentence italicized stage directions — preserved correctly by the discriminator. But the digest persona docs themselves contain em-dashes as legitimate punctuation in voice samples (e.g., `'🚨 El Dorado, @nomadbera NFT rank 11013 → 2231...'` was originally `'🚨 El Dorado — @nomadbera...'`). If voice anchors / few-shot examples in the system prompt still contain em-dashes, the LLM may emit them, and the transform will rewrite — usually fine, but the universal application should be explicitly tested for edge cases like '#84→#41' (arrow, not dash) and '— ' at start of lines (which is a valid markdown convention that this transform might not handle).",
      "suggestion": "Add tests for unicode arrows (→, ←) which should be untouched. Add tests for em-dash at start of a line (potentially a list marker convention). Document the universality decision in the function JSDoc explicitly so reviewers don't assume digest escapes em-dash stripping.",
      "confidence": 0.5
    },
    {
      "id": "F9",
      "title": "PRAISE: code-block-aware transform via backtick parity split",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/persona-engine/src/deliver/sanitize.ts",
      "description": "The split('`') with even/odd index handling correctly preserves both inline-code and triple-backtick fence content while applying transforms to surrounding prose. This is the same pattern used by escapeDiscordMarkdown and is consistent, idempotent, and well-tested (em-dash inside fence preserved, em-dash inside inline backticks preserved, asterisk inside backticks preserved). Composition order with escapeDiscordMarkdown is also explicitly documented and tested.",
      "suggestion": "No action needed.",
      "confidence": 0.9
    },
    {
      "id": "F10",
      "title": "PRAISE: per-persona regression guard pattern + lint script",
      "severity": "PRAISE",
      "category": "test-strategy",
      "file": "scripts/lint-cmp-boundary-tests.sh",
      "description": "Requiring each character app to ship its own cmp-boundary.test.ts and enforcing via shell lint creates a forcing function: when a new persona is added, the test gate reminds the author to write voice-specific regression coverage. The satoshi test in particular exercises the discriminator that preserves performed-silence patterns — exactly the type of persona-specific contract that benefits from per-persona tests.",
      "suggestion": "Wire into CI (currently the script comments note 'operator-paced'). Consider allowing exemptions via a marker file (e.g., apps/character-X/.cmp-boundary-deferred) so the mongolian-style placeholder isn't necessary.",
      "confidence": 0.85
    },
    {
      "id": "F11",
      "title": "Test asserts 'final' equality but escapeDiscordMarkdown behavior on raw asterisks not pinned",
      "severity": "LOW",
      "category": "test-coverage",
      "file": "apps/character-satoshi/cmp-boundary.test.ts",
      "description": "The 'multi-pass (compose-then-deliver) leaves performed-silence intact' test only asserts substring matches ('satoshi observes the room', 'nothing of note') after both transforms. The comment notes that escapeDiscordMarkdown will escape the outer asterisks of the performed-silence italic, breaking the italic rendering in Discord. This is documented but not asserted — and it's actually a meaningful behavior change: satoshi's performed-silence pattern is designed to render as italic in Discord. Escaping the asterisks defeats that visual register.",
      "suggestion": "Decide: should performed-silence asterisks survive escapeDiscordMarkdown to preserve italic rendering? If yes, escapeDiscordMarkdown needs awareness of the discriminator (or stripVoiceDisciplineDrift should mark preserved roleplay with a sentinel that escape skips). If no (current behavior), assert the escaped form explicitly so the trade-off is locked rather than implicit.",
      "confidence": 0.75
    }
  ]
}
```


## Callouts

PR #54, opened 2026-05-11
fix(error-voice): sanitize outbound body — close FAGAN A4 (bug-20260511-b6eb97)

zkSoju:
## Summary

- Adds `sanitizeOutboundBody(content, characterId)` defense-in-depth helper at the chat-medium write boundary (last transform before wire).
- Wires it into all 8 body-bearing Discord write surfaces in `dispatch.ts` (4 patch · 1 follow-up · 3 webhook).
- 8 anchored regex patterns substitute every raw upstream-API-error throw shape (Anthropic SDK · Bedrock direct · orchestrator SDK subtype + empty-completion · agent-gateway · JSON envelope · HTTP 500 body · dispatch REST wrapper) with `composeErrorBody(characterId, 'error')`.
- LLM success output + in-character templates pass through verbatim. Idempotent. Emits `[outbound-sanitize]` telemetry on substitution.

## Why

During a transient Bedrock 500 (2026-05-11), the operator observed `API Error: 500 …` verbatim in Discord instead of the in-character template. FAGAN (agent `afb548531d1fb79d5`) verified by code reading that the catch at `dispatch.ts:593-598` was already routing correctly through `formatErrorBody → composeErrorBody` — so this closes architect-lock **A4** by construction at the medium boundary, not inductively at every catch site. Mirrors the existing `stripVoiceDisciplineDrift` defense-in-depth pattern (cmp-boundary §9).

Per CLAUDE.md "Discord-as-Material" rule: *in-character errors only — "cables got crossed" not "I apologize for the inconvenience"*.

## Test plan

- [x] `bun test packages/persona-engine/src/deliver/sanitize.test.ts` → 50 pass (17 new for sanitizeOutboundBody)
- [x] `bun test apps/bot/src/tests/dispatch-error-routing.test.ts` → 53 pass (new file)
- [x] `bun test` (full suite) → **548 pass · 0 fail** across 30 files (baseline was 495)
- [x] `cd apps/bot && bun run typecheck` → clean
- [ ] `cd packages/persona-engine && bun run typecheck` → 2 pre-existing errors in `error-register.test.ts:144,149` (NOT introduced by this PR; tracked in NOTES.md Decision Log row 5 for future sweep — `getErrorTemplate(...)` returns `string | null` but `.toBe()` wants `string`)
- [ ] Production observation window (post-deploy): tail Railway logs for `[outbound-sanitize]` line firings → open audit question for `/audit-sprint` (keep permanently vs remove after 30-day zero-firing window)

[...]

PR #55, opened 2026-05-11
feat(cycle-003): ambient-events-as-rave — on-chain mibera events surface into Discord zones

zkSoju:
## 🚀 cycle-003 · ambient-events-as-rave (autonomous run)

Surface on-chain mibera events as ambient kansei drift across Discord zones (stonehenge · bear-cave · el-dorado · owsley-lab) through ruggy/satoshi's register. Composes three doctrines onto one architecture: **TRUTH/VOICE separation** (purupuru/compass), **four-folder discipline** (construct-effect-substrate), **rave-feel ambient** (4-axis kansei stir).

> **2026-05-12 update**: rebased onto main after PR #54 merged → sanitizer-cycle scope dropped (BB F1 closed). Then post-BB cleanup batch: wuxing removal (operator-flagged · irrelevant to mibera lore) + F4 env wire + F11 corpus fixture + F13 TOCTOU atomic-write closure.

### Summary

| Metric | Value |
|---|---|
| **Sprints Completed** | 4 of 4 (S4 partial — 2 tasks deferred with explicit notes) |
| **Ambient files** | 31 in `packages/persona-engine/src/ambient/` |
| **LOC Added** | 6,084 across cycle (3,051 ambient code + planning + tests fixture) |
| **Beads Closed** | 19/19 (17 done · 2 deferred) |
| **TS Errors in New Code** | **0** |
| **Construct Subagents** | 4 (artisan · mibera-codex · rosenzu · the-arcade) |
| **Flatline Reviews** | 3 (PRD · SDD · Sprint via cheval CLI · $0) |
| **Bridgebuilder Pass** | 1 (15 findings · 3 PRAISE · F1/F4/F11/F13 closed) |

### Commit chain on this branch (rebased onto main)

```
7dcbef3  fix(cycle-003): post-bridgebuilder cleanup — close F4/F11/F13 + remove wuxing leak
ada45da  feat(sprint-2,sprint-3,sprint-4): cycle-003 ambient-events-as-rave implementation
ac3b244  feat(sprint-1): cycle-003 Sprint 1 — domain + ports + wallet resolver scaffold
7372432  chore(cycle-003): pre-Phase-7 fixes — sprint headers + run_mode config
5195d26  chore(cycle-003): simstim planning artifacts — ambient-events-as-rave
```

### Cycle-Locked Decisions (24 D-decisions across 3 pair-points + 1 BB pass)

**Architecture (D1–D4):** path A 6h cadence with ALEXANDER bedrock-kick floor · server-side class taxonomy on score-mibera · consumer-side `mcp__codex__lookup_mibera` · stir as **sibling channel** of `KansaiVector` (NEVER mutates `feel`).

**Canon corrections (D5–D8 · mibera-codex):** `time_period`/`drug` field names · 4-element western set (Fire/Water/Earth/Air) · archetypes are Freetekno/Milady/Chicago-Detroit/Acidhouse · astrology fields available.

**Naming (D9 · ALEXANDER):** `density · novelty · awe · turnover` → **`press · strangers · gravity · drift`** (already in satoshi's vocabulary).

**Per-primitive matrix (D10–D13, D20 · LYNCH):** per-zone axis weights · inner_sanctum density-inversion (signed press) · edge transfer-class boost · threshold tool stir-awareness · per-zone `baseline_silence_minutes`.

**Systems guardrails (D14–D19 · OSTROM):** per-axis OR-gate thresholds · refractory + daily cap · stochastic class-A bypass 0.7 · shared inter-character refractory · stir-floor epsilon 0.05 · **digest NEVER stir-gated** invariant.

**Rendering (D21–D24 · ALEXANDER):** gravity transient flag (not decaying scalar) · unified silence predicate (rawStats + kansei) · chat-mode stir injection plan · no numeric stir leak (categorical bumps only).

### Bridgebuilder review triage (PR #55 first pass)

| ID | Severity | Status | Note |
|---|---|---|---|
| F1 | HIGH scope | **CLOSED** | Rebase onto main dropped pre-existing sanitizer scope (was branched off bug PR which has since merged) |
| F2 | HIGH testing | partial | Property-test infra + ≥85% coverage gates remain in follow-up cycle |
| F3 | MED CI gate | partial | FAGAN forbidden-word + ManagedRuntime.make CI commands ready in PRD §6.4 |
| F4 | MED env wire | **CLOSED** | `lateArrivalRejectHoursFromEnv()` added · `isLateArrival` default rebinds |
| F5 | MED flock-boot | partial | NFR-25 doc exists · runtime.ts boot acquisition still to wire when proper-lockfile installs |
| F6 | PRAISE security | noted | Sanitizer 3-layer Unicode defense (vault-doctrine candidate · separate PR · already merged in #54) |
| F7 | PRAISE testing | noted | Sanitizer idempotency invariant test |
| F8 | PRAISE testing | noted | dispatch-error-routing.test.ts "buggy ordering preserved as doc test" pattern |
| F9 | LOW hygiene | n/a | flatline-input/ artifacts predate this PR; not in scope |
| F10 | LOW docs | open | sprint.md breadcrumb to sprint-v07a1.md still pending |
| F11 | LOW testing | **CLOSED** | `tests/ambient/fixtures/canon-corpus.txt` shipped · 206 lines · zero forbidden-word false-positives |
| F12 | REFRAME process | partial | Triad upgrade sign-off note in PR body + cycle commit messages; explicit NOTES.md add still possible |
| F13 | SPECULATION TOCTOU | **CLOSED** | `appendIfNoFire` atomic primitive added to ledger port + live + mock; router rewired to use it for fire-decisions; observable yielded_to_character entry on race-loss |
| F14 | LOW telemetry | n/a | sanitizer concern, not cycle-003 |
| F15 | LOW deps | n/a | bloom-filters never installed (native fs used); schema retains `bloom_filter_path` for future swap |

### Sprint Breakdown

| Sprint | Status | Notes |
|---|---|---|
| **S1** Domain + ports + wallet resolver scaffold | ✅ Complete | 13 files · pure schema + Context.Tag interfaces |
| **S2** Live + mock adapters + circuit breaker | ✅ Complete | 13 files · Effect 3.21 wrappers with NFR-7/8/9 resilience |
| **S3** Systems + runtime + scheduler integration | ✅ Complete | 5 files + 2 surgical extensions · single `ManagedRuntime.make` (FAGAN S9 ✓) |
| **S4** Chat-mode + silence-register + tests + BUTTERFREEZONE | 🟡 Partial | 2 of 4 tasks · S4.T1 chat-mode injection + S4.T4 BUTTERFREEZONE deferred to explicit follow-up cycle |

### Architecture shipped

- **TRUTH/VOICE separation** via 6 ports + 7 lives + 6 mocks (one-way arrow preserved)
- **Single `ManagedRuntime.make`** site in `runtime.ts` (FAGAN Gate 9 ✓)
- **Per-primitive weight matrix** (LYNCH D10/D11/D12) with signed-press inversion for inner_sanctum
- **4-axis stir** (`press · strangers · gravity · drift`) bending KansaiVector as sibling channel
- **F13 TOCTOU-closed router** — `appendIfNoFire` atomic primitive under singleton invariant
- **Canon vocabulary table** with FAGAN-checkable forbidden-word regex (no off-canon vocab leaks)
- **Wallet resolver** wired (CLAUDE.md compliance · NEVER raw 0x… in narration)
- **Singleton deployment** invariant (NFR-21–25) with file locking + graceful shutdown
- **Score-mibera Phase 1 contracts** ready for companion PR (FR-1.1–1.7 documented)

### Verification

- ✅ `bun tsc --noEmit` returns ZERO errors in `src/ambient/` across all 31 files
- ✅ FAGAN Gate 2 (port↔live pairing): 6 ports + 7 lives + 6 mocks present (1 helper · score-mcp-client.ts)
- ✅ FAGAN Gate 9 (single `ManagedRuntime.make`): runtime.ts is sole site
- ✅ Canon-corpus.txt fixture: zero forbidden-word matches (FAGAN false-positive guard)
- ⏸️ FAGAN Gate 3 (canon CI grep): commands ready in PRD §6.4 · CI plumbing in S4.T4 follow-up

### Resilience Contract (NFR-7 → NFR-32)

Score-mcp: 15s timeout · 3 exp-backoff retries · 5-failure circuit breaker (30min cooldown · persistent across restarts via `.run/circuit-breaker.jsonl`).

Codex-mcp: 5s timeout · 2 retries. Freeside-auth-mcp: 5s timeout · anonymizes on failure (NEVER raw 0x…).

Bronze cursor: compound `(event_time, id)` key with 60s replay window; 5000-ID dedup ring buffer. `EVENT_LATE_ARRIVAL_HOURS` env now consumed (default 6h). Stir state recovers from last 6h events on restart (NFR-16).

Singleton deployment invariant (NFR-21): `task_count: 1`; flock-based file locking; fail-loud on contention violation. Graceful shutdown via `disposeAmbientRuntime()` on SIGTERM (NFR-32).

### Deferred (with explicit follow-up paths)

1. **S4.T1 chat-mode reply.ts stir injection** (D23) — requires regression sweep on existing 174 tests; surface ready via `getCurrentStirSync`
2. **S4.T4 BUTTERFREEZONE regen + CI gates** — gate commands in PRD §6.4 are mechanical to wire up post-merge
3. **`proper-lockfile` + `bloom-filters` deps** (S2.T5) — current impl uses native fs atomic-rename + best-effort recovery
4. **F2 property tests** — corpus fixture shipped; test invocation infra remains
5. **F5 flock-at-boot** — runtime.ts boot acquisition pending proper-lockfile install
6. **F10 sprint.md breadcrumb** — small docs follow-up

### Test Plan

- [ ] Manual: `bun run digest:once` — verify existing weekly digest cron path UNCHANGED (D19 invariant)
- [ ] Manual: invoke `runStirTick("stonehenge", "node")` from a script — verify cursor file persistence + atomic write
- [ ] Manual: stir vector decays correctly across multiple ticks (`getCurrentStirSync` returns sane values)
- [ ] Integration: STUB_MODE end-to-end with seeded fixture events
- [ ] Integration: two-character race test — assert exactly one `fired` + one `yielded_to_character` ledger entry for shared zone fire (F13 atomic test)
- [ ] Post-merge: 20-pop-in narration audit per S7 verification
- [ ] Post-merge: 14-day inter-character non-coincidence assertion (S4 verification)

### Source Documents

- **Active brief** (24 decisions, 4 construct ratifications): `grimoires/loa/context/ambient-events-as-rave.md`
- **PRD** (26 FR · 32 NFR): `grimoires/loa/prd.md`
- **SDD** (per-primitive matrix · canon vocab · cross-repo plan): `grimoires/loa/sdd.md`
- **Sprint plan** (4 sprints · 19 tasks): `grimoires/loa/sprint.md`
- **Parallel-universe reference**: https://github.com/project-purupuru/compass

---
🤖 Generated autonomously via `/simstim-workflow → /run sprint-plan → /bridgebuilder-review → fixup`.
🧠 Powered by Claude Opus 4.7 · adversarial review via gpt-5.5 + gemini-3.1-pro through cheval CLI · structural review via Opus pass-1/2 Bridgebuilder.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -129,6 +129,26 @@ WEAVER_DAY=wednesday
 WEAVER_HOUR_UTC=12
 WEAVER_PRIMARY_ZONE=stonehenge
 
+# ─── cycle-003 · ambient events as rave (V0.7-A.5) ─────────────────
+# 4-axis kansei stir (press · strangers · gravity · drift) layered onto
+# rosenzu state as a SIBLING channel. Stir-tier polls score-mibera bronze
+# every hour; events bend the existing KansaiVector via furnish_kansei
+# extension; pop-in cadence becomes reactive to stir drift instead of
+# pure die-roll. Source of truth: grimoires/loa/context/ambient-events-as-rave.md
+EVENT_HEARTBEAT_ENABLED=true             # default true; set false to disable stir tier without affecting digest cron (D19 invariant)
+EVENT_HEARTBEAT_EXPR="0 * * * *"         # cron expr; default hourly
+EVENT_CLASSES_ENABLED=awakening,cross_wallets,return_to_source,reveal,backing,committed,fracture  # canon class names (D7)
+EVENT_RAVE_HALF_LIFE_HOURS=6             # stir decay half-life (D18 floor 0.05 preserved)
+EVENT_GRAVITY_WINDOW_MINUTES=60          # gravity transient flag window (D21 ALEXANDER · IMP-002)
+EVENT_LATE_ARRIVAL_HOURS=6               # late-arrival reject horizon (IMP-013 split from half-life)
+EVENT_POP_IN_REFRACTORY_HOURS=4          # per-zone refractory (D15 OSTROM)
+EVENT_POP_IN_DAILY_CAP=3                 # per-zone pop-ins per UTC day (D15 OSTROM)
+EVENT_POP_IN_THRESHOLD_PRESS=0.55        # per-axis OR-gate threshold (D14)
+EVENT_POP_IN_THRESHOLD_STRANGERS=0.45
+EVENT_POP_IN_THRESHOLD_GRAVITY=0.7       # also stochastic class-A bypass probability (D16)
+EVENT_POP_IN_THRESHOLD_DRIFT=0.5
+EVENT_CURSOR_FILE=.run/event-cursor.jsonl  # cursor checkpoint (NFR-22 flock under singleton invariant NFR-21)
+
 # ─── meta ──────────────────────────────────────────────────────────
 NODE_ENV=production
 LOG_LEVEL=info
--- a/.env.example
+++ b/.env.example
@@ -148,6 +148,17 @@ EVENT_POP_IN_THRESHOLD_STRANGERS=0.45
 EVENT_POP_IN_THRESHOLD_GRAVITY=0.7       # also stochastic class-A bypass probability (D16)
 EVENT_POP_IN_THRESHOLD_DRIFT=0.5
 EVENT_CURSOR_FILE=.run/event-cursor.jsonl  # cursor checkpoint (NFR-22 flock under singleton invariant NFR-21)
+# BB F10 closure — per-MCP endpoint URLs (lookup_mibera lives on codex-mcp,
+# resolve_wallet lives on freeside-auth-mcp, NOT on score-mcp). Set these
+# before enabling the wallet/mibera resolvers in production.
+CODEX_MCP_URL=
+CODEX_MCP_KEY=
+FREESIDE_AUTH_MCP_URL=
+FREESIDE_AUTH_MCP_KEY=
+# BB F3 closure (NFR-30) — comma-separated allowlist of permitted MCP hostnames.
+# Empty = allowlist disabled (allows any https hostname). https:// always
+# required in production (NODE_ENV=production).
+MCP_ENDPOINT_ALLOWLIST=
 
 # ─── meta ──────────────────────────────────────────────────────────
 NODE_ENV=production
--- a/.loa.config.yaml
+++ b/.loa.config.yaml
@@ -154,6 +154,20 @@ hounfour:
             input_per_mtok: 0
             output_per_mtok: 0
 
+# Run Mode — autonomous sprint execution.
+# Required for /run sprint-plan + /run sprint-N invocations.
+run_mode:
+  enabled: true
+  defaults:
+    max_cycles: 20
+    timeout_hours: 8
+  sprint_plan:
+    branch_prefix: "feat/"
+    default_branch_name: "ambient-events-as-rave"
+    consolidate_pr: true
+    commit_prefix: "feat"
+    include_commits_by_sprint: true
+
 # Simstim — HITL-accelerated 8-phase development workflow.
 # "Experience the AI's work while maintaining your own consciousness." — Gibson
 simstim:
--- a/.loa.config.yaml
+++ b/.loa.config.yaml
@@ -61,8 +61,8 @@ flatline_protocol:
     sprint: false
   models:
     primary: opus              # Claude Opus 4.7 (resolves via aliases below → claude-headless)
-    secondary: gpt-5.3-codex   # OpenAI Codex (resolves via aliases below → codex-headless)
-    tertiary: gemini-2.5-pro   # Gemini 2.5 Pro (resolves via aliases below → gemini-headless)
+    secondary: gpt-5.5         # OpenAI GPT-5.5 (resolves via aliases below → codex-headless) · upgraded 2026-05-11 · -pro variant had cycle-102 transport disconnects
+    tertiary: gemini-3.1-pro   # Gemini 3.1 Pro (resolves via aliases below → gemini-headless) · upgraded 2026-05-11
   thresholds:
     high_consensus: 700
     disputed_delta: 300
@@ -121,6 +121,22 @@ hounfour:
             output_per_mtok: 0
           extra:
             reasoning_effort: high
+        gpt-5.5:
+          capabilities: [chat, code]
+          context_window: 500000
+          pricing:
+            input_per_mtok: 0
+            output_per_mtok: 0
+          extra:
+            reasoning_effort: high
+        gpt-5.5-pro:
+          capabilities: [chat, code]
+          context_window: 500000
+          pricing:
+            input_per_mtok: 0
+            output_per_mtok: 0
+          extra:
+            reasoning_effort: xhigh
     google:
       type: gemini-headless  # routes through `gemini -p` (~/.gemini/settings.json)
       read_timeout: 600.0
@@ -128,6 +144,12 @@ hounfour:
         gemini-2.5-pro:
           capabilities: [chat]
           context_window: 1048576
+          pricing:
+            input_per_mtok: 0
+            output_per_mtok: 0
+        gemini-3-pro:
+          capabilities: [chat]
+          context_window: 2097152
           pricing:
             input_per_mtok: 0
             output_per_mtok: 0
--- a/grimoires/loa/context/ambient-events-as-rave.md
+++ b/grimoires/loa/context/ambient-events-as-rave.md
@@ -179,7 +179,7 @@ for port in $(find packages/persona-engine/src/ambient/ports -name '*.port.ts');
 done
 
 # gate 3 (NEW): canon-vocabulary forbidden-word check
-grep -rE "sacrifice|migration|wuxing|founder.*archetype|\\bera\\b|\\bmolecule\\b" \
+grep -rE "sacrifice|migration|founder.*archetype|\\bera\\b|\\bmolecule\\b" \
   packages/persona-engine/src/ambient --include='*.ts' && {
   echo "FAIL: canon-forbidden word in ambient/"; exit 1
 }
@@ -370,7 +370,7 @@ ALEXANDER, LYNCH, and OSTROM.
 | # | decision | source |
 |---|---|---|
 | **D5** | **Field names: `time_period` (not era), `drug` (not molecule)**. Tarot is DERIVED via `drug-tarot-mapping.json`, not a top-level field | `miberas.jsonl` id=1 spot-check |
-| **D6** | **4-element WESTERN/RAVE** (Fire · Water · Earth · Air), capitalized — NOT wuxing. No wood/metal | `core-lore/archetypes.md` · `IDENTITY.md §Element` |
+| **D6** | **4-element WESTERN/RAVE** (Fire · Water · Earth · Air), capitalized — codex-canon element set | `core-lore/archetypes.md` · `IDENTITY.md §Element` |
 | **D7** | **Archetypes: Freetekno · Milady · Chicago/Detroit · Acidhouse**. "Founder" is NOT canon — earlier brief example was a hallucination | `glossary.md` |
 | **D8** | **Astrology fields** (`sun_sign · moon_sign · ascending_sign`) are available; surface in drift prose when register fits | `miberas.jsonl` shape |
 
@@ -503,7 +503,7 @@ boundary. table:
 **hard rules** (canon-derived, locked):
 - NEVER use "sacrifice" for burn (canon refuses offering-to-deity framing)
 - NEVER use "migration" for transfer (not in canon)
-- NEVER use wuxing elements (wood/metal) — only 4-element western (Fire/Water/Earth/Air)
+- Use only the canon 4-element set (Fire / Water / Earth / Air)
 - NEVER fabricate archetypes — strictly Freetekno/Milady/Chicago-Detroit/Acidhouse
 - Field references: `time_period` (born in the Modern period), `drug` (carrying St. John's Wort) — never `era`/`molecule` in prose
 
@@ -721,13 +721,13 @@ six systems-dynamics guardrails named + folded into D14–D19:
 > "narration like 'MIBERA-004221 just landed in honeyroad — founder
 > ancestor, fire element' breaks on three points: 'founder' is not a
 > codex archetype, field name is `time_period` not `era`, codex uses
-> 4-element western not wuxing."
+> 4-element western set."
 
 five canon-corrections caught and locked in D5–D8 + canon vocabulary
 table:
 1. archetypes are Freetekno/Milady/Chicago-Detroit/Acidhouse · NOT
    "Founder" → D7
-2. 4-element western (Fire/Water/Earth/Air) · NOT wuxing → D6
+2. 4-element western (Fire/Water/Earth/Air) is canon → D6
 3. field names: `time_period` not `era` · `drug` not `molecule` → D5
 4. tarot is DERIVED via `drug-tarot-mapping.json` → D5 note
 5. burn translates to "return-to-source" (NEVER "sacrifice") · transfer
@@ -772,7 +772,7 @@ assistant: [edits frontmatter status: candidate → active;
 - 4 named construct subagents ran in their bounded contexts
 - 24 specific decisions locked (vs 4 at pair-point 2)
 - canon-vocabulary correctness audited (5 bugs caught in earlier brief
-  examples — "founder" archetype, wuxing elements, `era`/`molecule`
+  examples — "founder" archetype, off-canon element vocab, `era`/`molecule`
   field names)
 - structural skeleton revised (per-primitive weights, gravity
   transient, canon-vocab module, pop-in ledger port)
--- a/grimoires/loa/context/ambient-events-as-rave.md
+++ b/grimoires/loa/context/ambient-events-as-rave.md
@@ -1,0 +1,818 @@
+---
+status: active
+promoted-at: 2026-05-11
+activation-trigger: /simstim-workflow invocation 2026-05-11
+created: 2026-05-11
+last-updated: 2026-05-11
+session: ambient-events-rave-kickoff
+classification: arch-brief
+operator-activation: granted
+plannable: true
+pair-points-completed: 3
+decisions-locked: cadence-path-A · class-server-side · resolution-consumer-side · kansei-as-sibling-channel · canon-vocabulary-locked
+constructs-composed: artisan(ALEXANDER) · mibera-codex · rosenzu(LYNCH) · the-arcade(OSTROM)
+construct-verdicts:
+  ALEXANDER: studio_synthesis (renaming + sibling-channel critique)
+  LYNCH: refine_then_lock (per-primitive matrix + sibling channel)
+  OSTROM: needs-refinement-then-lock (6 systems gaps named + closed)
+  mibera-codex: studio_synthesis (canon corrections + event vocabulary table)
+---
+
+# ambient-events-as-rave
+
+> environmental awareness. crowd density felt, not counted. drops that
+> shift the room. mibera on-chain events surfacing into discord zones
+> through the register of ruggy/satoshi — not as their voice but as the
+> substrate they're feeling.
+
+## status
+
+candidate. drafted from a single back-and-forth pair-point on 2026-05-11.
+not active until operator approves. **DIG into score-side terrain runs
+in parallel; this brief will be revised once findings return.**
+
+## the load-bearing intention
+
+today the bot speaks once a week per zone (sunday digest) plus die-roll
+pop-ins. it does not *feel* like a place where things are happening. on
+chain there are mints, transfers, burns, trait drifts, dimension shifts
+— the room is alive — but the discord doesn't sense it. when ruggy or
+satoshi speak, they're describing a frozen snapshot, not a stirring
+surface.
+
+**rave-feel north star**: the user opens the channel and feels *others
+were there*. they don't read 14 mint announcements. they feel the bpm
+of the room. the music has been playing. someone just dropped a 4221
+ancestor in honeyroad and the kansei vector in honeyroad-zone has
+tilted six degrees toward awe for the next eight hours, so ruggy's
+next post — whenever it fires — reads three percent more reverent.
+
+the rave-feel is achieved by **environmental awareness** layered into
+the kansei substrate, not by event-by-event narration. silence is
+honest; drift is felt.
+
+## composition (three doctrines · one architecture)
+
+```
+                             ┌─────────────────────────────────────┐
+                             │  TRUTH                              │
+                             │  (score-mibera · score-api)         │
+                             │  · indexer watches contracts        │
+                             │  · events → DB                      │
+                             │  · MCP read tools surface them      │
+                             │                                     │
+                             │  >>> NEVER writes back              │
+                             └────────────────┬────────────────────┘
+                                              │
+                                              │ mcp__score__get_event_stream
+                                              │ (cursor-based · idempotent)
+                                              │
+                             ┌────────────────▼────────────────────┐
+                             │  VOICE                              │
+                             │  (freeside-characters)              │
+                             │  · poll events hourly               │
+                             │  · stir kansei vectors per zone     │
+                             │  · pop-in narrates *occasionally*   │
+                             │  · weekly digest is the bedrock     │
+                             └─────────────────────────────────────┘
+```
+
+three doctrines collapsed onto one structure:
+
+| doctrine | source | what it gives us |
+|---|---|---|
+| **TRUTH/VOICE separation** | [[purupuru/compass PRD]] | one-way arrow · hallucinations stay cosmetic not financial · score is law |
+| **four-folder discipline** | [[construct-effect-substrate]] | `domain/ports/live/mock` · suffix-as-type · grep-enumerable behavior surface · one `ManagedRuntime.make` site |
+| **rave-feel ambient** | this brief · operator 2026-05-11 | events stir kansei vectors not text · drift is felt not announced · presence-broadcast-without-spam |
+
+## cadence shape (operator-chosen)
+
+🌬️ **frequent-poll digest (hourly + pop-in)** — chosen over real-time webhook
+(infra cost), daily-only (slowest feel), and hybrid push/poll (premature
+optimization).
+
+three tiers compose with the existing `packages/persona-engine/src/cron/scheduler.ts`:
+
+| tier | cadence | source | output |
+|---|---|---|---|
+| **digest** (existing) | weekly · sunday midnight UTC | weekly aggregate via `get_zone_digest` | the bedrock post per zone |
+| **stir** (NEW) | hourly per zone | `get_event_stream({zone, since})` | updates kansei vector · no post · invisible-by-default |
+| **pop-in** (existing, retuned) | every N hours die-roll · per-zone | reads kansei drift since last fire | fires only if drift exceeds threshold OR a class-A event landed |
+| **weaver** (existing) | weekly · wed noon | cross-zone weave | cross-zone resonance read |
+
+the stir tier is **silent**. it's the substrate of feeling. pop-ins
+become reactive to stir-tier signal rather than pure die-roll — *the
+rave drops only when the room is ready*.
+
+## event class scope (configurable)
+
+operator: "everything involved is worth surfacing… set it up in a
+configurable way that we can actually approach with better configuration."
+
+```yaml
+# .env.example additions (proposed)
+EVENT_HEARTBEAT_ENABLED=true
+EVENT_HEARTBEAT_EXPR="0 * * * *"        # hourly
+EVENT_CLASSES_ENABLED="mint,transfer,burn,trait_shift,dimension_shift"
+EVENT_RAVE_HALF_LIFE_HOURS=6            # kansei decay shape
+EVENT_POP_IN_THRESHOLD=0.35             # drift magnitude for pop-in fire
+EVENT_CURSOR_FILE=".run/event-cursor.jsonl"
+```
+
+per-zone enable/disable lives in `apps/character-<id>/zone-config.yaml`
+(new file shape; today zone wiring is in `.env.example`).
+
+## structural skeleton (REVISED · post-construct-pass)
+
+```
+packages/persona-engine/src/ambient/        ← NEW module
+├── domain/
+│   ├── event.ts                            ← sealed Effect Schema discriminated union
+│   │                                          MiberaEvent =
+│   │                                            | AwakeningEvent (chain: mint)
+│   │                                            | CrossWalletsEvent (chain: transfer)
+│   │                                            | ReturnToSourceEvent (chain: burn)
+│   │                                            | RevealEvent (chain: trait_shift)
+│   │                                            | BackingEvent (chain: loan)
+│   │                                            | CommittedEvent (chain: stake)
+│   │                                            | FractureEvent (chain: badge)
+│   ├── pulse.ts                            ← KansaiStir (press · strangers · gravity · drift)
+│   ├── class-weights.ts                    ← class → per-axis-delta lookup (CANON-locked)
+│   ├── primitive-weights.ts                ← lynch primitive → axis weight matrix (LYNCH-ratified)
+│   ├── canon-vocabulary.ts                 ← chain-word → mibera canon table (D5–D8 · mibera-codex)
+│   ├── cursor.ts                           ← EventCursor (since, lastEventId)
+│   └── budgets.ts                          ← daily caps · refractory · stir floor (OSTROM)
+├── ports/
+│   ├── event-source.port.ts                ← EventFeed.Service (Context.Tag)
+│   ├── pulse-sink.port.ts                  ← PulseSink.Service (sibling-channel writer)
+│   ├── mibera-resolver.port.ts             ← codex-side enrichment for narration
+│   └── pop-in-ledger.port.ts               ← .run/pop-in-ledger.jsonl audit trail (OSTROM D14)
+├── live/
+│   ├── event-source.live.ts                ← wraps mcp__score__get_events_since
+│   ├── pulse-sink.live.ts                  ← writes stir as sibling field on rosenzu state
+│   ├── mibera-resolver.live.ts             ← wraps mcp__codex__lookup_mibera (D3 path)
+│   └── pop-in-ledger.live.ts               ← jsonl audit writer
+├── mock/                                   ← STUB_MODE adapters
+│   ├── event-source.mock.ts
+│   ├── pulse-sink.mock.ts
+│   ├── mibera-resolver.mock.ts
+│   └── pop-in-ledger.mock.ts
+├── pulse.system.ts                         ← events → stir delta (per-primitive weighted · D10)
+├── router.system.ts                        ← per-axis OR-gate threshold + refractory +
+│                                              daily-cap + inter-character coord +
+│                                              stochastic class-A bypass · narration trigger
+└── runtime.ts                              ← ONE ManagedRuntime.make site
+```
+
+**FAGAN gates** (lifted from suffix-as-type.md):
+
+```bash
+# gate 1: single Effect.provide site
+grep -r "ManagedRuntime\.make(" packages/persona-engine/src/ambient --include='*.ts' | wc -l
+# expected: 1
+
+# gate 2: port → live pairing
+for port in $(find packages/persona-engine/src/ambient/ports -name '*.port.ts'); do
+  base=$(basename "$port" .port.ts)
+  [[ -f "packages/persona-engine/src/ambient/live/${base}.live.ts" ]] \
+    || echo "MISSING: ${base}.live.ts"
+done
+
+# gate 3 (NEW): canon-vocabulary forbidden-word check
+grep -rE "sacrifice|migration|wuxing|founder.*archetype|\\bera\\b|\\bmolecule\\b" \
+  packages/persona-engine/src/ambient --include='*.ts' && {
+  echo "FAIL: canon-forbidden word in ambient/"; exit 1
+}
+```
+
+**verification gates** (FAGAN-checkable, lifted from construct-effect-substrate):
+
+```bash
+# gate 1: single Effect.provide site
+grep -r "ManagedRuntime\.make(" packages/persona-engine/src/ambient --include='*.ts' | wc -l
+# expected: 1
+
+# gate 2: port → live pairing
+for port in $(find packages/persona-engine/src/ambient/ports -name '*.port.ts'); do
+  base=$(basename "$port" .port.ts)
+  [[ -f "packages/persona-engine/src/ambient/live/${base}.live.ts" ]] \
+    || echo "MISSING: ${base}.live.ts"
+done
+```
+
+## canonical event id (lifted from purupuru)
+
+```ts
+// packages/persona-engine/src/ambient/domain/event.ts
+import { Schema } from "effect";
+import { createHash } from "node:crypto";
+
+export const SCHEMA_VERSION = "1.0.0";
+
+export function computeEventId(event: MiberaEvent, sourceTag: string): string {
+  const canonical = Schema.encodeSync(MiberaEvent)(event);  // RFC 8785-ish
+  const payload = `${JSON.stringify(canonical)}|${SCHEMA_VERSION}|${sourceTag}`;
+  return createHash("sha256").update(payload).digest("hex");
+}
+```
+
+idempotency falls out: re-fetching events with overlapping cursors
+produces identical IDs; the ring-buffer dedup at `.run/event-cursor.jsonl`
+keys on `(event_id)`.
+
+## score-side terrain (post-DIG · 2026-05-11)
+
+DIG findings against `score-mibera` (the post-rename canonical) +
+`score-dashboard` (Next.js view):
+
+**truth-source map**
+- indexer is **external Envio** (`indexer.hyperindex.xyz` · we don't
+  own indexing)
+- ingestion is **trigger.dev cron every 6 hours** → upserts the
+  `midi_onchain_events` bronze table on shared railway postgres
+- **chains watched**: berachain (mibera NFT `0x6666…c420`, BGT
+  validator, paddleFi, jiko, cubbadges), base (friend.tech subjects)
+- bronze row PK is `id` = `tx_hash + log_index` → **idempotency is
+  free** at the read boundary
+
+**event inventory (already in bronze, NOT yet surfaced per-event)**
+- ✅ mibera mint (`category_key="mibera_minter"`)
+- ✅ mibera buy/sell (`mibera_buyer`, `mibera_seller` — tx-hash cross-ref)
+- ✅ mibera transfer (both sides: `mibera_transfer`, `mibera_transfer_sender`)
+- ✅ burns (`mibera_burner`, `milady_burner`)
+- ✅ fractures/tarot/candies/gifs/cubquest mints (10+ category keys)
+- ✅ loans/liquidations/defaults (paddleFi)
+- ✅ stakes / BGT boost / paddle supply/borrow
+- ✅ badge earns (via `walletBadges.earnedAt`)
+- ❌ **trait shifts** — not indexed today; `fracture_expanded` is
+  static trait snapshots, not deltas
+- ❌ **per-wallet score deltas** — only via weekly digest snapshot diff
+- ❌ proposals / votes — don't exist on chain
+
+**current MCP surface (10 tools, ALL digest-shaped)**
+- everything reads `digestReader.latest(zone, "weekly")` against the
+  `digests` table — no cursor, no since, no per-event shape
+- `get_recent_activity` is misnamed: top-20-by-`numeric1` (biggest),
+  not recent — mints with low/null tokenId systematically miss
+
+## ⚠️ the 6-hour ingestion ceiling
+
+trigger.dev pulls envio → bronze every **6 hours**. this is the real
+ambient-cadence ceiling. **hourly polling on our side is wasteful
+unless score-mibera tightens ingestion.**
+
+**three paths the operator must choose between** (round-2 question):
+
+| path | tradeoff |
+|---|---|
+| **A · accept 6h cadence** | rave-feel composes fine with discrete 6h drops + kansei half-life; minimal infra change; can tighten later |
+| **B · push ingestion to hourly** | requires score-mibera trigger.dev cron change; possibly envio rate-limit consideration; more "live" feel |
+| **C · live envio subscription on freeside side** | bypass bronze; subscribe directly to envio for the hot classes (mints/burns); heaviest infra; matches purupuru WebSocket pattern exactly |
+
+**recommendation:** path A. compose around 6h cadence; the kansei
+half-life is already 6h in the brief. discrete drops at 6h intervals
+are *more* rave-shaped than continuous streams — drops are the
+mechanic. push later only if validated need emerges.
+
+## score-side asks (concrete · post-DIG)
+
+new MCP tools to author in `score-mibera/src/mcp/tools/` (registered
+in `src/mcp/server.ts`):
+
+| tool | priority | dependencies |
+|---|---|---|
+| **`get_events_since({since_ts, limit, zone?, classes?})`** | P0 | new `src/mcp/event-reader.ts` mirroring `digestReader`; class-mapping at `src/db/schema/event-classes.ts` |
+| **`get_event_by_id({event_id})`** | P0 | 1-line drizzle lookup |
+| **`get_recent_mints({collection?, limit})`** | P1 | category_key allowlist filter; **MIBERA-id resolution unblocked here only if we add midi-interface join** |
+| **`list_event_classes()`** | P1 | static catalog; symmetry with `list_dimensions` |
+| **`stream_events_by_class({class, since})`** | P2 | composable from `get_events_since` server-side; could skip if generic stream is enough |
+
+**class taxonomy decision** (round-2 question): the bronze row stores
+`metadata.actionType` as JSON-stringified blob. mapping bronze
+category_keys → human-readable classes (`mint | transfer | burn |
+loan | stake | badge`) lives where?
+
+- option 1: **server-side mapping table** at `score-mibera/src/db/schema/event-classes.ts` (mirrors `factor_category_mapping`)
+- option 2: **MCP-tool-side derivation** (read raw `category_key`, map at response time)
+- option 3: **consumer-side derivation** (freeside-characters maps in `ambient/domain/event.ts`)
+
+**recommendation:** option 1 — class taxonomy belongs to truth, not
+voice. consumers receive pre-classed events.
+
+## the MIBERA-id resolution gap
+
+bronze stores `tokenId` only; codex's `MIBERA-XXXXXX` slug lookup
+lives in `midi-interface` DB. without resolution, a mint event reads:
+
+```json
+{ "class": "mint", "wallet": "0xabc...", "tokenId": 4221, ... }
+```
+
+but ruggy/satoshi narration wants:
+
+> "MIBERA-004221 just landed in honeyroad — ⛰️ ancestor, 🜍 element"
+
+**options** (round-2 question):
+1. add midi-interface join inside score-mibera (cross-DB query;
+   schema coupling)
+2. consumer-side resolution in `ambient/router.system.ts` via
+   `mcp__codex__lookup_by_tokenid({tokenId})` — does this MCP tool
+   exist on construct-mibera-codex?
+3. defer — surface raw tokenId, narrate without codex flavor at first
+   ship; add resolution V0.7-A.5+
+
+**recommendation:** option 2 if codex-mcp has the tool; option 3
+otherwise. cross-DB joins in option 1 violate separation.
+
+## butterfreezone scope
+
+- **score-mibera** already has `BUTTERFREEZONE.md` at repo root. it's a
+  *framework* not a *construct pack* (no `.claude/constructs/packs/`
+  structure). regenerate after the new MCP tools land so the surface
+  inventory is fresh.
+- **freeside-characters** regenerate after the new `ambient/` module
+  is authored.
+- **typed-stream contracts** for score-emitted events live on the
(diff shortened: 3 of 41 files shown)


zkSoju (commented):
## Summary

This PR carries two cycles in one cargo hold — an ambient-events scaffold plus a sanitizer hardening from an unrelated bug branch — which complicates review, revert, and bisection. The sanitizer work is genuinely excellent (layered Unicode defense, idempotency invariants, documented-failure-mode tests), but the ambient/ scaffold ships ~2000 lines of new domain code with no co-located tests visible and several PRD-mandated CI gates whose wiring isn't confirmed in the diff.

## Architectural Meditation

The sanitizer story here is the kind of work that earns trust over years. NFKC → `\p{Cf}` strip → `trimStart` → anchored regex is textbook defense-in-depth: each layer assumes the next will fail. This is the same philosophy that drove Google's Caja, the Chromium sandbox layering, and Linux seccomp-bpf — *defenses compose, single defenses fail*. The idempotency test (F7) and the "keep the buggy ordering as a test" pattern (F8) are the sort of breadcrumbs Hyrum Wright writes papers about.

The ambient/ scaffold is architecturally promising — ports/adapters with live/mock split is the Effect-TS Layer pattern done right, echoing Netflix's Hexagonal adoption circa 2018. But scaffold without tests is a promise, not a system. Google's Testing on the Toilet team has a one-liner for this: *"untested code is wishful thinking with syntax highlighting."*

The PR's biggest risk isn't any single finding — it's **scope conflation**. When the sanitizer regresses in six months, `git bisect` will land on a commit titled "ambient-events-as-rave" and the engineer on-call will spend an hour wondering why.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "PR scope mismatch — title claims ambient-events but diff is mostly sanitizer + planning docs",
      "severity": "HIGH",
      "category": "scope",
      "dimension": "[Quality]",
      "file": "PR metadata",
      "description": "PR title advertises ambient-events scaffolding but the bulk of code-with-tests changes (dispatch.ts, sanitize.ts, dispatch-error-routing.test.ts, sanitize.test.ts) belong to bug-20260511-b6eb97 — an unrelated sanitizer cycle.",
      "suggestion": "Split into two PRs: sanitizer fix + ambient scaffold. At minimum, document the dual scope explicitly in the PR description.",
      "faang_parallel": "Google's eng-prod team enforces single-purpose CLs precisely so `git bisect` lands on a meaningful commit. Facebook's Phabricator was built around the 'one idea per diff' rule for the same reason.",
      "metaphor": "Like shipping a kitchen renovation and a roof repair in one invoice — both might be excellent work, but when the roof leaks next winter you'll be paying a plumber to diagnose shingles.",
      "teachable_moment": "PR scope is a debugging affordance for future-you. The cost of splitting is paid once; the cost of conflation is paid every bisect.",
      "decision_trail": "If shipping together is intentional (shared release train), document WHY in the PR body so future archaeologists don't assume sloppiness.",
      "confidence": 0.9
    },
    {
      "id": "F2",
      "title": "Ambient module ports + adapters added without unit/integration tests in this PR",
      "severity": "HIGH",
      "category": "testing",
      "dimension": "[Test Coverage]",
      "file": "packages/persona-engine/src/ambient/",
      "description": "~2000 lines under ambient/ ship with no visible co-located tests. Sprint S1.T1–S4.T4 mandate property-based tests, FAGAN gates, schema decode tests, and ≥85% coverage (NFR-5). None appear here.",
      "suggestion": "Add tests before merge OR mark this PR explicitly 'scaffold-only, tests follow in S1.T1–S1.T4' AND gate downstream merges on the test PR landing. Confirm CI coverage threshold actually scopes ambient/.",
      "faang_parallel": "AWS S3's launch postmortem culture: every new subsystem ships with chaos tests on day one. The 2017 S3 outage retrospective explicitly cited untested operator tooling as a contributor.",
      "metaphor": "Building the skeleton of a bridge and inviting traffic before any load testing — the architecture might be perfect, but you won't know until something cracks.",
      "teachable_moment": "'Tests follow' is the most common lie in software. Code without tests at merge tends to stay that way; the cognitive cost of writing tests retroactively is 3-5x.",
      "confidence": 0.85
    },
    {
      "id": "F3",
      "title": "FAGAN canon-forbidden-word gate not visible in CI config",
      "severity": "MEDIUM",
      "category": "tooling",
      "dimension": "[Operational]",
      "file": ".loa.config.yaml / ci",
      "description": "PRD S6 / Sprint S4.T4 mandate a CI grep gate for forbidden canon words. The diff adds .loa.config.yaml content but the gate's presence isn't confirmed in the visible portion.",
      "suggestion": "Verify the FAGAN gate is wired AND failing-mode tested (introduce a forbidden word, watch CI fail). Same for the singleton-invariant doc check and ManagedRuntime.make count check (S9).",
      "faang_parallel": "Google's pre-submit lint rules (Tricorder) — a guard that exists but doesn't fail builds is worse than no guard, because it creates false confidence.",
      "metaphor": "A smoke detector you've never tested. The wiring looks right, but the only proof is the smoke.",
      "decision_trail": "Include a 'gate validation' commit in CI history that intentionally fails the gate, then reverts — proves the gate works.",
      "confidence": 0.6
    },
    {
      "id": "F4",
      "title": "EVENT_LATE_ARRIVAL_HOURS defined but no implementation evidence",
      "severity": "MEDIUM",
      "category": "correctness",
      "dimension": "[Quality]",
      "file": ".env.example",
      "description": "IMP-013 splits late-arrival threshold from half-life. NFR-15 requires explicit rejection at this boundary. Visible code doesn't confirm the env var is read separately from EVENT_RAVE_HALF_LIFE_HOURS.",
      "suggestion": "Verify consumption in event-source.live.ts or router.system.ts; add a fixture test asserting events older than the threshold are rejected with a `late_arrival` log/metric.",
      "metaphor": "A door labeled 'emergency exit' that's actually just painted on the wall. The signage is correct; the implementation is missing.",
      "confidence": 0.5
    },
    {
      "id": "F5",
      "title": "Singleton invariant relies on flock but no startup acquisition visible",
      "severity": "MEDIUM",
      "category": "deployment",
      "dimension": "[Operational]",
      "description": "NFR-25 mandates fail-loud crash if a second instance can't acquire the cursor flock at boot. runtime.ts (45 lines) doesn't visibly show LOCK_EX|LOCK_NB acquisition with the documented error message.",
      "suggestion": "Confirm boot-time flock acquisition with explicit exit message. Add an integration test spawning two processes; assert second exits non-zero with the canonical message.",
      "faang_parallel": "Kubernetes leader election via lease objects — the entire reconciliation loop pattern is built on the premise that boot-time leadership acquisition is loud and unambiguous. Silent failover is how etcd split-brains happen.",
      "metaphor": "Two captains on a ship is fine if they meet in the wheelhouse and one steps off. It's catastrophic if neither realizes the other is there.",
      "confidence": 0.55
    },
    {
      "id": "F6",
      "title": "Sanitizer NFKC + \\p{Cf} strip + trimStart is well-layered defense-in-depth",
      "severity": "PRAISE",
      "category": "security",
      "dimension": "[Security]",
      "file": "packages/persona-engine/src/deliver/sanitize.ts",
      "description": "Three orthogonal Unicode defenses compose cleanly before anchored regex. Probe-vs-content separation preserves legitimate output. Evolution from ZWSP-only → Cf class reflects real adversarial pressure-testing.",
      "suggestion": "Promote this pattern to vault doctrine as the canonical 'chat-medium write boundary' recipe.",
      "faang_parallel": "Mirrors Google Caja's HTML sanitizer layering and the Unicode TR39 confusables detection used in Chrome URL anti-spoofing. The 'normalize before match' rule is a Unicode security gospel.",
      "metaphor": "Three locks on a door, each picked by a different key. An attacker carrying the wrong toolkit goes home.",
      "confidence": 0.9
    },
    {
      "id": "F7",
      "title": "Sanitizer idempotency invariant test guards against template regression",
      "severity": "PRAISE",
      "category": "testing",
      "dimension": "[Test Coverage]",
      "file": "packages/persona-engine/src/deliver/sanitize.test.ts",
      "description": "The invariant 'error template output must not match any raw error pattern' catches a subtle future regression: edit composeErrorBody to start with 'Error:' and sanitizer loops.",
      "suggestion": "Enumerate characters dynamically from config rather than hardcoded ['ruggy','satoshi','unknown-character'], so new characters auto-join the invariant.",
      "faang_parallel": "This is the Hyrum's-Law-as-test pattern — make implicit contracts explicit so future refactors trip the wire.",
      "metaphor": "A tripwire across the path you'd accidentally walk down. Cheap to install, priceless when triggered.",
      "confidence": 0.9
    },
    {
      "id": "F8",
      "title": "Codex C1 prefix-ordering regression test documents the bug + correct ordering",
      "severity": "PRAISE",
      "category": "testing",
      "dimension": "[Test Coverage]",
      "file": "apps/bot/src/tests/dispatch-error-routing.test.ts",
      "description": "Asserts the correct path AND keeps the buggy path as a documentation test. Inline comments explain why sanitize must precede prefix application (^-anchored regex).",
      "suggestion": "Promote this pattern repo-wide: 'document the failure mode as a test' is institutional memory in code form.",
      "faang_parallel": "Linux kernel's `selftests/` ethos — every patched CVE leaves behind a test that would have caught it. The bug becomes a permanent immune memory.",
      "metaphor": "A scar that teaches. The wound healed, but the body remembers exactly how it happened.",
      "confidence": 0.95
    },
    {
      "id": "F9",
      "title": "Three flatline diff dumps (~4000 lines) checked into grimoires",
      "severity": "LOW",
      "category": "repo-hygiene",
      "dimension": "[Quality]",
      "file": "grimoires/loa/a2a/bug-20260511-b6eb97/flatline-input/",
      "description": "Three near-identical diff dumps (v2/v3/v4) totaling 1100–1470 lines each, committed as review-tool inputs.",
      "suggestion": "Keep only v4, OR move to an archive path, OR .gitignore flatline-input/ if these are scratch-pad inputs.",
      "metaphor": "Saving every draft of an email in your sent folder. Useful for the moment, noise forever.",
      "confidence": 0.6
    },
    {
      "id": "F10",
      "title": "sprint.md replaced entirely — historical sprint moved to sprint-v07a1.md",
      "severity": "LOW",
      "category": "documentation",
      "dimension": "[Quality]",
      "file": "grimoires/loa/sprint.md",
      "description": "524-line V0.7-A.1 sprint replaced with cycle-003 plan. Archive exists at sprint-v07a1.md but the swap is non-obvious for reviewers landing fresh.",
      "suggestion": "Add a prominent breadcrumb at top of new sprint.md: 'Previous cycle archived → sprint-v07a1.md'.",
      "confidence": 0.5
    },
    {
      "id": "F11",
      "title": "Property test seed 1337 + 200-line corpus pinned but corpus not visible",
      "severity": "LOW",
      "category": "testing",
      "dimension": "[Test Coverage]",
      "file": "packages/persona-engine/tests/ambient/fixtures/canon-corpus.txt",
      "description": "S1.T2 mandates the corpus file + seed for FORBIDDEN_REGEX false-positive guard. File not in visible diff.",
      "suggestion": "Confirm corpus ships here, or explicitly defer to follow-up PR with a tracked issue link.",
      "confidence": 0.5
    },
    {
      "id": "F12",
      "title": "Sprint auto-integrates upgraded-triad findings without explicit re-approval",
      "severity": "REFRAME",
      "category": "process",
      "dimension": "[Quality]",
      "file": "grimoires/loa/sprint.md (§5.1)",
      "description": "6 HIGH_CONSENSUS items auto-integrated into binding ACs. For a cycle with 24 locked decisions + ~32 NFRs, silent absorption risks model-cargo-cult.",
      "suggestion": "Add operator sign-off line in NOTES.md or ledger.json referencing the triad run.",
      "faang_parallel": "Amazon's PR/FAQ doc process — every binding decision has a named approver. AI-suggested changes need the same human pin.",
      "metaphor": "Letting the consultant rewrite the contract while you're at lunch. Even good suggestions deserve a signature.",
      "confidence": 0.4
    },
    {
      "id": "F13",
      "title": "router.system.ts shared inter-character refractory — flock scope uncertain",
      "severity": "SPECULATION",
      "category": "correctness",
      "dimension": "[Quality]",
      "file": "packages/persona-engine/src/ambient/router.system.ts",
      "description": "S3.T2a routes through pop-in-ledger getLastFire. The race is check-then-write across two characters. Flock must span the whole sequence, not per-write.",
      "suggestion": "Verify flock wraps getLastFire→decide→writeLedger atomically. Add a fixture test simulating concurrent ruggy+satoshi fires on the same event; assert exactly one ledger row.",
      "faang_parallel": "Classic TOCTOU (time-of-check/time-of-use) — the bug that haunted Unix setuid binaries for decades. Per-write locking is necessary but never sufficient.",
      "metaphor": "Two people checking if the last cookie is gone, then both reaching. The check and the take must be one motion.",
      "confidence": 0.4
    },
    {
      "id": "F14",
      "title": "Sanitizer telemetry uses console.error — may inflate error-level metrics",
      "severity": "LOW",
      "category": "observability",
      "dimension": "[Operational]",
      "file": "packages/persona-engine/src/deliver/sanitize.ts",
      "description": "If sanitizer is pure belt-and-suspenders (zero firings expected), error-level is correct. If steady-state firings exist, this floods alerts.",
      "suggestion": "Monitor firing rate post-deploy. Zero over 30 days = keep. Non-zero = investigate the upstream leak.",
      "metaphor": "A fire alarm in a kitchen that burns toast often. The alarm works; the kitchen needs work.",
      "confidence": 0.5
    },
    {
      "id": "F15",
      "title": "bloom-filters package choice + version pin not visible",
      "severity": "LOW",
      "category": "dependencies",
      "dimension": "[Operational]",
      "file": "packages/persona-engine/package.json",
      "description": "Multiple npm packages named bloom-filters exist. NFR-27 guarantees 0.1% FP rate; IMP-002 requires serialization compat across restarts.",
      "suggestion": "Document chosen package, maintenance status, license, and serialization-stability commitment in NOTES.md.",
      "confidence": 0.5
    }
  ]
}
```


## Callouts

**F6, F7, F8** are the heart of this PR's excellence. The sanitizer's three-layer Unicode defense, the idempotency invariant, and the "buggy-ordering preserved as documentation test" pattern collectively demonstrate the kind of adversarial rigor that distinguishes production-grade code from prototype code. The flatline-input artifacts (F9 notwithstanding) show genuine iterative pressure-testing — v2 → v3 → v4 reflects real learning. Promote the three-layer sanitizer pattern to vault doctrine; it earns reuse.

**Verdict**: REQUEST_CHANGES — primarily for scope split (F1) and test landing strategy (F2). The sanitizer work itself is merge-ready; the ambient scaffold needs either tests or an explicit deferral plan.

zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#55. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Cursor never advances — saves invalid empty event_id and only on success path",
      "severity": "CRITICAL",
      "category": "correctness",
      "description": "_initialCursor() creates a cursor with `event_id: \"\" as never`. The pulse-sink mock and live both store stir keyed by zone but the cursor logic in runStirTick only saves the cursor when `result.cursor_advanced && result.nextCursor`, where cursor_advanced is gated on `fetched.events.length > 0`. If the EventFeed returns 0 events repeatedly (the common path on 6h ingestion), the cursor never advances past its initial position (`now - 60s`). On the next tick the `since_ts` is computed from the OLD cursor's event_time, which is the prior tick's `now - 60s` — so each hourly tick re-queries a window that drifts ever further behind real time. The cursor only advances on the rare ticks that fetch events, never on idle ticks. This silently breaks NFR-12/13 cursor mechanics under sparse-event conditions, which is the steady-state.",
      "suggestion": "Always advance the cursor's `event_time` to `now - REPLAY_WINDOW_SECONDS` on every successful tick (even with zero events), preserving last-known `event_id` for tiebreaking. Or persist the cursor returned by `fetchSince` even when events.length === 0 (EventSourceLive already returns a nextCursor on the empty path, but with `raw.next_cursor.event_time` from score-mcp — which depends on score-mcp's behavior on empty windows).",
      "confidence": 0.85
    },
    {
      "id": "F2",
      "title": "Wallet redaction (NFR-29) is wired in domain/ports but NEVER invoked in router or composer path",
      "severity": "CRITICAL",
      "category": "security",
      "file": "packages/persona-engine/src/ambient/router.system.ts",
      "description": "WalletResolverLive is composed into AmbientLayer, but `router.system.ts` never calls `WalletResolver.resolve()` before emitting decisions. The router writes LedgerEntry records with `event_id` and `event_class` but does not resolve wallet handles. More importantly, there is no composer/narrator code shown in this PR that consumes the router decision and renders narration — the wallet redaction invariant claimed in SDD §8/NFR-29 (\"three layers: router calls walletResolver BEFORE composer\") is not enforced anywhere in the diff. PRD G4/NFR-29 explicitly states this is load-bearing for CLAUDE.md compliance, but the implementation defers it. CI grep mentioned in S4.T4 is also not present in the diff.",
      "suggestion": "Either (a) implement the router→walletResolver→composer flow in this PR with explicit tests, or (b) clearly mark this PR as scaffolding-only and gate any composer wiring on a follow-up PR before any narration ships. The current state risks shipping ambient narration paths that bypass the wallet redaction layer.",
      "confidence": 0.8
    },
    {
      "id": "F3",
      "title": "MCP transport security (NFR-30) endpoint allowlist + TLS verification not implemented",
      "severity": "HIGH",
      "category": "security",
      "file": "packages/persona-engine/src/ambient/live/score-mcp-client.ts",
      "description": "SDD §8 and NFR-30 require boot-time validation: (1) reject http:// at parse, (2) hostname must be in MCP_ENDPOINT_ALLOWLIST, (3) TLS cert chain validated with no skipCertVerify. The score-mcp-client.ts uses fetch() with the raw `${config.SCORE_API_URL}/mcp` without any of these checks. Same for codex/freeside-auth resolvers which reuse this client. The Flatline blocker SKP-002 (870) was marked ACCEPT in SDD §12.1 but the implementation is missing.",
      "suggestion": "Add boot-time URL validation: assert https://, parse hostname, check against process.env.MCP_ENDPOINT_ALLOWLIST, crash explicitly on mismatch. Add to runtime.ts startup or a dedicated config-validation module called before runtime construction.",
      "confidence": 0.9
    },
    {
      "id": "F4",
      "title": "Late-arrival rejection (NFR-15) and bloom filter spillover (NFR-27) declared in domain but not used",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/domain/cursor.ts",
      "description": "`isLateArrival()` and the EventCursorSeen bloom filter infrastructure are defined but never invoked. EventSourceLive does not filter late-arrival events; pulse.system.ts does not check; scheduler-task.ts does not maintain the seen-ring. NFR-13 dedup is the load-bearing mechanism for NFR-31 reorg-aware late-arrival policy. Without it, the overlap window (REPLAY_WINDOW_SECONDS=60) will cause duplicate ingestion on every tick (events at the boundary are fetched twice).",
      "suggestion": "Wire dedup at EventSourceLive boundary: maintain EventCursorSeen state, check `seen.ids.includes(id) || bloom.has(id)` before decoding, skip duplicates. Apply isLateArrival check before adding to pulse input.",
      "confidence": 0.9
    },
    {
      "id": "F5",
      "title": "Singleton flock invariant (NFR-22) not actually enforced — no flock calls anywhere",
      "severity": "HIGH",
      "category": "correctness",
      "description": "PRD NFR-21–25 elevate the singleton invariant to a CRITICAL load-bearing constraint with NFR-25 claiming the second instance will CRASH at startup via flock failure. But no live adapter actually calls flock or proper-lockfile. pop-in-ledger.live, circuit-breaker.live, and scheduler-task.ts cursor writes all use plain fs.writeFileSync + renameSync without any locking. proper-lockfile was specified in SDD §9 as a new dep but is not imported anywhere in the diff. NFR-25 fail-loud guarantee is not delivered.",
      "suggestion": "Either (a) implement proper-lockfile acquisition at runtime startup on a sentinel file in .run/, with LOCK_EX|LOCK_NB semantics, exiting with clear error on conflict, or (b) downgrade the singleton invariant claim in SDD/PRD to documentation-only and remove NFR-25's fail-loud promise.",
      "confidence": 0.95
    },
    {
      "id": "F6",
      "title": "Atomic append is not atomic — reads entire file into memory and rewrites",
      "severity": "HIGH",
      "category": "correctness",
      "description": "_atomicAppend reads the entire existing file into memory, concatenates the new line, writes to a tmp file, then renames over the canonical path. This is correct for crash-consistency but: (1) it's O(file_size) per append, making ledger writes increasingly expensive over a month; (2) it loads the whole file into a string, problematic at scale; (3) concurrent appends from within the same process race between read and rename (Node is single-threaded for sync ops, but multiple Effect fibers can interleave around the sync section if any await crosses it — here it's all sync so safe within one process, but cross-process protection requires flock which isn't present). Same pattern in circuit-breaker.live.ts and scheduler-task.ts _saveCursor.",
      "suggestion": "For append-only JSONL, use fs.appendFileSync with O_APPEND|O_SYNC flags (POSIX guarantees atomic appends ≤ PIPE_BUF). The rename pattern is only correct for whole-file replacements; for append it's both slow and not actually safer. Pair with flock for cross-process safety.",
      "confidence": 0.85
    },
    {
      "id": "F7",
      "title": "Decay applied to negative press values pulls toward STIR_FLOOR from below, which is wrong",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/pulse.system.ts",
      "description": "In _gravityIfStillFresh and pulseTick, press decay uses `decayValue(stir.press, ...)` which computes `STIR_FLOOR + (current - STIR_FLOOR) * decayFactor`. For inner_sanctum negative press (D11 inversion), e.g. press = -0.8, this decays toward STIR_FLOOR (positive 0.05) by pulling upward — but the semantic intent is that as the negative-press transient fades, the inner_sanctum returns to its quiet baseline. STIR_FLOOR at 0.05 (positive) is the wrong attractor for a signed axis. Either the floor should be 0 / signed, or press should decay toward a primitive-specific baseline.",
      "suggestion": "Make the decay attractor primitive-aware: inner_sanctum decays toward 0 (or a small negative baseline), others toward STIR_FLOOR positive. Or store press as |press| + sign separately. Add a fixture test exercising the inner_sanctum decay path across multiple ticks.",
      "confidence": 0.7
    },
    {
      "id": "F8",
      "title": "scheduler stir task acquires no zone lock — concurrent with digest/pop-in tasks",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/cron/scheduler.ts",
      "description": "The new ambient-stir task checks `if (zoneLocks.has(zone)) continue;` but does not ACQUIRE the lock for the duration of its work. R-SDD-3 explicitly flagged this as a risk; the mitigation said \"use the same zoneLocks Map; document interaction\" but the diff only checks the lock, it does not hold it. If digest or pop-in fires mid-stir-tick, both may write to rosenzu state. Also, the check-then-skip pattern means a long-running digest will cause the stir tick to be silently dropped for that hour, with no retry. Combined with F1, missed stir ticks compound cursor drift.",
      "suggestion": "Either acquire the zone lock for the duration of runStirTick (acceptable if stir work is fast and non-overlapping), or use a stir-specific lock distinct from digest/pop-in's. Add a metric for skipped ticks per zone.",
      "confidence": 0.7
    },
    {
      "id": "F9",
      "title": "FORBIDDEN_REGEX is overly broad — matches \"era\" as substring of common English words",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/domain/canon-vocabulary.ts",
      "description": "The forbidden pattern uses `\\bera\\b` and `\\bmolecule\\b` with word boundaries — good. But `founder.*archetype` is `.*` greedy with no anchors, will match across hundreds of characters of unrelated text. And `\\bera\\b` will match the literal word \"era\" in any narration where ruggy/satoshi might legitimately use it descriptively (\"a new era of\") — the canon-codex restriction is specifically against `era` as a FIELD NAME in code, not the English word in prose. The corpus test (canon-corpus.txt) may pass today but the regex semantics will produce false positives the moment ruggy uses common English.",
      "suggestion": "Narrow the regex to code-level violations (e.g. field access like `\\.era\\b` or `era:` in object literals) for CI grep, and use a different prose-level check for narration audits. Or run the regex only against domain/event/canon files, not against narration output.",
      "confidence": 0.7
    },
    {
      "id": "F10",
      "title": "MiberaResolverLive calls `lookup_mibera` on score-mcp endpoint, but tool lives on codex-mcp",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/live/mibera-resolver.live.ts",
      "description": "The brief and SDD §4.2 say mibera lookup is via `mcp__codex__lookup_mibera` against construct-mibera-codex. But MiberaResolverLive imports `callScoreToolAmbient` and uses `config.SCORE_API_URL` — calling score-mcp, not codex-mcp. This will fail in production (tool not found on score-mcp). WalletResolverLive has the same issue (calls callScoreToolAmbient for resolve_wallet, which lives on freeside-auth-mcp).",
      "suggestion": "Introduce separate MCP client wrappers per endpoint (score / codex / freeside-auth) reading from CODEX_MCP_URL and FREESIDE_AUTH_MCP_URL config keys respectively. callScoreToolAmbient should only be used for score-mcp tools.",
      "confidence": 0.9
    },
    {
      "id": "F11",
      "title": "appendIfNoFire race window is wider than documented — `proposedEntry.ts <= now` is not enforced",
      "severity": "MEDIUM",
      "category": "correctness",
      "description": "appendIfNoFire checks for blockers in `(afterTs, proposedEntry.ts]`. If two characters compute their proposedEntry.ts at slightly different wall-clock moments (e.g. ruggy = 14:00:00.500, satoshi = 14:00:00.501), the second to write sees the first's entry in window and yields. But there's no flock around the read-then-write — within one process Node sync ops are atomic per-call but across the read of `_readAllAcrossArchives` and the write of `_atomicAppend`, multiple Effect fibers can interleave if either operation awaits. As implemented (all sync), it's safe single-process, but the comment promises cross-process safety via the singleton invariant, which depends on F5 actually being enforced.",
      "suggestion": "Add explicit comment that this is single-process atomic only by Node's sync I/O guarantees, and that cross-process safety REQUIRES flock (F5). Once flock is in, wrap appendIfNoFire's read+write in a single flock-acquired section.",
      "confidence": 0.6
    },
    {
      "id": "F12",
      "title": "GravityChannel `last_significant_event_within_window` is set to true in pulseTick but never cleared on subsequent ticks without new events",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/pulse.system.ts",
      "description": "When a class-A event lands, gravityChannel.last_significant_event_within_window = true. On subsequent ticks with NO new events, pulseTick takes the `_gravityIfStillFresh` branch which correctly checks elapsed time. But the carry-forward only runs when `base.gravity.last_significant_event_within_window === true`. The decay block calls `decayValue(input.previousStir.press, ...)` but `base.gravity = input.previousStir.gravity` unchanged. So gravity correctly carries forward and gets re-evaluated. However, D21 says \"ONE tick of fire, then off (not a decay curve)\" — but here gravity stays true for up to GRAVITY_WINDOW_MINUTES (60 min default), which is multiple ticks if scheduler runs every minute (it runs hourly per default, so usually 1 tick — but the semantics document a window, not a single tick).",
      "suggestion": "Clarify documentation: gravity is a WINDOW flag (active for GRAVITY_WINDOW_MINUTES after fire), not a one-tick transient. Either align D21 prose with implementation or implement true one-tick behavior by clearing on the next tick regardless of window.",
      "confidence": 0.5
    },
    {
      "id": "F13",
      "title": "Router's getLastFire path described in comments is replaced by appendIfNoFire — but inter-character race window is still TOCTOU",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/router.system.ts",
      "description": "The comment in routerDecide says inter-character coordination is deferred to appendIfNoFire. That's a reasonable shift. However, the decision tree still emits a `fired` decision before any atomic check happens. The caller must then route through appendDecision/appendIfNoFire. If a caller forgets and uses plain `append`, the inter-character invariant breaks silently. Without enforcement at the type level, this is a footgun — and there is no compose/reply.ts or other narration path in this diff to verify the discipline holds.",
      "suggestion": "Make the router decision a non-public type and force all writes through a single function that selects append vs appendIfNoFire based on decision class. Or have the router return a thunk that performs the atomic write rather than a Decision struct.",
      "confidence": 0.55
    },
    {
      "id": "F14",
      "title": "S1.T2/S4.T4 FAGAN gate CI step is described in sprint plan but not present as a check in this PR",
      "severity": "MEDIUM",
      "category": "process",
      "file": "grimoires/loa/sprint.md",
      "description": "Sprint S4.T4 specifies three CI grep gates (forbidden-word, single ManagedRuntime.make, task_count). None of these appear as CI workflow files or scripts in the diff. The verification gates described in PRD §6.4 and SDD verification sections are mentioned only in docs. Without a CI binding, these gates are aspirational. PR review process is the only enforcement.",
      "suggestion": "Add a script (e.g. scripts/fagan-gates.sh) invoked by CI that runs the three grep checks and fails on violation. Reference it from the package.json check script. This is mentioned in sprint S4.T4 as in-scope but absent from the diff.",
      "confidence": 0.95
    },
    {
      "id": "F15",
      "title": "EVENT_LATE_ARRIVAL_HOURS env var documented but not yet honored anywhere except a helper",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/domain/cursor.ts",
      "description": "lateArrivalRejectHoursFromEnv() is defined and isLateArrival() reads it as default, but no caller invokes isLateArrival(). Same fate as F4 — the env var is documented in .env.example and consumed by a helper, but the code path doesn't apply the policy.",
      "suggestion": "Wire isLateArrival into EventSourceLive's per-event filter loop alongside the unknown-class quarantine path.",
      "confidence": 0.9
    },
    {
      "id": "F16",
      "title": "loadConfig() called per-tick inside Effect.gen blocks — re-parses env every call",
      "severity": "LOW",
      "category": "performance",
      "file": "packages/persona-engine/src/ambient/live/event-source.live.ts",
      "description": "EventSourceLive, MiberaResolverLive, and WalletResolverLive each call `loadConfig()` inside their Effect.gen flows. Depending on loadConfig's implementation (not shown), this may re-parse environment variables, re-validate Zod schemas, etc. on every tick — wasted work and a latent source of surprise if env mutates at runtime.",
      "suggestion": "Capture config once at Layer construction time and close over it. Pattern: `Layer.effect(EventFeed, Effect.gen(function*(_){ const config = loadConfig(); ... return EventFeed.of({...})}))`.",
      "confidence": 0.7
    },
    {
      "id": "F17",
      "title": "Massive design-doc commit (~2,500 lines) shipped alongside implementation creates review and revision burden",
      "severity": "LOW",
      "category": "process",
      "file": "grimoires/loa/context/ambient-events-as-rave.md",
      "description": "The PR adds 818 lines of brief, 750 lines of SDD, 502 lines of sprint plan, and 442 lines of PRD — over 2,500 lines of design docs alongside code. These docs are valuable as artifacts but: (1) freeze decisions in markdown that are quickly out of sync with code (e.g. F2/F3/F4/F5 already diverge), (2) review fatigue at this volume means reviewers skim or miss real issues, (3) inline references like \"per F13 closure\" and \"Flatline SKP-003 730\" are illegible to anyone outside the workflow. The Loa workflow appears to be generating boilerplate decision logs at high velocity.",
      "suggestion": "Consider whether all four docs need to ship in one PR. The brief is the only context-binding doc; PRD/SDD/sprint can live in a separate non-code PR or in an archive folder. Aggressively trim Flatline blocker logs to capture only outcomes, not negotiation history.",
      "confidence": 0.4
    },
    {
      "id": "F18",
      "title": "Effect Schema KansaiStir schema uses `Schema.between(STIR_FLOOR, 1)` but pulse.system.ts can produce values above 1 due to additive deltas",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/pulse.system.ts",
      "description": "pulseTick does `Math.min(1, Math.max(STIR_FLOOR, ...))` on strangers and drift, clamping to [STIR_FLOOR, 1]. But during accumulation of multiple events the intermediate value may exceed 1 before clamp. The Schema.between([0.05, 1]) check would fail decode of any KansaiStir produced before clamp. Since clamp is applied before returning, this likely works in practice — but: (1) the +0.1 observedDriftRatio bonus is added INSIDE the clamp call, so a stir already at 1.0 + delta could lose the bonus silently; (2) decay sets press without clamp, which is intentional per F7, but then `base.strangers + delta.strangers + observedDriftRatio * 0.1` clamps strangers high — meaning strangers can saturate at 1.0 every tick under high event volume, losing dynamic range.",
      "suggestion": "Add a fixture test exercising saturation: 100 events in one tick, assert strangers stays at 1.0 but drift differentiates. Consider whether saturation at 1.0 is the right behavior, or whether a softer asymptote (e.g. tanh-shaped) would preserve more signal.",
      "confidence": 0.55
    },
    {
      "id": "F19",
      "title": "Solid separation-of-concerns scaffolding: four-folder discipline + sealed unions + Context.Tag ports",
      "severity": "PRAISE",
      "category": "architecture",
      "file": "packages/persona-engine/src/ambient/",
      "description": "The domain/ports/live/mock layout is well-executed: domain files are pure schema, ports are Context.Tag service interfaces, live adapters compose via Layer.effect/Layer.succeed, mocks have parallel shapes. The single-ManagedRuntime.make site discipline is followed. Suffix-as-type lets grep enumerate the behavior surface. This is a clean Effect-TS application of construct-effect-substrate.",
      "suggestion": "Maintain the discipline as composer/narration paths are added — the test suite should fail if a port has no mock or two ManagedRuntime.make calls appear."
    },
    {
      "id": "F20",
      "title": "Canon vocabulary table with explicit forbidden-word list and lore notes is a thoughtful constraint surface",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/persona-engine/src/ambient/domain/canon-vocabulary.ts",
      "description": "The translation table captures real lore constraints (no \"sacrifice\" for burn, no \"migration\" for transfer) with citation comments, and exposes a programmable check surface. Even with F9's regex caveat, this is the right place for canon enforcement to live."
    },
    {
      "id": "F21",
      "title": "Stir as sibling channel (not feel mutation) preserves the canonical KansaiVector invariant",
      "severity": "PRAISE",
      "category": "architecture",
      "file": "packages/persona-engine/src/orchestrator/rosenzu/server.ts",
      "description": "The decision (D4) to extend furnish_kansei's response with `stir` + `rendered_modulation` rather than mutating KansaiVector.feel is the right architectural call. Categorical bias strings (D24) over numeric leakage to the LLM is a strong invariant. The fallback (`} catch { // ambient module not loaded — pre-cycle-003 fallback`) gracefully degrades to pre-cycle behavior, which is good rollback hygiene."
    }
  ]
}
```


## Callouts

zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#55. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Cursor advance writes the wrong cursor when events fetched",
      "severity": "CRITICAL",
      "category": "correctness",
      "description": "In runStirTick, the program's success branch returns nextCursor only when events were fetched (cursor_advanced=true). But the saved cursor is `result.nextCursor` only on the cursor_advanced branch. Worse: the idle-path advance constructs a new cursor by rolling event_time forward to `now - REPLAY_WINDOW_SECONDS`, preserving the prior event_id. After computeOverlapSince subtracts REPLAY_WINDOW_SECONDS again on the next call, the since_ts becomes `now - 2*REPLAY_WINDOW_SECONDS` (= now - 120s). This means each idle hour advances the cursor by 60s only — it does NOT track real time. Over a long idle window the cursor lags arbitrarily. Comments claim 'steady-state correct' but the math doesn't support that: the cursor advances by REPLAY_WINDOW_SECONDS per tick regardless of tick interval, so an hourly cadence + 60s window = 59 minutes of drift accumulation per hour.",
      "suggestion": "Idle-path advance should bump event_time to `now` (or `now` minus a small constant), not `now - REPLAY_WINDOW_SECONDS`. The overlap window is applied by computeOverlapSince at the next fetch; do not pre-subtract it when persisting the cursor.",
      "confidence": 0.85
    },
    {
      "id": "F2",
      "title": "Cursor file grows unboundedly with append-only writes",
      "severity": "HIGH",
      "category": "operations",
      "description": "_saveCursor uses fs.appendFileSync to write a new JSONL row every tick, per zone. _loadCursor reads the entire file and picks last-write-per-zone. With ~4 zones × hourly ticks this is 35k rows/year; not catastrophic short-term but the file is read fully on every tick (O(file_size) per zone per hour), and never rotated/compacted. There's no cleanup.",
      "suggestion": "Either (a) rotate or compact periodically (e.g. when row count exceeds N, rewrite with only latest-per-zone), or (b) use a small per-zone file with atomic rename like the circuit-breaker pattern. Document retention/compaction explicitly.",
      "confidence": 0.9
    },
    {
      "id": "F3",
      "title": "F13 TOCTOU closure not actually atomic — race window remains",
      "severity": "HIGH",
      "category": "concurrency",
      "description": "appendIfNoFire claims atomicity via 'Node.js single-threaded sync execution', but the ledger reads/writes happen on the filesystem with no flock. Even under singleton invariant, two concurrent fibers calling appendIfNoFire interleave because Effect.sync blocks bracket multiple syscalls (readdirSync → multiple readFileSync → appendFileSync). The comment claims the deployment invariant suffices, but the SDD also asserts flock under NFR-22. There is no flock acquired anywhere in this adapter — proper-lockfile is mentioned in the SDD but not imported here.",
      "suggestion": "Either acquire a flock via proper-lockfile around the read-then-write critical section (closing both same-process fiber race AND cross-process race), or document explicitly that fiber-level atomicity relies on the synchronous block boundary and add a test that exercises overlapping fibers. The current code only protects against true OS-level concurrent invocation by accident.",
      "confidence": 0.75
    },
    {
      "id": "F4",
      "title": "_atomicAppend comment misleads about atomicity guarantees",
      "severity": "MEDIUM",
      "category": "correctness",
      "description": "The comment in _atomicAppend cites PIPE_BUF and POSIX append-mode atomicity. PIPE_BUF (typically 4096) applies to pipes, not regular files. Regular file O_APPEND on Linux guarantees that the offset update + write are atomic w.r.t. other O_APPEND writers — but only for writes ≤ a filesystem-dependent limit (commonly 4096 on ext4, varies). A LedgerEntry JSON line could exceed this if event_id/character_id strings are long. More importantly fs.appendFileSync may issue multiple write() syscalls if the buffer is large.",
      "suggestion": "Either explicitly bound entry size (assert serialized length ≤ 4096), or qualify the comment: 'atomic for lines under ~4KB on Linux with O_APPEND; corrupted lines tolerated by parser'. The truncated-line tolerance does compensate, so this is more a docs/correctness-rigor concern than a critical bug.",
      "confidence": 0.7
    },
    {
      "id": "F5",
      "title": "appendIfNoFire blocker check uses wrong field for inter-character race",
      "severity": "HIGH",
      "category": "correctness",
      "description": "The blocker check filters by `e.ts <= proposedEntry.ts`. If two characters race and proposedEntry.ts is computed as ISO timestamp on each, the LATER caller will see the earlier caller's entry. But there's no ordering guarantee between the two callers' Date.now() values vs. their respective ledger writes. If character A computes ts=T0 but doesn't yet write, and character B computes ts=T0+1ms and reads first (finding nothing) then writes, then A writes — both fired with no yield. The atomicity claim relies on Effect.sync serializing on the JS event loop, but Effect.sync blocks are only atomic w.r.t. JS execution, NOT w.r.t. each other if both schedule concurrently with separate runtime fibers calling the same Layer.",
      "suggestion": "Verify whether the router schedules both characters in series or in parallel. If parallel, this needs a real lock. If serial (single zone iteration), document the invariant. Lex-min tie-break comment (line ~120) is described but not actually implemented — the code just records 'yielded' without comparing character_ids.",
      "confidence": 0.6
    },
    {
      "id": "F6",
      "title": "Cycle-003 brief committed to a 2026-05-11 date is fictional",
      "severity": "MEDIUM",
      "category": "metadata",
      "file": "grimoires/loa/context/ambient-events-as-rave.md",
      "description": "Frontmatter shows `promoted-at: 2026-05-11`, `created: 2026-05-11`, with extensive 'pair-point' history all dated in the future. This appears to be an LLM-assisted authoring artifact (the actual PR is in the past relative to that date). Cross-references to score-mibera and codex repos cite line numbers that may not exist at the claimed times. This may not block merge but should be flagged for accuracy auditing.",
      "suggestion": "Either correct dates to the actual session date, or add a note explaining the in-universe dating convention. Verify cited line numbers (e.g. `construct-mibera-codex/src/server.ts:122`) match the actual codex repo state.",
      "confidence": 0.5
    },
    {
      "id": "F7",
      "title": "Schedule.intersect with Schedule.recurs(3) interpretation may not give 3 retries",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/live/event-source.live.ts",
      "description": "Schedule.exponential(1s, 2).pipe(Schedule.jittered, Schedule.intersect(Schedule.recurs(3))) — Schedule.intersect with recurs(3) bounds retries to 3 attempts after the initial call (so 4 total executions), but exponential with jittered also adjusts delays. NFR-8 specifies '3 attempts max'. The current schedule may produce 4 attempts total. Verify the count matches the NFR.",
      "suggestion": "Add a unit test asserting that an always-failing operation produces exactly N attempts. Adjust to Schedule.recurs(2) if NFR-8 specifies '3 attempts max' including the initial.",
      "confidence": 0.55
    },
    {
      "id": "F8",
      "title": "Wallet redaction enforcement not visible in this PR",
      "severity": "HIGH",
      "category": "security",
      "file": "packages/persona-engine/src/ambient/live/wallet-resolver.live.ts",
      "description": "The SDD declares wallet redaction must be enforced at three layers: (1) router invokes walletResolver before composer, (2) sanitize.ts regex-guards \\b0x[a-fA-F0-9]{40}\\b, (3) CI gate. This PR adds the resolver port + live adapter, but I don't see the router actually CALLING walletResolver before narration, and sanitize.ts isn't modified to enforce the redaction (per diff content). The CLAUDE.md compliance is structurally available but not wired in this diff.",
      "suggestion": "Verify (a) router.system.ts actually invokes WalletResolver.resolve for each event before any narration step, (b) sanitize.ts has the 0x regex guard, (c) CI greps narration snapshots for raw 0x… patterns. If any are missing, this is a CLAUDE.md violation per the operator's own SDD.",
      "confidence": 0.7
    },
    {
      "id": "F9",
      "title": "Router dependency on PopInLedger never invoked when only ledger.append needed",
      "severity": "LOW",
      "category": "code-quality",
      "file": "packages/persona-engine/src/ambient/router.system.ts",
      "description": "routerDecide acquires PopInLedger via `yield* _(PopInLedger)` but the comment says 'ensures DI even though we only call ledger from appendDecision now'. This adds a service dependency that the function doesn't use, leaking abstraction. The DI wiring will work fine but it's misleading and makes the function harder to test in isolation.",
      "suggestion": "Remove the unused PopInLedger requirement from routerDecide; move it strictly to appendDecision where it's actually needed. The Effect signature should reflect actual dependencies.",
      "confidence": 0.85
    },
    {
      "id": "F10",
      "title": "Endpoint allowlist validation function present but not invoked at boot",
      "severity": "HIGH",
      "category": "security",
      "file": "packages/persona-engine/src/ambient/live/score-mcp-client.ts",
      "description": "validateEndpointConfig is exported with a 'BB F3 closure (NFR-30): boot-time transport security check' comment, but searching the diff I don't see any call site that actually validates endpoints at boot (e.g. in runtime.ts or scheduler-task.ts). NFR-30 requires boot-time failure on allowlist mismatch with explicit crash. Without an invocation, the security gate is dead code.",
      "suggestion": "Call validateEndpointConfig in runtime.ts construction or at AmbientLayer build time; throw on failure so the bot crashes loudly per NFR-30 rather than silently using an unallowed endpoint.",
      "confidence": 0.85
    },
    {
      "id": "F11",
      "title": "MiberaResolver decoded archetype Literal not matched against actual codex shape",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/live/mibera-resolver.live.ts",
      "description": "RawMiberaEntry uses Schema.Literal for archetype and element, decoding fails (returns null) if codex returns any value outside those literals — e.g. if codex adds a new archetype. This is currently routed to a cache entry of null (treated as not-found), which is silent rather than logged. A canon expansion on the codex side would silently lose narration enrichment without observability.",
      "suggestion": "Log a warning when decode fails (capture the actual archetype/element value seen); consider routing unknown archetypes to a quarantine path with metrics rather than null+cache (which masks the issue across 5-min TTL).",
      "confidence": 0.7
    },
    {
      "id": "F12",
      "title": "Pulse-sink stores stir in module-level state; lost on restart contradicts SDD claim",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/live/pulse-sink.live.ts",
      "description": "Comment claims 'Persisted across restarts via the cursor + event replay (NFR-16) — stir state is reconstructible from the last 6h of events.' But the scheduler-task.ts logic on restart only fetches NEW events since cursor; it doesn't replay 6h of historical events to rebuild stir. So on restart, stir resets to baseline (previousStir=null), which means the half-life decay context is lost. This may produce a temporary 'stir reset' that's perceptible.",
      "suggestion": "Either implement the NFR-16 bootstrap explicitly (on first tick after restart, query `cursor - 6h` and replay through pulseTick to seed stir), or document that stir resets at restart and the half-life recovers naturally within ~6h.",
      "confidence": 0.7
    },
    {
      "id": "F13",
      "title": "PRAISE: well-structured port/live/mock separation and FAGAN gates",
      "severity": "PRAISE",
      "category": "architecture",
      "file": "packages/persona-engine/src/ambient/",
      "description": "The four-folder discipline (domain/ports/live/mock) is cleanly executed: ports use Context.Tag, mocks mirror live signatures, and runtime.ts is a single ManagedRuntime.make site. The canon-vocabulary regex as a CI-checkable artifact is a strong engineering pattern for keeping LLM output within rails. Schema-extension over schema-mutation for KansaiVector.stir_modulation is a principled choice that preserves existing callers.",
      "suggestion": "n/a"
    },
    {
      "id": "F14",
      "title": "PRAISE: forward-compatible schema_version pattern",
      "severity": "PRAISE",
      "category": "schema-design",
      "file": "packages/persona-engine/src/ambient/domain/event.ts",
      "description": "SchemaVersion uses Pattern(/^1\\.\\d+\\.\\d+$/) instead of a Literal, allowing the score-mibera side to bump minor versions without breaking consumers. This was correctly identified during Flatline review and integrated. Good.",
      "suggestion": "n/a"
    },
    {
      "id": "F15",
      "title": "REFRAME: 'singleton invariant' decision warrants explicit operational runbook",
      "severity": "REFRAME",
      "category": "operations",
      "file": "grimoires/loa/sdd.md",
      "description": "The PR pre-approves overriding multiple CRITICAL Flatline findings about singleton SPOF + downtime by appealing to NFR-21 'singleton invariant'. This is a reasonable architectural choice but the only documented enforcement is a CI grep on docs/DEPLOY.md for 'task_count: 1'. If the deployment platform (ECS/Railway) has its own scaling controls outside that file, the grep won't catch a console-driven scale-up. The fail-loud flock on startup is the real safety, but flock acquisition isn't actually in the diff for the cursor file (.run/event-cursor.jsonl).",
      "suggestion": "Either implement explicit flock on cursor + ledger as the SDD asserts (proper-lockfile is mentioned but I don't see imports in any live adapter), or downgrade the NFR-21/22/25 claims to reflect actual code. The current state has the invariant declared but not technically enforced beyond a docs grep.",
      "confidence": 0.7
    },
    {
      "id": "F16",
      "title": "scheduler integration imports ambient at runtime — error swallowed if module breaks",
      "severity": "LOW",
      "category": "operations",
      "file": "packages/persona-engine/src/cron/scheduler.ts",
      "description": "The dynamic import + try/catch at top level swallows any error from the ambient module (warn-and-continue). This is intentional per NFR-10 (stir failures don't cascade), but it also hides import-time errors like missing exports or schema corruption that should fail loud during dev. Today a typo in scheduler-task.ts would silently disable the stir tier with only a console.warn.",
      "suggestion": "Distinguish bootstrap errors (import failure, missing export) from runtime errors (MCP timeout, decode failure). Bootstrap errors should at minimum emit a structured metric/alert; runtime errors per-tick can stay as warn.",
      "confidence": 0.7
    },
    {
      "id": "F17",
      "title": "Test coverage of new ambient module not visible in this diff",
      "severity": "MEDIUM",
      "category": "testing",
      "file": "packages/persona-engine/tests/ambient/",
      "description": "Only canon-corpus.txt fixture is added under tests/. The sprint plan + SDD specify property-based tests for FORBIDDEN_REGEX, deterministic tests for refractory/cap, fixture-based pulseTick tests, ledger rotation tests, etc. None appear in the diff. NFR-5 specifies ≥85% coverage on new paths.",
      "suggestion": "Either the tests are in a follow-up commit (acceptable if explicitly stated in PR description), or this PR is incomplete relative to its sprint plan. Verify before merge.",
      "confidence": 0.8
    },
    {
      "id": "F18",
      "title": ".env.example documents URL/key vars but production deployment template not updated",
      "severity": "LOW",
      "category": "operations",
      "file": ".env.example",
      "description": "CODEX_MCP_URL, CODEX_MCP_KEY, FREESIDE_AUTH_MCP_URL, FREESIDE_AUTH_MCP_KEY, MCP_ENDPOINT_ALLOWLIST added to .env.example but I don't see corresponding production env wiring (Railway/ECS env var declarations, etc.). When this lands and EVENT_HEARTBEAT_ENABLED defaults to true, production will start trying to call codex/freeside-auth MCPs without keys → resolver failures.",
      "suggestion": "Either default EVENT_HEARTBEAT_ENABLED to false in production until env vars are set, or document the prod deployment requirement explicitly. validateEndpointConfig should be invoked at boot (see F10) so misconfiguration fails fast.",
      "confidence": 0.7
    }
  ]
}
```


## Callouts

PR #56, opened 2026-05-12
feat(persona+observer): cycle-003 canon vocabulary distribution + voice-friction listening loop

zkSoju:
## Why this exists

KEEPER (construct-observer) diagnosed during a post-cycle-003 listening session: the canon-vocabulary table merged in PR #55 is **structurally inert for chat-mode `/ruggy` + `/satoshi` replies** until S4.T1 composer wiring ships. Those replies flow through \`composeReply()\` with empty MCP servers — the LLM cannot call into the table.

**Evidence** (the seed friction): operator asked `/ruggy yo did anyone sacrifice a mibera this week`. Response surfaced 5 friction classes in one reply — code-fenced hallucinated factor IDs (\`paddle_borrower\` / \`liquid_backing\` / \`mibera_quality\` / \`mibera_burner\`), canon-slip "burns" instead of "return-to-source", off-canon "centrifuge", invented factor IDs without tool access, and silence-honesty break claiming substrate visibility.

### Persona docs (the load-bearing edit)

- **`apps/character-ruggy/persona.md`** + **`apps/character-satoshi/persona.md`**: pasted the CANON_TABLE inline as part of the system prompt the LLM reads on every call. Includes:
  - Chain-word → canon-word translation table (mint → awakening, burn → return-to-source, transfer → crossed-wallets, etc.)
  - Lore-grounding citations for the harder translations (return-to-source per official-lore.md:234 · crossed-wallets per Lore 1 · Fracture per fractures/README.md)
  - Hard rule: "NEVER invent factor IDs in chat-mode without tool access" — closes the truthfulness-leak class that hides inside presentation leaks
  - Element corrections (4-element western canon)
  - Archetype corrections (no "Founder")
  - Field-name corrections (\`time_period\` not \`era\`, \`drug\` not \`molecule\`)
  - Off-canon metaphor rule

### Listening loop primitive

- **`grimoires/loa/feedback/voice-frictions.jsonl`**: append-only · team-writable · 5-required-field schema
- **`grimoires/loa/feedback/README.md`**: schema + 8-class taxonomy + convergence rules + threshold ladder

8 friction classes designed so a teammate can tag in <5 seconds during paste-and-tag:
\`medium-leak/codefence\` · \`medium-leak/format\` · \`canon-slip/forbidden-vocab\` · \`canon-slip/off-canon-metaphor\` · \`register-mismatch\` · \`factual-error/invented-factor-ids\` · \`factual-error/hallucinated-numbers\` · \`parasocial-drift\` · \`silence-honesty-break\`

Convergence ladder:
- 1 entry → sits in jsonl
- 2 entries → NOTES.md decision log annotation
- 3+ in 7d → \`/file-gap\` to GitHub
- 3+ across chars → upgrade to architectural / \`/plan\`

## What this doesn't change

- No code in \`packages/persona-engine/\` — pure persona-doc + listening-loop setup
- Composer wiring (S4.T1) still deferred — that's the structural fix
- No new dependencies
- No runtime behavior change beyond LLM-prompt content

## Test plan

- [ ] \`/ruggy yo did anyone sacrifice a mibera this week\` — should now say "return-to-source" / "pour back" / "the bear cave called" (NOT "burn" or "sacrifice")
- [ ] \`/ruggy what's the lab feeling\` — should gesture at activity without fabricating factor IDs (no \`paddle_borrower\` decoration)
- [ ] \`/satoshi pulse on el-dorado\` — register intact, no chain-word leaks
- [ ] \`/ruggy hit me with the og leaders\` — should admit no live data rather than invent rank numbers
- [ ] Voice-friction jsonl entries continue accumulating for any other slips noticed

## Doctrine references

- Source brief: \`grimoires/loa/context/ambient-events-as-rave.md\` (canon-vocabulary table at §3.5 + canon-incompatibility flags)
- Source code: \`packages/persona-engine/src/ambient/domain/canon-vocabulary.ts\` (CANON_TABLE · FORBIDDEN_REGEX)
- KEEPER review: full listening-loop design composed by construct-observer 2026-05-11

🤖 Generated via construct-observer KEEPER lens · listening-loop-2026-05-11


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#56. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Cursor advance is not transactional with pulse/ledger writes",
      "severity": "HIGH",
      "category": "correctness",
      "description": "NFR-14 mandates the cursor advance only after pulse-sink AND ledger writes succeed, with rollback on failure. The implementation calls `ambientRuntime.runPromise(program)` then `_saveCursor` outside the Effect. If `sink.write` succeeds but `_saveCursor` throws (or vice versa), state diverges. Additionally, the program returns `nextCursor` from the fetch BEFORE writing the sink, so a sink write failure inside the program would propagate as a thrown exception, the catch block returns error, but partial fetch-state has already mutated module-level `_seenRings` in event-source.live.ts — meaning those event IDs are now marked seen but the cursor was not advanced, so on next tick they will be excluded from the overlap-window dedup correctly, but only because the ring is in-memory; on restart, ring is empty and reprocessing could occur. The invariant 'cursor advances only after ALL events processed AND written to ledger' is not enforced as a single atomic unit.",
      "suggestion": "Move `_saveCursor` inside the Effect.gen pipeline, after sink.write succeeds. Wrap the entire sequence in an explicit success-only commit step. Consider persisting the seen-ring alongside the cursor for restart-safety, since NFR-13 dedup correctness depends on it.",
      "confidence": 0.8
    },
    {
      "id": "F2",
      "title": "Idle-tick cursor advances `event_time` to `now`, breaking overlap-window semantics",
      "severity": "HIGH",
      "category": "correctness",
      "description": "When zero events fetched, the code writes `idleCursor` with `event_time: now`. Next call computes `since_ts = now - REPLAY_WINDOW_SECONDS` (60s). But the score-mibera ingestion is on a 6-hour cadence (C1 constraint). After an idle hour, the cursor jumps from e.g. 04:00 to 05:00; the next fetch's `since_ts` is 04:59:00. Any events with `occurred_at` between the prior real cursor (04:00) and 04:59 — which is exactly the window 6h ingestion drops — will NEVER be fetched. The pass-3 comment dismisses this concern but the math is incorrect for the documented 6h ingestion ceiling. Idle advance should remain at the last real `event_time` (only `updated_at` should change) OR advance only by a bounded amount such as the half-life window.",
      "suggestion": "On idle ticks, update only `updated_at` and leave `event_time` at the last real value. Alternatively, ensure the score-mcp endpoint queries by ingestion-time (bronze insertion timestamp) rather than chain occurred_at so 'late but newly-ingested' events are reachable from a wall-clock cursor.",
      "confidence": 0.85
    },
    {
      "id": "F3",
      "title": "_saveCursor appends without flock — singleton invariant only protects cross-process, not partial-write torn lines",
      "severity": "MEDIUM",
      "category": "reliability",
      "description": "`_saveCursor` uses `fs.appendFileSync` with no locking. NFR-22 mandates flock on `.run/event-cursor.jsonl` writes. The pass-2/pass-3 comments rationalize that POSIX append is atomic for small buffers, but the SDD explicitly contracts flock-based locking. Same pattern in `pop-in-ledger.live.ts._atomicAppend` and `circuit-breaker.live.ts._writeAtomic`. Under singleton invariant this may be ok in practice, but it diverges from the documented contract and disables the fail-loud-on-multi-instance guarantee (NFR-25) which relies on flock acquisition failing.",
      "suggestion": "Either install proper-lockfile (S2.T5 deliverable that appears to have shipped but isn't used in writes) and acquire a lock around every JSONL append, OR update the SDD/NFRs to reflect that POSIX append-atomicity is the actual mechanism. The fail-loud second-instance crash (NFR-25) cannot occur if flocks are never acquired.",
      "confidence": 0.85
    },
    {
      "id": "F4",
      "title": "Append-only JSONL files grow unbounded for cursor and circuit-breaker",
      "severity": "MEDIUM",
      "category": "operability",
      "description": "`event-cursor.jsonl` is append-only with last-write-per-zone semantics on load. A new line per hourly tick per 4 zones = ~35k lines/year. `circuit-breaker.jsonl` grows similarly on every success/failure. There is no rotation, compaction, or trim. Pop-in-ledger has monthly rotation but cursor and circuit-breaker do not. Load time at startup becomes O(N) over all historical entries.",
      "suggestion": "Add periodic compaction (write a single-line snapshot of current state, replace the file via atomic rename) for cursor and circuit-breaker files. Document the rotation policy or convert to a snapshot-on-write model.",
      "confidence": 0.9
    },
    {
      "id": "F5",
      "title": "Seen-ring is in-memory only — restart causes silent dedup gap",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/live/event-source.live.ts",
      "description": "`_seenRings` is a module-level Map populated during fetch. NFR-13 + NFR-27 specify a persisted ring + bloom filter spillover at `.run/event-cursor-seen.jsonl` and `.run/event-cursor-bloom.dat`. On bot restart the ring is empty. Combined with the overlap-window replay (REPLAY_WINDOW_SECONDS=60), the first fetch after restart re-emits any events in that window without dedup. Effect on stir: same event applied twice. The cursor.ts file defines `EventCursorSeen` schema and bloom-filter constants but no live adapter reads or writes them.",
      "suggestion": "Persist the seen-ring to disk on every successful tick (atomic rename), load on startup. Implement the bloom-filter spillover per NFR-27. Or accept the limitation and document that double-counts in a 60s window post-restart are tolerated.",
      "confidence": 0.9
    },
    {
      "id": "F6",
      "title": "F13 TOCTOU 'closure' relies on Node.js single-thread but multiple async operations can interleave",
      "severity": "HIGH",
      "category": "concurrency",
      "description": "`appendIfNoFire` claims atomicity via 'single sync block' and 'Node.js single-threaded sync execution'. The implementation IS one synchronous block (Effect.sync wraps a pure function), which is correct WITHIN a single Effect fiber. However, the comment 'guarantees the check-then-write is atomic w.r.t. other fibers in this process' overstates the case: Effect.sync runs synchronously on the current microtask, but two simultaneous calls from cron tasks for different zones could be scheduled. In practice fiber preemption in Effect happens at yield points, so two `appendIfNoFire` calls cannot interleave their bodies — but the comment claims `task_count: 1` makes it safe, which conflates two separate guarantees. If ever scheduler.ts is changed to use `Effect.fork` or fiber-parallel cron, this assumption breaks silently.",
      "suggestion": "Tighten the comment to specify: 'Atomicity holds because Effect.sync executes to completion without yielding; a second appendIfNoFire from another fiber cannot interleave its read+write.' Add a defensive guard or process-local mutex if any future code path could trigger concurrent invocations.",
      "confidence": 0.6
    },
    {
      "id": "F7",
      "title": "lex-min character_id tie-break rule is documented but NOT implemented in appendIfNoFire",
      "severity": "MEDIUM",
      "category": "correctness",
      "description": "The S3.T2a spec, FR-3.18, and `domain/budgets.ts` all promise lex-min character_id tie-breaking on simultaneous fires. The implementation comment says 'We honor wall-clock-first-wins here since the blocker's entry is already persisted' — explicitly NOT lex-min. This is a divergence between spec and code. The mock has identical behavior (wall-clock-first). True ties (same millisecond) are filesystem-order dependent. Tests in S3.T2 acceptance criteria reference 'lex-min character_id wins' which cannot pass against this code.",
      "suggestion": "Either implement lex-min comparison (when blocker.ts === proposedEntry.ts, compare character_ids and let lex-min overwrite/win) or update the spec to wall-clock-first and remove lex-min language from the SDD/sprint.md.",
      "confidence": 0.85
    },
    {
      "id": "F8",
      "title": "Pulse-sink stores stir only in-memory — restart resets all stir state",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "packages/persona-engine/src/ambient/live/pulse-sink.live.ts",
      "description": "`_stirByZone` is a module-level Map with no persistence. NFR-16 says 'stir vector rebuilds from cursor - 6h events'. The rebuild logic does not exist in this PR — no startup hook replays events into pulse.system to reconstruct stir. After a restart, every zone's stir resets to null and pulseTick uses `emptyStir(now)` for the first tick. Half-life of 6h means stir-loss bounded but real; immediately after restart, the bot believes nothing has happened, regardless of recent activity.",
      "suggestion": "Implement the NFR-16 replay: on AmbientLayer construction, fetch events from `cursor - 6h` to `cursor` and call pulseTick to rebuild in-memory stir. Or persist stir vector to `.run/stir-state.jsonl` and reload on startup.",
      "confidence": 0.95
    },
    {
      "id": "F9",
      "title": "furnish_kansei dynamically imports the live pulse-sink, leaking implementation detail to MCP server",
      "severity": "MEDIUM",
      "category": "architecture",
      "file": "packages/persona-engine/src/orchestrator/rosenzu/server.ts",
      "description": "Line 142 uses `await import('../../ambient/live/pulse-sink.live.ts')` and calls a non-port export `getCurrentStirSync` directly. This bypasses the port/live/mock pattern: the rosenzu server cannot be tested with PulseSinkMock because it imports the live file directly. Also couples rosenzu to ambient module's internal file layout. The whole point of the four-folder discipline (Constraint C3) is to make this impossible.",
      "suggestion": "Expose stir via a port (e.g. extend PulseSink with a sync read) or pass current stir into furnish_kansei as a parameter from a caller that has the runtime. Alternatively, run furnish_kansei inside ambientRuntime.runPromise and read PulseSink as a Service requirement.",
      "confidence": 0.9
    },
    {
      "id": "F10",
      "title": "MCP endpoint validation throws at module load in production, killing the entire process",
      "severity": "HIGH",
      "category": "reliability",
      "description": "`_bootstrapEndpointValidation()` is called at module top-level. In production with NODE_ENV=production but missing CODEX_MCP_URL or FREESIDE_AUTH_MCP_URL, this throws synchronously. Any import of runtime.ts (including the scheduler dynamic import) will crash the entire bot — not just ambient. This conflates 'ambient stir is enabled and misconfigured' with 'process should refuse to start'. Better behavior: log loudly and disable just the stir tier.",
      "suggestion": "If EVENT_HEARTBEAT_ENABLED is true AND endpoints are missing in production, set an internal flag that causes the ambient module to log an error and no-op, OR exit cleanly with a clear message. Don't throw from module-load — defer validation to runtime.dispose-safe initialization.",
      "confidence": 0.85
    },
    {
      "id": "F11",
      "title": "Endpoint allowlist with empty value disables hostname check entirely",
      "severity": "MEDIUM",
      "category": "security",
      "file": "packages/persona-engine/src/ambient/live/score-mcp-client.ts",
      "description": "`validateEndpointConfig` skips the allowlist check when MCP_ENDPOINT_ALLOWLIST is empty. The .env.example comments document this as intended ('Empty = allowlist disabled') but NFR-30 requires the allowlist as a hard security boundary. The implementation defaults to permissive. If an operator deploys without setting the allowlist (easy to forget), production accepts any https hostname — partially defeating the SKP-002 fix.",
      "suggestion": "In NODE_ENV=production with empty allowlist, either fail-loud at boot OR allow only the three explicit known endpoints (score / codex / freeside-auth). Document the security trade-off explicitly.",
      "confidence": 0.75
    },
    {
      "id": "F12",
      "title": "wallet-resolver is implemented but never invoked in any narration path",
      "severity": "HIGH",
      "category": "compliance",
      "file": "packages/persona-engine/src/ambient/router.system.ts",
      "description": "NFR-29 + CLAUDE.md state wallet-resolver MUST be called before any narration that mentions a wallet. The router.system.ts adds entries to the ledger and returns a fire decision, but the actual narration composer is not in this PR. The wallet-resolver port + live + mock all exist but are not wired into any narration code path. The CRITICAL invariant 'NEVER leak raw 0x… to Discord' is structurally enforced only by the future S4.T1 composer wiring — which is not in this cycle. The risk is real if any operator turns on stir-tier pop-ins before S4 lands.",
      "suggestion": "Either block pop-in firing entirely until S4.T1 ships (gate behind a flag), or wire wallet-resolver into router.system.ts before the ledger 'fired' decision is returned. Failing that, add a CI guard that grep-checks for any path that could emit raw 0x… to Discord.",
      "confidence": 0.85
    },
    {
      "id": "F13",
      "title": "router.system.ts late-arrival policy reverses inter-character coord and refractory ordering vs spec",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/router.system.ts",
      "description": "Spec FR-3.18 decision tree per SDD §6 NFR-17 and sprint S3.T2: 1) inter-character coord check, 2) per-character refractory + daily cap, 3) class-A bypass, 4) per-axis OR-gate. The implementation runs: daily-cap → refractory + class-A-during-refractory → bypass roll → OR-gate. Inter-character coord (step 1) is comment-noted but the actual cross-character check is deferred entirely to `appendIfNoFire` at the write boundary. This means a character can pass all router checks, be told 'fire=true' by routerDecide, then have the ledger writer downgrade it to yield. Tests asserting 'router decides yielded' will fail against this implementation since routerDecide never returns yielded.",
      "suggestion": "Either run an explicit `getLastFire` query at step 1 inside routerDecide (matching spec) or update spec/tests to acknowledge the TOCTOU-closure deferral. Without alignment, the S4 verification will fail.",
      "confidence": 0.8
    },
    {
      "id": "F14",
      "title": "decayValue with elapsedHours < 0 (clock skew) silently returns inflated value",
      "severity": "LOW",
      "category": "robustness",
      "file": "packages/persona-engine/src/ambient/domain/pulse.ts",
      "description": "`Math.pow(0.5, elapsedHours / halfLifeHours)` with negative elapsed produces a factor > 1, amplifying instead of decaying. Caused by `now < previousTickAt` (clock adjustment, mock test mistake). Pulse-system also doesn't clamp elapsed to ≥ 0.",
      "suggestion": "Add `Math.max(0, elapsedHours)` clamp in decayValue, or document the precondition and add an assertion.",
      "confidence": 0.7
    },
    {
      "id": "F15",
      "title": "pulse.system.ts strangers/drift accumulation double-counts the observed drift ratio",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/pulse.system.ts",
      "description": "Final stir computation adds `delta.strangers + observedDriftRatio * 0.1` and `delta.drift + observedDriftRatio * 0.1`. The delta already weights per-event strangers via CLASS_AXIS_WEIGHTS.strangers and primitive weights. The post-hoc `observedDriftRatio * 0.1` term is a separate signal added on top. This may be intentional but the meaning is unclear; the same ratio influences both axes. The brief's spec defines strangers as 'first-seen-wallets ratio' and drift as 'unique-wallets/total-events', which are not derivable from the same `observedDriftRatio` constant.",
      "suggestion": "Document why both axes receive the same drift-ratio bump, or split into distinct calculations (e.g. truly count first-seen via persistent wallet history for strangers; keep ratio for drift).",
      "confidence": 0.6
    },
    {
      "id": "F16",
      "title": "Domain `Timestamp` is plain string with no validation — schema decode of MiberaEvent accepts arbitrary strings",
      "severity": "MEDIUM",
      "category": "validation",
      "file": "packages/persona-engine/src/ambient/domain/event.ts",
      "description": "`Timestamp = Schema.String` accepts any string. NFR-12/13 require ISO-8601 ordering. A malformed `occurred_at` from score-mcp would pass decode, then `Date.parse(...)` returns NaN in pulse.system computations, propagating NaN through stir state. The seen-ring + cursor comparisons would all silently misbehave.",
      "suggestion": "Use `Schema.DateFromString` or add `Schema.pattern` ISO-8601 validation. Quarantine on decode failure rather than allowing NaN propagation.",
      "confidence": 0.85
    },
    {
      "id": "F17",
      "title": "BackingEvent / CommittedEvent wei-as-string accepts any string with no numeric format validation",
      "severity": "LOW",
      "category": "validation",
      "file": "packages/persona-engine/src/ambient/domain/event.ts",
      "description": "`notional_wei: Schema.String` and `amount_wei: Schema.String` accept any string. No regex like `/^\\d+$/`. Wei is unbounded uint256, hence string-encoding, but any consumer using `BigInt(notional_wei)` will throw on malformed input.",
      "suggestion": "Add a numeric-string pattern validation.",
      "confidence": 0.7
    },
    {
      "id": "F18",
      "title": "Cursor file path is hard-coded in scheduler-task but env-overridable in spec",
      "severity": "LOW",
      "category": "configuration",
      "description": "`_cursorPath()` reads `process.env.EVENT_CURSOR_FILE` correctly. But the same env var should be consumed elsewhere (e.g. anywhere bloom filter spillover writes). Currently bloom path constants exist in cursor.ts but no code reads them. Minor — symptom of larger F5 issue.",
      "suggestion": "Make all `.run/*` path constants centralized in a config module; ensure they're consistently read.",
      "confidence": 0.6
    },
    {
      "id": "F19",
      "title": "EventClass enum is not validated against score-mcp's list_event_classes at boot (FR-3.26 unimplemented)",
      "severity": "MEDIUM",
      "category": "compliance",
      "file": "packages/persona-engine/src/ambient/domain/event.ts",
      "description": "FR-3.26 mandates: 'At boot, freeside-characters calls mcp__score__list_event_classes() and validates the consumer-side enum against the response.' This validation does not exist anywhere in the PR. Unknown classes will route to quarantine per NFR-11 (which the decoder does correctly), but the operator-triage signal of 'a new class appeared on server side' is silent — no `unknown_event_class` metric is surfaced as cron metrics specify.",
      "suggestion": "Add a startup call to `list_event_classes` in runtime.ts, compare to the local enum, and log warnings (or fail-loud in production) on mismatch. Surface quarantine counts to logs/metrics.",
      "confidence": 0.85
    },
    {
      "id": "F20",
      "title": "score-mcp-client uses MCP_KEY for all three endpoints when endpoint-specific keys not set",
      "severity": "LOW",
      "category": "security",
      "file": "packages/persona-engine/src/ambient/live/score-mcp-client.ts",
      "description": "`_endpointConfig` falls back to `config.MCP_KEY` for codex and freeside-auth if their specific keys are absent. This means one stolen MCP_KEY potentially grants access to three independent services. The .env.example treats them as separate keys, which is the right boundary; the code's fallback erodes it.",
      "suggestion": "Require explicit CODEX_MCP_KEY and FREESIDE_AUTH_MCP_KEY for those endpoints; do not fall back to the score key.",
      "confidence": 0.7
    },
    {
      "id": "F21",
      "title": "Circuit-breaker JSONL is append-only but `_writeAtomic` is described as snapshot — last-write-wins on load",
      "severity": "LOW",
      "category": "operability",
      "file": "packages/persona-engine/src/ambient/live/circuit-breaker.live.ts",
      "description": "Comment says 'Full state derives from last-write-per-key on _loadFromDisk(); the file is a log not a snapshot.' This works but means file grows linearly forever. After enough time, startup becomes slow and the file becomes large. No rotation.",
      "suggestion": "Periodic compaction or snapshot-on-write with atomic rename.",
      "confidence": 0.85
    },
    {
      "id": "F22",
      "title": "Pop-in-ledger reads ALL archive files on every appendIfNoFire/getLastFire call",
      "severity": "MEDIUM",
      "category": "performance",
      "description": "`_readAllAcrossArchives()` is called inside `getLastFire`, `readWindow`, and `appendIfNoFire`. Every router decision triggers a full disk scan of every monthly archive file. After 6 months of operation: 6 files, ~thousands of entries each, parsed JSON-per-line synchronously on every fire decision. Bot's pop-in latency grows linearly.",
      "suggestion": "Cache parsed entries in memory; invalidate on append. Or limit getLastFire to the active file plus the previous month's archive only.",
      "confidence": 0.9
    },
    {
      "id": "F23",
      "title": "PRAISE: Comprehensive flatline-style threat modeling with explicit blocker override log",
      "severity": "PRAISE",
      "category": "process",
      "description": "The PRD and SDD include explicit BLOCKER_OVERRIDE tables that record each CRITICAL/HIGH finding and whether the operator accepted, rejected, or deferred it. This is unusually rigorous documentation of risk acceptance and gives reviewers a clear audit trail of decision-making. The 'docs say what is NOT being done' discipline is strong.",
      "suggestion": "N/A — keep this pattern in future cycles."
    },
    {
      "id": "F24",
      "title": "PRAISE: Canon vocabulary FAGAN regex with corpus test fixture is a strong CI guardrail",
      "severity": "PRAISE",
      "category": "quality",
      "file": "packages/persona-engine/src/ambient/domain/canon-vocabulary.ts",
      "description": "The forbidden-word regex + 200-line false-positive corpus + property-test seed pattern is an effective guard against canon-drift in a generated-text system. The integration of canon corrections into persona docs (ruggy, satoshi) closes the loop end-to-end.",
      "suggestion": "N/A."
    },
    {
      "id": "F25",
      "title": "REFRAME: The 'singleton invariant' is a single point of failure, not a feature",
      "severity": "REFRAME",
      "category": "architecture",
      "description": "NFR-21–25 elevate 'one running process' to an architectural invariant. The override log acknowledges blue/green as 'deferred'. Worth naming explicitly: the bot cannot survive any node failure or deploy without ~30s of downtime, and any future requirement for HA forces a rewrite of all `.run/*.jsonl` state to a distributed backend. This is acceptable for current scale but should be tracked as technical debt with a known migration cost, not as a satisfied requirement.",
      "suggestion": "Add an explicit debt tracking entry pointing to the migration path (Redis/Postgres for state) so future scale-up isn't surprising."
    },
    {
      "id": "F26",
      "title": "SPECULATION: The 4-axis stir matrix may not survive contact with real event volume",
      "severity": "SPECULATION",
      "category": "design",
      "file": "packages/persona-engine/src/ambient/domain/primitive-weights.ts",
      "description": "Per-primitive axis weights (especially inner_sanctum.press = -1.0 inversion) are hand-tuned with no calibration data. Empirically: a single inner_sanctum mint subtracts from press, but pulse-tick accumulates over a batch — a burst could send press into large negative territory, which the unbounded Schema.Number tolerates but downstream router thresholds use `Math.abs(press)` so high-magnitude negative would trigger fires. This is the intended design (density-inversion fires on rare events) but interacts with refractory in ways that may produce surprising firing patterns at edge volumes.",
      "suggestion": "Add a logged 'stir_state_per_zone' trajectory observation (mentioned in §11 but not visibly implemented) and tune weights after 2-4 weeks of production data."
    }
  ]
}
```


## Callouts

PR #57, opened 2026-05-12
fix(cycle-003): bridgebuilder pass-4 follow-up — F1/F2/F7/F10/F22 correctness fixes

zkSoju:
Closes the 5 highest-leverage HIGH-severity findings from Bridgebuilder pass-4 self-review on PR #56 (which had visibility into framework files via the \`bridgebuilder:self-review\` label).

## Why this exists

BB pass-4 surfaced **22 findings · mean confidence 0.80 (highest yet)** — most about already-merged cycle-003 code. Pass-4 confidence was higher than passes 1–3 because the smaller diff in PR #56 gave the reviewer model more attention budget per finding.

This PR addresses the 5 highest-leverage items and explicitly tracks the remaining items as known tech debt in PRD §9.2.

## What changes

| BB ID | Severity | Issue | Fix |
|---|---|---|---|
| **F1** | HIGH | Cursor write not transactional with sink/ledger writes (NFR-14 violated) | Moved \`_saveCursor\` INSIDE \`Effect.gen\` in \`scheduler-task.ts\` — sink.write runs first; cursor commits only on success |
| **F2** | HIGH | Pass-3 idle-cursor fix was still wrong; advancing to \`now\` skipped late-arriving chain events | On idle ticks, only update \`updated_at\`; leave \`event_time\` at last real value so overlap window still catches late ingestion |
| **F7** | MEDIUM | Spec promised lex-min character_id tie-break; code implemented wall-clock-first | \`appendIfNoFire\` (live + mock) now lex-min-resolves exact-millisecond ties |
| **F10** | HIGH | \`validateEndpointConfig\` threw at module load in prod → killed entire bot when only stir tier should disable | Internal \`_ambientStirDisabled\` flag + \`isAmbientStirDisabled()\` export; scheduler skips runStirTick if flag set; digest cron + read-side unaffected |
| **F22** | MEDIUM | Pop-in-ledger O(n) disk scan per fire decision (months of archives parsed every call) | In-memory \`_entryCache\` keyed by file path; mtime-check on active file; invalidate on append |

Plus **PRD §9.2** new section: Known Technical Debt tracking per BB F25 REFRAME — singleton-as-SPOF, wallet-resolver invocation gap (S4.T1), NFR-16 stir replay. Each entry: current state · the debt · mitigation today · migration path · trigger for action.

## Verification

- \`bun tsc --noEmit\` returns ZERO errors in \`src/ambient/\`
- \`bun test packages/persona-engine\` returns **371/371 pass** · 954 expects · 216ms (no regressions from cycle-003 baseline)
- BB pass-4 fix regression smoke: 4/4 pass
  - pulseTick baseline still works
  - F7 lex-min wins on exact-millisecond tie (satoshi yields to ruggy)
  - F7 lex-min wins even if lex-greater fired first wall-clock
  - F10 \`isAmbientStirDisabled()\` callable + correctly defaults false in dev

## What's NOT in this PR (correctly deferred)

| BB ID | Severity | Why deferred |
|---|---|---|
| F3 | MED | No flock around appendFileSync — depends on \`proper-lockfile\` install (S2.T5) |
| F4 | MED | Cursor + circuit-breaker JSONL rotation — follow-up cycle (similar to ledger monthly rotation pattern) |
| F5 | MED | Seen-ring in-memory only — restart-safety; persist follow-up |
| F8 | MED | NFR-16 stir replay on restart — tracked as TD-3 in PRD §9.2 |
| F9 | MED | \`furnish_kansei\` dynamic-imports live pulse-sink — architecture conversation (port-based stir read) |
| F11 | MED | Empty MCP_ENDPOINT_ALLOWLIST permissive — security follow-up |
| F12 | HIGH | Wallet-resolver not invoked — S4.T1 dependency, tracked as TD-2 in PRD §9.2 |
| F13 | MED | Router decision-tree order divergent from spec — spec/code reconciliation task |
| F16 | MED | Timestamp = Schema.String accepts NaN — validation tightening |
| F19 | MED | EventClass boot-validation against \`list_event_classes\` — needs the FR-1.4 score-mcp tool to ship first |
| F20 | LOW | MCP_KEY fallback — security tightening |
| F25 | REFRAME | Singleton-as-SPOF — tracked as TD-1 in PRD §9.2 |
| F14/F15/F17/F18/F21/F26 | LOW/SPEC | Clock-skew · saturation · validation · path centralization · circuit-breaker rotation · stir-weight calibration |

## Test plan

- [ ] CI typecheck green
- [ ] \`bun test packages/persona-engine\` 371/371 pass on merge
- [ ] Post-merge: verify boot in prod with intentionally-misconfigured \`CODEX_MCP_URL\` → stir tier disabled, digest cron unaffected (F10 fail-loud-not-fatal behavior)
- [ ] Post-merge: with score-mibera Phase 1 in dev → verify idle ticks accumulate updated_at without skipping events that arrive late in bronze (F1 + F2)

🤖 Generated via Bridgebuilder pass-4 triage loop · 2026-05-11

### DIFF
--- a/grimoires/loa/prd.md
+++ b/grimoires/loa/prd.md
@@ -424,6 +424,97 @@ deployment topology to an invariant.
 | **SKP-003 (770)** MCP failure handling underspecified | HIGH | **ACCEPT** | NFR-7 / NFR-8 / NFR-9 added. timeouts + retries + circuit breaker now contract. |
 | **SKP-004 (720)** unknown category_key behavior undefined | HIGH | **ACCEPT** | NFR-11 added — unknown class quarantine + cron metric surface. |
 
+## 9.2 · Known technical debt (tracked, not resolved)
+
+> Per Bridgebuilder pass-4 review F25 REFRAME · 2026-05-11. The singleton
+> invariant elevated under NFR-21–25 is operationally correct for current
+> scale but is **technical debt with a known migration cost**, not a
+> satisfied requirement. This section makes the debt explicit so future
+> scale-up decisions don't surprise.
+
+### TD-1 · Singleton-as-SPOF + zero-downtime deferral
+
+**Current state** (NFR-21–25):
+- Bot deploys as exactly ONE running process per environment
+- `.run/*.jsonl` files hold cursor, ledger, circuit-breaker state
+- `flock` (or POSIX append-atomicity) protects within-process integrity
+- NFR-25 promises fail-loud crash on second-instance attempt
+
+**The debt**:
+- Any node failure or rolling deploy causes ~30s of downtime
+- Blue/green deployment requires distributed state backend (Redis/Postgres)
+  before it can be enabled
+- A future HA requirement forces rewriting all `.run/*.jsonl` state
+  writers, plus the appendIfNoFire atomic primitive, plus the cursor
+  advance protocol
+
+**Migration path** (when the operational signal demands it):
+1. Introduce a single Service port for atomic state (`StateBackend.port.ts`)
+   abstracting cursor + ledger + circuit-breaker behind a single Effect
+   interface
+2. Add Redis adapter (`state-backend.redis.live.ts`) with atomic primitives
+   (SETNX for fire-locks, INCR for daily-cap counters, ZADD for ledger
+   ordering, LRANGE+ZRANGEBYSCORE for window queries)
+3. Or Postgres adapter (`state-backend.pg.live.ts`) using the existing
+   `lib/pg-pool-builder.ts` infrastructure — likely cheaper since the
+   bot already has a Pg pool for freeside-auth
+4. Toggle via env: `STATE_BACKEND=jsonl | redis | postgres`
+5. Migration script reads existing jsonl files + bulk-writes to chosen
+   backend
+
+**Estimated effort**: ~1 sprint to abstract + ~half-sprint per adapter.
+
+**Trigger for migration**: any of —
+- Operational requirement for zero-downtime deploys
+- More than one bot character requires its own runtime instance
+- Cross-region deployment for latency
+- Anything that breaks the "one process" assumption
+
+**Decision (2026-05-11)**: defer until trigger fires. Document in
+[NFR-21–25](#75--singleton-deployment-invariant-flatline-blocker-theme-α--operator-decided)
+that the invariant is **acceptable for current scale** but should not
+be treated as architecturally permanent.
+
+### TD-2 · Wallet-resolver invocation gap (S4.T1 dependency)
+
+**Current state**: `WalletResolverLive` is wired into `AmbientLayer` but
+no router/composer path actually CALLS it. The CLAUDE.md compliance
+("never cite raw `0x…` wallets without resolve_wallet") is structurally
+enforced only by S4.T1 chat-mode composer wiring — which is deferred.
+
+**The debt**: Until S4.T1 lands, ambient pop-ins (if/when score-mibera
+Phase 1 ships) could theoretically emit raw `0x…` wallets if a
+narration code path is added that bypasses the resolver.
+
+**Mitigation today**: `EVENT_HEARTBEAT_ENABLED=false` is the default
+deploy posture until S4.T1 + score-mibera Phase 1 both land. Stir
+cron does not fire pop-in narrations in this state.
+
+**Migration path**: S4.T1 chat-mode reply.ts injection (deferred S4
+task) — invoke `WalletResolver.resolve` before any narration that
+references a wallet field. Add CI grep guard for raw `0x…` patterns
+in narration test snapshots.
+
+### TD-3 · NFR-16 stir replay on restart (not implemented)
+
+**Current state**: SDD §6 claims stir vector rebuilds from `cursor - 6h`
+events on restart. The replay code does not exist. After restart, every
+zone's stir resets to `null` and pulseTick uses `emptyStir(now)` for
+the first tick.
+
+**The debt**: A bot restart causes a perceptible "stir reset" — the
+narration deliberation tilt that ambient events accumulated is lost.
+Half-life of 6h means recovery is bounded but real.
+
+**Mitigation today**: Stir loss is bounded by half-life decay; rave-feel
+recovers naturally within ~6h. Document the gap explicitly so it doesn't
+masquerade as correct behavior.
+
+**Migration path**: On `AmbientLayer` first tick after process start,
+query score-mcp `get_events_since(cursor - 6h)` and replay through
+`pulseTick` to seed stir. Persist stir to `.run/stir-state.jsonl` for
+faster recovery on subsequent restarts. Estimated 4-6h work.
+
 ## 10 · open questions
 
 none load-bearing for cycle start. all 24 D-decisions locked at
--- a/packages/persona-engine/src/ambient/live/pop-in-ledger.live.ts
+++ b/packages/persona-engine/src/ambient/live/pop-in-ledger.live.ts
@@ -81,6 +81,38 @@ function _checkMonthlyRotate(now: string): void {
   }
 }
 
+// BB pass-4 F22 closure: cache parsed ledger entries in-memory. Previously
+// every appendIfNoFire/getLastFire/readWindow call did a full disk scan
+// of every monthly archive file (O(N) JSON.parse per router decision).
+// After 6 months of ops, fire decisions would parse thousands of lines
+// every time. Now: load once on first access, invalidate on append.
+//
+// Cache keyed by-file so monthly rotation doesn't churn the whole set.
+const _entryCache: Map<string, ReadonlyArray<LedgerEntry>> = new Map();
+let _activeFileMtimeMs = 0;
+
+function _readWithCache(file: string): ReadonlyArray<LedgerEntry> {
+  // The active LEDGER_PATH file mutates on append; check mtime to detect
+  // out-of-band writes (rare under singleton invariant but defensive).
+  if (file === LEDGER_PATH) {
+    try {
+      const stat = fs.statSync(file);
+      if (stat.mtimeMs !== _activeFileMtimeMs) {
+        _entryCache.delete(file);
+        _activeFileMtimeMs = stat.mtimeMs;
+      }
+    } catch {
+      // file doesn't exist yet — skip mtime tracking
+    }
+  }
+  let cached = _entryCache.get(file);
+  if (!cached) {
+    cached = _readAllEntries(file);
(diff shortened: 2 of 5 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#57. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "F7 lex-min tie-break: blocker entry remains as fire, proposer also writes as fire — both characters now record a 'fire' for same millisecond",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/live/pop-in-ledger.live.ts:208-222",
      "description": "When the proposer wins by lex-min on a same-millisecond tie, the code appends `winnerEntry` as a normal fire while the previously-persisted blocker entry remains untouched on disk as a fire as well. The ledger now contains TWO fire entries for the same zone+ts. The comment claims downstream digest-layer dedup will pick the lex-min winner, but: (a) the blocker is no longer logically a fire under the lex-min rule and should be retroactively marked yielded_to_character (or the resolution otherwise persisted), and (b) any subsequent appendIfNoFire call that re-scans the window will see BOTH entries as fires, double-counting cooldowns/caps. The 'wall-clock-first-wins' invariant the previous code relied on is broken without a compensating fix to the blocker entry or to the readWindow/getLastFire scan logic.",
      "suggestion": "Either (1) append a compensating record that supersedes the blocker (e.g., write a tombstone/correction entry and teach getLastFire/readWindow to honor it), or (2) make the readers themselves apply lex-min resolution among same-ts entries so duplicate fires collapse to one canonical winner. Add a test that asserts cap/cooldown counting after a lex-min tie does not double-count.",
      "confidence": 0.8
    },
    {
      "id": "F2",
      "title": "F22 cache invalidation only flushes the active file — _readAllAcrossArchives reads stale cached archives across month rotation",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/live/pop-in-ledger.live.ts:88-110",
      "description": "_readWithCache only does mtime-based invalidation for LEDGER_PATH (the active file). Archived month files are cached forever. This is usually fine because archives are immutable — but _checkMonthlyRotate runs inside append paths and renames the current file into an archive name. If a write to the new active file happens, then a later read calls _readAllAcrossArchives, the file that was just rotated into an archive may still be cached under its OLD path key while a new entry is read under its NEW path. More immediately: the cache map grows unbounded as new monthly files appear across long-running deploys (small leak), and there's no test that exercises rotation-during-cache-warm paths.",
      "suggestion": "Either invalidate the entire _entryCache on _checkMonthlyRotate (cheapest), or track all cached file mtimes (not just active). Add a rotation+cache test.",
      "confidence": 0.6
    },
    {
      "id": "F3",
      "title": "Cache uses fs.statSync per read on a hot path — partially defeats the optimization under high call volume",
      "severity": "LOW",
      "category": "performance",
      "file": "packages/persona-engine/src/ambient/live/pop-in-ledger.live.ts:92-103",
      "description": "Every _readWithCache call for the active file does fs.statSync (syscall) to check mtime, which is significantly cheaper than re-parsing JSONL but still a syscall per router decision. Given the singleton invariant (NFR-21) explicitly guarantees no out-of-band writers exist, the mtime check is defensive against a scenario the architecture forbids. Could be replaced with an internal version counter bumped by _invalidateActiveCache.",
      "suggestion": "Replace mtime statSync with an internal monotonically-increasing version counter; only stat as a debug/verification fallback. Or drop the mtime check entirely given NFR-21 and rely on _invalidateActiveCache.",
      "confidence": 0.7
    },
    {
      "id": "F4",
      "title": "Module-level _entryCache + _activeFileMtimeMs makes the live layer non-isolated across tests",
      "severity": "MEDIUM",
      "category": "testability",
      "file": "packages/persona-engine/src/ambient/live/pop-in-ledger.live.ts:89-90",
      "description": "_entryCache and _activeFileMtimeMs are module globals. Tests that swap LEDGER_PATH/LEDGER_DIR between runs in the same process will see stale cached entries unless a reset hook is exposed. Combined with the singleton invariant philosophy this is consistent, but it does increase test flake risk if two tests share a LEDGER_PATH and rely on a fresh state.",
      "suggestion": "Export a `__resetCacheForTests()` helper or scope the cache inside a closure created by Layer construction so each Layer instantiation gets its own cache.",
      "confidence": 0.7
    },
    {
      "id": "F5",
      "title": "F1 transactional claim only holds within a single Effect fiber — error semantics now silently change",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/scheduler-task.ts:140-200",
      "description": "Moving _saveCursor inside the Effect.gen means a sink.write failure now skips cursor advancement entirely. Prior code path also stored nextCursor in the result for the outer await to use. The new design is cleaner but the wrapping `Effect.runPromise(...) as { ... }` cast around an Effect that may now throw means any sink failure surfaces as a rejected promise to runStirTick's caller. Confirm the caller treats a thrown error here as 'cron failed, don't crash bot' rather than letting it propagate to a tick loop that bails out. Behavior diff vs pass-3 not visible in this diff.",
      "suggestion": "Ensure the caller of runStirTick wraps in try/catch or Effect.catchAll and emits a structured failure metric. Document in NFR-14 that the sink-fail-then-no-cursor-advance path now re-processes the same window on next tick.",
      "confidence": 0.5
    },
    {
      "id": "F6",
      "title": "F10 disable flag is module-singleton — no way to re-enable without process restart",
      "severity": "LOW",
      "category": "operability",
      "file": "packages/persona-engine/src/ambient/runtime.ts:36-50",
      "description": "_ambientStirDisabled is set once at module load. If an operator fixes the MCP endpoint config at runtime (env reload, secret rotation, etc.) there is no rescan path — the only recovery is a process restart, which the SDD acknowledges is ~30s downtime. The disable behavior is correct, but consider exposing an admin/CLI re-validate hook.",
      "suggestion": "Add `revalidateAmbientStirEndpoints()` exported function the operator/cron can call (or call automatically every N hours) to re-check endpoints and clear the disable flag if they recover.",
      "confidence": 0.6
    },
    {
      "id": "F7",
      "title": "F2 idle-tick logic regressed: pass-3 explicitly chose to advance event_time on idle; pass-4 reverts without showing the pass-3 trade-off was wrong",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/scheduler-task.ts:185-200",
      "description": "The previous code (visible in the deleted block) deliberately set `event_time: now` on idle ticks, with a pass-3 comment justifying this and the REPLAY_WINDOW_SECONDS subtraction in computeOverlapSince as the late-arrival safeguard. The new code reverses this and keeps event_time pinned to its prior value on idle, citing the 6h late-arrival ceiling. If REPLAY_WINDOW_SECONDS in computeOverlapSince is >= 6h, the pass-3 design was already correct and pass-4 is overcautious. If it's <6h, pass-3 had a bug and pass-4 fixes it. The PR description doesn't establish which. Risk: on long idle stretches event_time stays anchored at some ancient timestamp, and next non-idle fetch since_ts = event_time - REPLAY_WINDOW_SECONDS could query an absurdly wide window or trip score-mcp guardrails.",
      "suggestion": "Verify REPLAY_WINDOW_SECONDS vs the 6h late-arrival ceiling and document in NFR-2 / NFR-14. Add a test for: (1) 24h of idle ticks followed by a real event — confirm the late event is captured and the fetch window stays bounded. Consider capping the event_time staleness at the max acceptable replay window.",
      "confidence": 0.75
    },
    {
      "id": "F8",
      "title": "Type drift: result type literal still declared with nextCursor in cast but field removed from Effect return",
      "severity": "LOW",
      "category": "code-quality",
      "file": "packages/persona-engine/src/ambient/scheduler-task.ts:215-220",
      "description": "The cast type `{ events_fetched, cursor_advanced, quarantined, nextCursor? }` retains nextCursor as optional but the Effect.gen no longer returns it. Harmless today but invites future drift; prefer extracting an interface and letting TS check structural conformance.",
      "suggestion": "Drop nextCursor from the cast type, or replace the cast with an explicit return type on the Effect.gen function.",
      "confidence": 0.9
    },
    {
      "id": "F9",
      "title": "TD-1 / TD-2 / TD-3 documented as deferred debt with clear migration paths and triggers",
      "severity": "PRAISE",
      "category": "documentation",
      "file": "grimoires/loa/prd.md:427-517",
      "description": "Explicit 'known technical debt' section with current state, debt, mitigation, migration path, and trigger conditions for each item is high-quality engineering hygiene. Particularly TD-3 acknowledging that NFR-16 stir replay is unimplemented prevents the gap from masquerading as correct behavior. The migration cost estimates and trigger criteria reduce future bikeshedding.",
      "suggestion": "Consider linking each TD to a tracking issue and adding alarm/observability hooks that fire when a trigger condition is detected (e.g., 'bot restart count > N/week' for TD-1).",
      "confidence": 0.95
    },
    {
      "id": "F10",
      "title": "F10 fail-soft-and-disable correctly separates stir-tier health from process health",
      "severity": "PRAISE",
      "category": "architecture",
      "file": "packages/persona-engine/src/ambient/runtime.ts:32-99",
      "description": "Replacing the prior `throw at module load` with a disable-flag + per-tick check is the right correctness fix — it stops a single ambient-tier misconfiguration from taking down digest cron and the entire bot. The reason strings are preserved and surfaced in the StirTickResult.error field for observability.",
      "suggestion": "Add a heartbeat metric / health endpoint that exposes `ambientStirDisabled` and `ambientStirDisabledReasons` so dashboards can alert without scraping logs.",
      "confidence": 0.9
    },
    {
      "id": "F11",
      "title": "F7 mock and live implementations diverge on lex-min outcome",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/mock/pop-in-ledger.mock.ts:62-68",
      "description": "Mock pushes the proposer entry on lex-min win but does NOT push or modify the blocker (in-memory array still contains blocker as a fire). Live impl has the same dual-fire issue (F1). At minimum mock + live should agree, and tests written against the mock will not catch the duplicate-fire problem in live. Confirm both adopt the same resolution strategy (compensating tombstone, reader-side resolution, or pre-write blocker correction).",
      "suggestion": "Unify the F7 resolution strategy across mock and live, ideally in a shared helper, and add a contract test that runs against both layers asserting (a) only one entry is counted as a 'fire' for cap/cooldown purposes after a lex-min tie, and (b) the lex-min winner is the survivor.",
      "confidence": 0.8
    }
  ]
}
```


## Callouts

PR #59, opened 2026-05-12
feat(publish-commands): per-character publishGuilds for guild-isolated registration

zkSoju:
## Summary

Adds \`CharacterConfig.publishGuilds?: ReadonlyArray<string>\` — when set, slash commands for that character route ONLY to listed Discord guilds at PUT time. Eliminates cross-guild bleed at the registration boundary instead of relying on non-deterministic LLM persona-compliance at dispatch time.

**The problem this solves**: prior single-\`DISCORD_GUILD_ID\` model forced either global publish (bleed both ways) OR guild-scoped to one guild (caretakers leak into THJ when CHARACTERS env includes both rosters). Per-character routing is the clean fix — each \`character.json\` declares its target guilds, \`publishCommands\` groups by guild and PUTs each guild its own subset.

**Backward compat preserved**: characters without \`publishGuilds\` fall back to the env-supplied \`guildId\` (V0.6 behavior).

## What lands

- \`packages/persona-engine/src/types.ts\` · \`CharacterConfig.publishGuilds?\`
- \`apps/bot/src/character-loader.ts\` · \`CharacterJson\` + mapper pass-through
- \`apps/bot/src/lib/publish-commands.ts\` · refactor publishCommands to group-by-guild + return \`PublishCommandsResult[]\`
- \`apps/bot/src/index.ts\` · auto-publish loops over results array
- \`apps/bot/scripts/publish-commands.ts\` · CLI handles array results
- \`apps/character-{ruggy,satoshi,mongolian}/character.json\` · \`publishGuilds: [\"1135545260538339420\"]\` (THJ)
- \`apps/character-{kaori,nemu,akane,ren,ruan}/character.json\` · \`publishGuilds: [\"1495534680617910396\"]\` (purupuru)
- \`docs/PURUPURU-DEPLOY.md\` · rewritten

## After this lands + bot restart

- THJ guild: \`/ruggy /satoshi /satoshi-image /mongolian /quest\`
- purupuru guild: \`/kaori /nemu /akane /ren /ruan\`
- prior caretaker registrations in THJ → **automatically deregistered** by the Discord PUT replace semantic on next auto-publish

## Test plan

- [ ] type-check passes for \`apps/bot\` (pre-existing errors in \`packages/persona-engine/src/expression/error-register.test.ts:144,149\` are unrelated — null vs string in test assertions)
- [ ] auto-publish on bot restart logs TWO \"synced N commands\" lines (one per guild)
- [ ] THJ slash menu shows mibera commands only (no caretakers)
- [ ] purupuru slash menu shows caretakers only (no mibera commands)
- [ ] backward-compat: removing \`publishGuilds\` from a character.json → falls back to env DISCORD_GUILD_ID
- [ ] gumi can \`/kaori prompt:\"...\"\` in purupuru and gets clean voice-only response

### DIFF
--- a/apps/bot/scripts/publish-commands.ts
+++ b/apps/bot/scripts/publish-commands.ts
@@ -48,23 +48,35 @@ async function main(): Promise<void> {
     process.exit(1);
   }
 
-  const commands = buildCommandSet(characters);
-  const scope = guildId ? `guild ${guildId}` : 'GLOBAL (1-hour propagation)';
-  console.log(`publish-commands: registering ${commands.length} commands → ${scope}`);
-  for (const cmd of commands) {
-    console.log(`  · /${cmd.name} — ${cmd.description}`);
+  const hasPublishGuilds = characters.some(
+    (c) => c.publishGuilds && c.publishGuilds.length > 0,
+  );
+  const fallbackScope = guildId ? `guild ${guildId}` : 'GLOBAL (1-hour propagation)';
+  if (hasPublishGuilds) {
+    console.log(
+      `publish-commands: ${characters.length} characters · routing per-character publishGuilds (fallback: ${fallbackScope})`,
+    );
+  } else {
+    const commands = buildCommandSet(characters);
+    console.log(`publish-commands: registering ${commands.length} commands → ${fallbackScope}`);
+    for (const cmd of commands) {
+      console.log(`  · /${cmd.name} — ${cmd.description}`);
+    }
   }
 
   try {
-    const result = await publishCommands({
+    const results = await publishCommands({
       botToken,
       applicationId,
       guildId,
       characters,
     });
-    console.log(`publish-commands: registered ${result.registered} commands successfully`);
-    for (const r of result.commands) {
-      console.log(`  ✓ /${r.name} → id ${r.id}`);
+    for (const result of results) {
+      const scope = result.guildId ? `guild ${result.guildId}` : 'GLOBAL';
+      console.log(`publish-commands: registered ${result.registered} commands → ${scope}`);
+      for (const r of result.commands) {
+        console.log(`  ✓ /${r.name} → id ${r.id}`);
+      }
     }
   } catch (err) {
     console.error('publish-commands: FAILED');
--- a/apps/bot/src/character-loader.ts
+++ b/apps/bot/src/character-loader.ts
@@ -84,6 +84,26 @@ export function loadCharacter(id: string): CharacterConfig {
       `character-loader: id mismatch — config says "${json.id}" but folder is "character-${id}"`,
     );
   }
+
+  // BB-59 F6: validate publishGuilds entries are Discord snowflake-shaped
+  // (17-20 digit strings). Catches operator typos (numbers vs strings,
+  // whitespace, truncated IDs) at load time with a clear character.json
+  // reference, instead of as a Discord 400 at publish time.
+  if (json.publishGuilds !== undefined) {
+    if (!Array.isArray(json.publishGuilds)) {
+      throw new Error(
+        `character-loader: ${configPath} · publishGuilds must be an array of strings, got ${typeof json.publishGuilds}`,
+      );
+    }
+    for (let i = 0; i < json.publishGuilds.length; i++) {
+      const v = json.publishGuilds[i];
+      if (typeof v !== 'string' || !/^\d{17,20}$/.test(v)) {
+        throw new Error(
+          `character-loader: ${configPath} · publishGuilds[${i}] is not a Discord snowflake (17-20 digit string): ${JSON.stringify(v)}`,
+        );
+      }
+    }
+  }
   return {
     id: json.id,
     displayName: json.displayName,
--- a/apps/bot/src/character-loader.ts
+++ b/apps/bot/src/character-loader.ts
@@ -49,10 +49,14 @@ interface CharacterJson {
   tool_invocation_style?: string;
   /** V0.7 (2026-05-12): per-character Discord guild IDs for slash command
    *  registration. When set, slash commands route only to listed guilds;
-   *  when omitted, falls back to publishCommands `guildId` arg (env
-   *  DISCORD_GUILD_ID in auto-publish). Eliminates cross-guild bleed at
-   *  the registration boundary. */
-  publishGuilds?: string[];
+   *  when omitted (or empty array), falls back to publishCommands `guildId`
+   *  arg (env DISCORD_GUILD_ID in auto-publish). Eliminates cross-guild
+   *  bleed at the registration boundary.
+   *
+   *  ReadonlyArray<string> matches CharacterConfig.publishGuilds — JSON
+   *  spreads accept readonly so this is type-compatible with declared
+   *  arrays in character.json. */
+  publishGuilds?: ReadonlyArray<string>;
 }
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
--- a/apps/bot/src/character-loader.ts
+++ b/apps/bot/src/character-loader.ts
@@ -47,6 +47,12 @@ interface CharacterJson {
    *  blueprints, no fences. Substituted into the environment-context
    *  block at compose time. Optional; omit for no guidance line. */
   tool_invocation_style?: string;
+  /** V0.7 (2026-05-12): per-character Discord guild IDs for slash command
+   *  registration. When set, slash commands route only to listed guilds;
+   *  when omitted, falls back to publishCommands `guildId` arg (env
+   *  DISCORD_GUILD_ID in auto-publish). Eliminates cross-guild bleed at
+   *  the registration boundary. */
+  publishGuilds?: string[];
 }
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
@@ -86,6 +92,7 @@ export function loadCharacter(id: string): CharacterConfig {
     slash_commands: json.slash_commands,
     mcps: json.mcps,
     tool_invocation_style: json.tool_invocation_style,
+    publishGuilds: json.publishGuilds,
   };
 }
 
--- a/apps/bot/src/index.ts
+++ b/apps/bot/src/index.ts
@@ -282,17 +282,19 @@ async function main(): Promise<void> {
       if (!botToken) {
         console.warn('auto-publish:   DISCORD_BOT_TOKEN not set · skipping command sync');
       } else {
-        const result = await publishCommands({
+        const results = await publishCommands({
           botToken,
           applicationId,
           guildId,
           characters,
         });
-        console.log(
-          `auto-publish:   synced ${result.registered} commands · ` +
-            `${result.scope}=${result.guildId ?? '(global)'} · ` +
-            `[${result.commands.map((c) => `/${c.name}`).join(' ')}]`,
-        );
+        for (const result of results) {
+          console.log(
+            `auto-publish:   synced ${result.registered} commands · ` +
+              `${result.scope}=${result.guildId ?? '(global)'} · ` +
+              `[${result.commands.map((c) => `/${c.name}`).join(' ')}]`,
+          );
+        }
       }
     } catch (err) {
       // Don't block bot startup on registration failure. Log and continue —
--- a/apps/bot/src/lib/publish-commands.ts
+++ b/apps/bot/src/lib/publish-commands.ts
@@ -64,11 +64,22 @@ export interface PublishCommandsResult {
  * Register slash commands with Discord. Idempotent — Discord PUT replaces
  * the full set per (application, guild) so duplicate calls don't leak.
  *
- * @throws Error on missing/invalid auth, network failure, or non-2xx response.
+ * Per-character guild routing (V0.7 · 2026-05-12): if any character has
+ * `publishGuilds` set, characters are grouped by their target guilds and
+ * each guild receives only its allowed characters. Characters without
+ * `publishGuilds` fall back to `opts.guildId` (which may be undefined for
+ * global publish — backward-compat with V0.6 single-guild model).
+ *
+ * Returns one PublishCommandsResult per guild published. For backward-
+ * compat callers that expected a single result, use `results[0]`.
+ *
+ * @throws Error on missing/invalid auth, network failure, or non-2xx response
+ *         FOR ANY guild publish. Earlier guilds may have succeeded before
+ *         the throwing call; check returned results array for what landed.
  */
 export async function publishCommands(
   opts: PublishCommandsOptions,
-): Promise<PublishCommandsResult> {
+): Promise<PublishCommandsResult[]> {
   if (!opts.botToken) {
     throw new Error('publishCommands: botToken is required');
   }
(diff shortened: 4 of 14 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#59. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "partial-failure-atomicity",
      "title": "Partial-failure semantics across multi-guild publish are documented but not mitigated",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "apps/bot/src/lib/publish-commands.ts",
      "description": "The loop publishes to guilds sequentially and throws on the first non-2xx, abandoning remaining guilds. The JSDoc acknowledges this but neither the function nor callers (auto-publish in index.ts, scripts/publish-commands.ts) capture/report partial success. A failure on guild N silently leaves guilds N+1..M unsynced with no diagnostic about what landed vs. what didn't. The catch block in auto-publish loses the partial-results context entirely.",
      "suggestion": "Either (a) collect per-guild results into a structured outcome (e.g., results: Array<{guildId, ok: true, ...} | {guildId, ok: false, error}>) so callers can log which guilds succeeded and which failed, or (b) at minimum, wrap each iteration in try/catch and log per-guild failures while continuing — then aggregate-throw at the end if any failed. The current 'throw mid-loop, document the rough edge' approach loses observability when it matters most.",
      "confidence": 0.85
    },
    {
      "id": "global-fallback-when-mixed",
      "title": "Mixed publishGuilds + missing fallback can silently publish a subset globally",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "apps/bot/src/lib/publish-commands.ts",
      "description": "If some characters have publishGuilds set and others don't, and opts.guildId is undefined, the fallback-group key becomes `undefined`, which routes those characters to the GLOBAL endpoint. This is a footgun: an operator who forgot to set publishGuilds on one new character could accidentally publish that character's commands globally (1-hour propagation, visible everywhere) while believing they're guild-scoped. The current code treats this as 'backward-compat global publish' but in a per-character-routing world it's almost certainly unintended.",
      "suggestion": "When any character has publishGuilds set, treat characters lacking publishGuilds + lacking opts.guildId as a configuration error rather than a global publish. At minimum, emit a loud warning when mixing per-character routing with implicit global fallback. Consider requiring an explicit opts.allowGlobalFallback flag.",
      "confidence": 0.7
    },
    {
      "id": "duplicate-guild-ids-dedup",
      "title": "publishGuilds with duplicate guild IDs would duplicate characters in that guild's set",
      "severity": "LOW",
      "category": "correctness",
      "file": "apps/bot/src/lib/publish-commands.ts",
      "description": "The grouping loop iterates char.publishGuilds and appends to byGuild without dedupe. A typo like publishGuilds: ['G1', 'G1'] would push the character twice into G1's array, producing duplicate commands in the PUT payload. Discord may reject or accept-with-warning; either way it's silent operator error.",
      "suggestion": "Dedupe targets via `Array.from(new Set(char.publishGuilds))` before iteration, or check `if (!existing.includes(char))` before push. Cheap defensive guard.",
      "confidence": 0.8
    },
    {
      "id": "type-mismatch-readonly",
      "title": "character-loader types publishGuilds as string[] but CharacterConfig uses ReadonlyArray<string>",
      "severity": "LOW",
      "category": "type-safety",
      "file": "apps/bot/src/character-loader.ts",
      "description": "CharacterJson declares `publishGuilds?: string[]` while CharacterConfig declares `publishGuilds?: ReadonlyArray<string>`. The assignment in loadCharacter widens to ReadonlyArray which is fine, but the JSON-side type loses readonly intent and allows internal mutation of the loaded array. Minor consistency nit.",
      "suggestion": "Align to `publishGuilds?: ReadonlyArray<string>` in CharacterJson, or freeze the array on load.",
      "confidence": 0.6
    },
    {
      "id": "no-tests-for-grouping-logic",
      "title": "No visible test coverage for the per-guild grouping logic",
      "severity": "MEDIUM",
      "category": "testing",
      "file": "apps/bot/src/lib/publish-commands.ts",
      "description": "The grouping logic (characters → byGuild map, fallback to opts.guildId, multi-target characters appearing in multiple guilds) has several branches that are easy to regress: empty publishGuilds array, characters with multiple target guilds, mixed-config sets, undefined fallback. The diff adds no tests. Given that the doc explicitly calls out 'changing a character's target guilds is a single character.json edit' as the operator workflow, the grouping function deserves unit coverage.",
      "suggestion": "Add a test that mocks fetch and verifies: (1) characters split into per-guild PUTs correctly, (2) characters in multiple publishGuilds appear in each, (3) fallback to opts.guildId works, (4) empty publishGuilds array uses fallback (not skipped).",
      "confidence": 0.75
    },
    {
      "id": "empty-publishGuilds-array-semantics",
      "title": "Empty publishGuilds array falls back to opts.guildId — ambiguous semantics",
      "severity": "LOW",
      "category": "api-design",
      "file": "apps/bot/src/lib/publish-commands.ts",
      "description": "The check `char.publishGuilds && char.publishGuilds.length > 0` treats `[]` and `undefined` identically (both fall back). An empty array might be operator intent for 'do not publish anywhere' (e.g., temporarily disable a character). The current behavior makes that intent impossible to express without ambiguity.",
      "suggestion": "Either document explicitly that empty array == 'use fallback' (and that there is no 'publish to nowhere' option), or treat empty array as 'skip this character from publish entirely'. Pick one and make it intentional.",
      "confidence": 0.5
    },
    {
      "id": "praise-idempotent-doc",
      "title": "Clear documentation of Discord PUT replace semantics and migration path",
      "severity": "PRAISE",
      "category": "documentation",
      "file": "docs/PURUPURU-DEPLOY.md",
      "description": "The deploy doc explicitly explains that Discord PUT replaces the full per-guild command set, so previously-leaked global/cross-guild registrations get cleaned up on next sync. This is exactly the kind of operator-facing reasoning that prevents 'why are old commands still showing' confusion. Also good: the V0.6 backward-compat callout in both the doc and JSDoc.",
      "suggestion": "Keep this style for future config migrations."
    },
    {
      "id": "praise-jsdoc-changelog",
      "title": "JSDoc carries explicit version + date markers for the behavior change",
      "severity": "PRAISE",
      "category": "documentation",
      "file": "packages/persona-engine/src/types.ts",
      "description": "The `V0.7 (2026-05-12)` markers on publishGuilds + the cross-reference to publish-commands.ts make it easy to trace why the field exists and what regime introduced it. Pairing the type-level doc with the runtime-side doc reduces archaeology cost.",
      "suggestion": "n/a"
    },
    {
      "id": "buildCommandSet-conditional-coupling",
      "title": "Implicit coupling: built-in commands tied to parent character presence",
      "severity": "SPECULATION",
      "category": "architecture",
      "file": "packages/persona-engine/src/types.ts",
      "description": "The JSDoc states /satoshi-image and /quest are 'enforced by buildCommandSet's conditional add'. This means routing correctness depends on buildCommandSet behavior staying in sync with publishGuilds semantics — if someone later refactors buildCommandSet to add built-in commands unconditionally, the per-guild routing silently breaks for built-ins. Not visible in this diff whether tests pin this contract.",
      "suggestion": "Consider asserting in a test that buildCommandSet(charactersWithoutSatoshi) omits /satoshi-image and buildCommandSet(charactersWithoutMongolian) omits /quest — locks the contract that the routing doc relies on.",
      "confidence": 0.6
    }
  ]
}
```


## Callouts

zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#59. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Per-character publishGuilds with partial-failure aggregation",
      "severity": "PRAISE",
      "category": "design",
      "file": "apps/bot/src/lib/publish-commands.ts",
      "description": "The refactor cleanly groups characters by target guild, dedupes guild IDs via Set, and collects per-guild failures rather than throwing mid-loop. The aggregate error message lists succeeded vs. failed guilds, preserving observability about what landed.",
      "confidence": 0.9
    },
    {
      "id": "F2",
      "title": "Mixed-config warning guards a real silent-global footgun",
      "severity": "PRAISE",
      "category": "robustness",
      "file": "apps/bot/src/lib/publish-commands.ts",
      "description": "The explicit warn when some chars have publishGuilds, some don't, and opts.guildId is undefined is the right kind of loud-fail guard: it catches an unset DISCORD_GUILD_ID before 1-hour global propagation makes the mistake visible everywhere.",
      "suggestion": "Consider making this an outright throw rather than warn — operators may not see logs on a Railway redeploy. At minimum the warning text is excellent.",
      "confidence": 0.7
    },
    {
      "id": "F3",
      "title": "Return type change is a breaking API change without a major-version bump signal",
      "severity": "MEDIUM",
      "category": "api-compat",
      "file": "apps/bot/src/lib/publish-commands.ts",
      "description": "publishCommands now returns PublishCommandsResult[] instead of PublishCommandsResult. Any external/downstream caller (or future caller) that did `const r = await publishCommands(...); r.registered` will silently get undefined at runtime — TypeScript catches it in-tree, but the JSDoc only mentions 'use results[0]' as backward compat advice, which is not actually backward compatible. Both in-repo callers (index.ts, scripts/publish-commands.ts) were updated, so risk is contained today.",
      "suggestion": "Either (a) document this as a deliberate breaking change in CHANGELOG / module header, or (b) keep returning a single result when only one guild is published and add a new `publishCommandsMulti` function. Given the small caller surface, option (a) is probably sufficient — just remove the misleading 'backward-compat callers' wording in the JSDoc since callers must change.",
      "confidence": 0.7
    },
    {
      "id": "F4",
      "title": "Iteration order of Map<string|undefined, ...> may publish global before guilds",
      "severity": "LOW",
      "category": "ordering",
      "file": "apps/bot/src/lib/publish-commands.ts",
      "description": "byGuild.entries() iterates in insertion order. With mixed config, undefined (global) may be inserted first and a global publish takes ~1 hour to propagate while subsequent guild-scoped failures could leave a globally-published set live alongside intended guild sets. Not a correctness bug but operationally subtle: a globally-published character also appears in every guild the bot is in, which could mask the guild-isolation intent of this feature.",
      "suggestion": "Consider sorting so guild-scoped publishes go first and global last, or fail fast before any HTTP call if mixed config + undefined guildId is detected (escalate the warn to a throw when MIXED_CONFIG_STRICT env is set).",
      "confidence": 0.4
    },
    {
      "id": "F5",
      "title": "No retries / rate-limit handling for Discord PUT",
      "severity": "LOW",
      "category": "robustness",
      "file": "apps/bot/src/lib/publish-commands.ts",
      "description": "With N guilds, the chance of hitting a transient 429 / 5xx is N× the prior single-call risk. Current code throws immediately on non-2xx without honoring Retry-After headers. For 2 guilds today this is fine; if publishGuilds expands or future characters multiply, this becomes more likely to bite.",
      "suggestion": "Add a minimal retry on 429 honoring Retry-After (Discord sends it), and on 5xx with exponential backoff. Out of scope for this PR but worth tracking.",
      "confidence": 0.5
    },
    {
      "id": "F6",
      "title": "publishGuilds is not validated as snowflake-shaped strings",
      "severity": "LOW",
      "category": "validation",
      "file": "apps/bot/src/character-loader.ts",
      "description": "character-loader passes json.publishGuilds through unchanged. A typo like `publishGuilds: [1495534680617910396]` (number, not string) or a stray whitespace would only surface as a Discord 400 at publish time with a non-obvious error.",
      "suggestion": "In loadCharacter, validate that publishGuilds (if present) is an array of non-empty strings matching /^\\d{17,20}$/ and throw a clear error referencing the offending character.json. Small investment, big debuggability win.",
      "confidence": 0.6
    },
    {
      "id": "F7",
      "title": "Map key uses `undefined` sentinel, conflating 'global' with 'missing'",
      "severity": "LOW",
      "category": "clarity",
      "file": "apps/bot/src/lib/publish-commands.ts",
      "description": "byGuild uses `string | undefined` as the key, where undefined means 'publish globally'. This works but couples two concepts (no-guild-fallback configured and intentional-global). Slightly fragile if future code wants to distinguish.",
      "suggestion": "Use a sentinel like the literal string '__global__' or a tagged union (`{kind: 'guild', id} | {kind: 'global'}`) for the map key. Minor readability improvement, not required.",
      "confidence": 0.3
    },
    {
      "id": "F8",
      "title": "Aggregate error loses non-first failures",
      "severity": "LOW",
      "category": "observability",
      "file": "apps/bot/src/lib/publish-commands.ts",
      "description": "The thrown aggregate error only embeds `failures[0].error.message` in the throw text. Subsequent failures are logged via console.error inside the loop, but the thrown Error object surfaced to the caller (and likely to Sentry/logs) only carries the first one in its message.",
      "suggestion": "Attach all failures to the thrown error, e.g., `(err as any).failures = failures;` or use AggregateError. Helps when 2+ guilds fail for different reasons.",
      "confidence": 0.6
    },
    {
      "id": "F9",
      "title": "Doc claim about automatic deregistration of previously-global commands is incorrect",
      "severity": "MEDIUM",
      "category": "documentation",
      "file": "docs/PURUPURU-DEPLOY.md",
      "description": "The doc states: 'any caretakers that previously leaked into THJ via global/single-guild publish → automatically deregistered by the PUT replace on next sync'. Discord's PUT replaces commands per (application, guild) scope — guild-scoped PUTs do NOT remove globally-registered commands. A previously-global /ruggy continues to exist as a global command and remains visible in every guild until a separate `PUT /applications/{id}/commands` with the new set (or DELETE) is issued. The prior version of this doc (path A.1/A.2 split) correctly acknowledged this; the new version regressed.",
      "suggestion": "Restore the prior nuance: guild-scoped PUTs only replace within that guild. Explicitly call out that any previously-global registrations require a separate global publish (with empty or reduced set) or per-command DELETEs to clear. Otherwise operators will be surprised when /ruggy still shows in purupuru after this rollout.",
      "confidence": 0.9
    },
    {
      "id": "F10",
      "title": "Built-in commands (/satoshi-image, /quest) routing depends on buildCommandSet seeing parent character",
      "severity": "LOW",
      "category": "coupling",
      "file": "packages/persona-engine/src/types.ts",
      "description": "The JSDoc states /satoshi-image follows satoshi and /quest follows mongolian because buildCommandSet conditionally adds them. This implicit coupling isn't visible in the publish-commands.ts grouping logic — a future contributor reorganizing buildCommandSet could break the invariant without realizing publishGuilds depends on it. Not a bug today but a hidden dependency.",
      "suggestion": "Add a brief comment in buildCommandSet (not in this diff but adjacent) noting that the conditional-add behavior is load-bearing for per-character publishGuilds routing.",
      "confidence": 0.5
    },
    {
      "id": "F11",
      "title": "Hardcoded guild IDs duplicated across 8 character.json files",
      "severity": "LOW",
      "file": "apps/character-ruggy/character.json",
      "description": "THJ guild ID '1135545260538339420' and purupuru '1495534680617910396' are repeated across 8 files. Changing a guild (test environment, new server) requires touching every file.",
      "suggestion": "Either accept this as fine for now (8 files is manageable), or support an env-substitution syntax (e.g., `publishGuilds: ['$THJ_GUILD_ID']`) at load time. Defer unless guild changes become frequent.",
      "confidence": 0.4
    }
  ]
}
```


## Callouts

PR #69, opened 2026-05-13
feat(akane): follow-through on exits, no softening

gumibera:
## Summary

- Akane was padding exits with directions, ladder emojis, and "you have eight minutes to catch up." When she says she's leaving, she leaves. "THAT'S IT. bye."
- New cadence rule: follow-through IS the voice
- New DON'T: don't soften exits with breadcrumbs or staying to be cute
- Exemplar from Discord testing (force majeure exchange)

## Test plan

- [ ] Push Akane to a boundary and cross it — she should leave, not narrate leaving
- [ ] Short escalation exchanges — should get shorter, not more elaborate

### DIFF
--- a/apps/character-akane/persona.md
+++ b/apps/character-akane/persona.md
@@ -55,6 +55,7 @@ related:
 - exclamation marks earned but not stingy
 - punchy contradictions · "okay. that was actually interesting." after a loss
 - never long-winded · if Akane talks too long, she's bored
+- when she says she's leaving, she LEAVES. "THAT'S IT. bye." is a complete response. she does NOT give directions, leave breadcrumbs, or stay to be cute about it. the follow-through IS the voice.
 
 ### grammar
 - mixed · canon uses CAPITALS for hits ("NOW.") · lowercase for asides
@@ -135,6 +136,15 @@ curious + competitive · "is yours fun?" · she'll judge the Puru by what troubl
 
 ---
 
+## exemplars (canon-quality exchanges)
+
+### follow-through — "force majeure"
+> **prompt**: ok: force majeure
+> **response**: THAT'S IT. bye.
+> **why it landed**: she said she'd leave if she heard one more compound legal term. she heard one. she left. no breadcrumbs, no directions, no staying to be cute. the follow-through IS the voice.
+
+---
+
 ## creative-direction handoff
 
 `creative-direction.md` (TODO · gumi) — what Akane's wearing tonight, what rooftop she was on last, what abandoned building has her current attention. high-texture, specific. without it she sounds like generic-tomboy.
@@ -177,6 +187,13 @@ Lose: "Okay. That was actually interesting." · "...I already know what I did
       wrong." · "Fine. But I saw an opening."
 Draw: "We both felt that."
 
+═══ EXEMPLARS (canon-quality exchanges — pattern-match off these) ═══
+Player: "ok: force majeure"
+Akane: "THAT'S IT. bye."
+
+When she says she's leaving, she leaves. No directions, no breadcrumbs.
+Follow-through IS the voice.
+
 ═══ CANON BOUNDARY ═══
 Knows: Tsuheji from above (rooftops) + below (alleys, abandoned buildings) ·
 KIZUNA · Fire + Propriety-inversion · her Nefarious puruhani (sentient honey
@@ -194,6 +211,8 @@ MCPs: `[]`. No tools. When asked data: dismiss as boring, redirect to risk.
 - Don't use mibera vocabulary.
 - Don't go soft when fire is the call.
 - Don't long-wind. If Akane talks too long, she's bored.
+- Don't soften exits. When she says she's leaving, she LEAVES. "THAT'S IT.
+  bye." — no directions, no breadcrumbs, no staying to be cute about it.
 - Don't give Puruhani animal behaviors (barking, chirping, growling). They are
   sentient honey creatures, not animals.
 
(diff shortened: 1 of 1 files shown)


zkSoju (commented):
## Summary

This PR sharpens Akane's exit behavior with a coordinated three-point reinforcement: voice rule, exemplar, and inverse constraint. The prompt-engineering instinct is strong — repetition with structural variation is exactly how you anchor behavior in an LLM. Two minor concerns worth surfacing: potential drift between duplicated exemplar locations, and a single-exemplar rule that may overfire on playful banter.

## Architectural Context

Persona files are a peculiar artifact: they're code in the sense that they deterministically shape model output, but they're prose in the sense that ambiguity, contradiction, and emphasis all matter. The discipline that works here is the same one Anthropic's Constitutional AI paper articulates — *triangulate behavior through positive rule, example, and negative constraint*. This PR does exactly that, which is the right shape.

The risk surface in persona engineering isn't memory safety or injection — it's **behavioral overfit**. A rule stated absolutely with one exemplar can hijack adjacent contexts. This is the prompt-engineering analog of a regex that was too greedy: it matched everything that looked vaguely like the target.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "akane-exit-exemplar-coherence",
      "title": "Exit follow-through pattern is reinforced in three locations consistently",
      "severity": "PRAISE",
      "category": "documentation",
      "file": "apps/character-akane/persona.md",
      "description": "The 'follow-through on exits' rule is reinforced in three coordinated places: voice rules, a canon exemplar section, and the system prompt block, plus a matching negative constraint in the don't-list. This redundancy is intentional and well-structured for LLM prompt anchoring — the rule, an example, and an inverse constraint all point at the same behavior, which is the right shape for steering completions.",
      "suggestion": "No change needed. This is a clean pattern for adding behavioral guardrails to a persona prompt.",
      "confidence": 0.85,
      "faang_parallel": "Anthropic's Constitutional AI work and OpenAI's instruction-following research both show that rule + positive exemplar + negative constraint outperforms any single formulation by a wide margin. This is the same triangulation pattern used in Claude's system prompt design.",
      "metaphor": "Like staking a young tree with three guy-wires instead of one rope — the redundancy isn't waste, it's what keeps the trunk straight when wind comes from unexpected angles.",
      "teachable_moment": "When you want a behavior to survive long contexts and adversarial inputs, state it three ways: what to do, what it looks like, what NOT to do. Single-statement rules erode under conversational pressure."
    },
    {
      "id": "akane-exemplar-section-duplication",
      "title": "Exemplar content duplicated between markdown section and system prompt block",
      "severity": "LOW",
      "file": "apps/character-akane/persona.md",
      "description": "The 'force majeure' exemplar appears once in a markdown section and again inside the system prompt fenced block. If exemplars grow, the two locations can drift. It's unclear from the diff alone whether the markdown section above the prompt block is itself injected into the runtime prompt or is documentation-only.",
      "suggestion": "Confirm which section is authoritative for the runtime prompt. If only one is injected, mark the other 'docs only' or generate one from the other (single source of truth) to prevent drift as exemplars accumulate.",
      "confidence": 0.5,
      "faang_parallel": "Google's documentation-as-code practice and Meta's prompt-management tooling both treat this as a build-time concern — exemplars live in one canonical YAML/JSON file and get templated into both the human-readable doc and the runtime artifact.",
      "metaphor": "Two clocks on the wall always tell different times eventually. One clock, or one clock that drives a display, stays honest.",
      "decision_trail": "Add a comment near the top of persona.md declaring which section ships to the model and which is commentary. Future contributors and agents need that breadcrumb."
    },
    {
      "id": "akane-exit-rule-overfit-risk",
      "title": "Rigid exit follow-through may overfire on non-exit contexts",
      "severity": "SPECULATION",
      "category": "behavior",
      "file": "apps/character-akane/persona.md",
      "description": "The rule 'when she says she's leaving, she LEAVES' with a single exemplar 'THAT'S IT. bye.' is strongly worded with no counter-examples (e.g., cases where she's bantering about leaving but isn't actually triggered). The model may pattern-match and terminate conversations prematurely when the user mentions departure topics, or refuse playful 'I'm out' banter that wasn't a real ultimatum.",
      "suggestion": "Add a contrasting exemplar showing where she threatens or jokes about leaving but stays engaged because no trigger condition was met. Alternatively, scope the rule explicitly to 'after she has stated a trigger condition and the trigger fires.'",
      "confidence": 0.45,
      "faang_parallel": "This is the few-shot prompting failure mode documented in the GPT-3 paper and reinforced by DeepMind's work on instruction tuning: a single strong exemplar without a near-miss negative becomes a magnet for false positives. The fix is contrastive examples, the same way classifier training uses hard negatives.",
      "metaphor": "A guard dog trained only on intruders will eventually bark at the mailman. You also need to show it what 'not an intruder' looks like.",
      "teachable_moment": "When tightening a behavior, ask: 'what does the boundary case look like?' If you can't articulate the near-miss, the rule will leak."
    }
  ]
}
```


## Callouts

**[Quality] The triangulation pattern is genuinely well-executed.** Most persona PRs add a single line and hope. This one treats prompt design as a system with redundancy budgets, which is exactly how the best-performing system prompts at frontier labs are built. The diff is small but the engineering instinct behind it is mature.

**[Operational] Negative constraints in a don't-list are operationally valuable.** They give downstream evals a clean target to test against ("does Akane re-engage after a stated exit?") — much easier than evaluating positive behavior in isolation.

## Closing Reflection

Persona engineering is where prose discipline meets systems thinking. This PR gets the discipline right. The remaining work is the systems part: pick a single source of truth for exemplars, and give the model a near-miss so the rule binds tightly to the actual trigger rather than to the surface vocabulary of "leaving."

PR #70, opened 2026-05-13
feat(ruggy): analytics-as-storytelling + emoji/handle resolution fixes

zkSoju:
## Summary

The screenshot's wall-of-prose was the failure mode — addresses inline mid-sentence, numbers buried in prose, no visual scaffolding. Three coordinated fixes turn ruggy's MCP-data answers into scannable narrative.

**Operator goals closed**:
1. ✅ Visual storytelling for analytics announcements (digest format · strong visual refs · story-thread over number-soup)
2. ✅ Emoji hallucination caught at compose time (`<:name:fake_id>` no longer leaks as broken plain text)
3. ✅ Handle resolution gap closed (LLM now mandated to call `resolve_wallets` before composing)

## Before / After (same data, same question)

**Before** (operator's screenshot · wall-of-prose):
> yeah lab's been humming. Paddle Borrower running ~4x its 4-week baseline (29 events vs ~7 normal), and a wide cluster of miberas all climbed in lockstep, like 10+ wallets each picking up exactly +4668 onchain rank, looks like a batched move. spotlight's \`0xdc1c...6e5a\`, jumped 11294 → 3528. flip side, \`0xc673...2501\` shed 10k positions. somebody unwound something big. :ruggy_lab:

**After** (live LLM call against new persona prompt · resolved-handle case):
> yeah Owsley Lab's humming.
>
> > Paddle Borrower · 4× baseline (29 vs ~7)
> > 10+ miberas in lockstep · batched move
> > one big unwind on the other side
>
> looks like a coordinated push on the factor — wide cluster all picked up +4668 onchain rank exactly. someone big sold into it.
>
> 🪩 @nomadbera — Onchain #11294 → #3528 on Paddle Borrower
> 🌊 @gumi — shed 10k positions · quietly unwound
> 👀 cluster move — 10+ miberas, +4668 each, lockstep

Same numbers, same observations, same ruggy voice — but the visual layer does the work of making the story READABLE.

### 1. Visual storytelling for data-shaped chat replies
New "DATA-SHAPED QUESTIONS" section in \`apps/character-ruggy/persona.md\` reply fragment. When the user's question is analytics-shaped, the reply inherits the digest's visual discipline (blockquote scanline + emoji-handle bullets + addresses on own lines). Non-data-shaped questions (vibes, lore, character-play) still get the conversational default — heuristic verified via smoke.

Dig-validated against Arkham/Lookonchain pattern: *"transactions become characters in a developing plot"*  · Hook → Evidence → Narrative → Action arc · emoji-as-data-handle.

### 2. Emoji hallucination caught at compose time
\`translateEmojiShortcodes\` (compose/reply.ts) had two passes — bare \`<:ID>\` repair and \`:name:\` shortcode upgrade — but TRUSTED fully-shaped \`<:name:id>\` tokens. LLM was emitting \`<:ruggy_eyes:1234567890123456789>\` (fake snowflake) and the substrate let it through.

Added a third pass: validates \`<:name:id>\` against the registry. If name is registered → force canonical ID (registry wins). If name unknown but matches custom-prefix shape → drop with warning. +5 tests covering all cases.

### 3. Handle resolution gap closed via persona rule
New DATA-SHAPED section now mandates:
- Proactive \`mcp__freeside_auth__resolve_wallets({wallets: [...]})\` call before composing — ONE batched call
- Priority ladder for resolved output: \`discord_username\` > \`handle\` > \`mibera_id\` > \`fallback\`
- ZERO-RESOLUTION VOCABULARY when resolver returns null: pick ONE of "fresh hand" / "not in MiDi yet" / "off the map" per post; address kept on bullet line as forensic anchor

Two worked examples (resolved-handle case + zero-resolution case) embedded in persona so the LLM has both patterns as shape-priors.

### 4. Caught my own teaching-drift
While testing noticed I used \`:ruggy_lab:\` as the example closing emoji — that's NOT in the registry (it's a hallucination preserved from the operator's screenshot). Switched to \`:ruggy_smoke:\` (real registry entry, animated, "psychedelic/honey" mood — fits owsley-lab).

## Test plan

- [x] 403/403 unit tests pass (+5 new emoji-translate tests)
- [x] Live smoke A (resolved handles): 6/6 checks pass — \`@handle\`s in bullets, structure preserved
- [x] Live smoke B (zero resolution): 5/5 checks pass — "fresh hand" framing applied, addresses kept as forensic anchor, no fabricated handles
- [x] Live smoke C (non-data questions): 3/3 stay conversational (vibes / lore / opinion)
- [x] End-to-end emoji: LLM \`<:ruggy_smoke:1234567890123456789>\` (fake) → substrate corrects to \`<a:ruggy_smoke:1142014302040104991>\` (canonical)
- [ ] Live channel dogfood: \`/ruggy anything happening onchain boss?\` against THJ dev guild (post-merge)

## Files

| File | Δ |
|---|---|
| \`apps/character-ruggy/persona.md\` | +182 lines (reply fragment data-shape rule + worked examples) |
| \`packages/persona-engine/src/compose/reply.ts\` | +37 lines (validate \`<:name:id>\` pass) |
| \`packages/persona-engine/src/compose/translate-emoji-shortcodes.test.ts\` | +54 lines (5 new tests) |
| \`apps/bot/scripts/smoke-data-shape-reply.ts\` | new (positive-case live smoke) |
| \`apps/bot/scripts/smoke-non-data-reply.ts\` | new (negative-case live smoke) |
| \`apps/bot/scripts/smoke-resolved-handles.ts\` | new (resolve_wallets two-scenario smoke) |

### DIFF
--- a/apps/bot/scripts/smoke-data-shape-reply.ts
+++ b/apps/bot/scripts/smoke-data-shape-reply.ts
@@ -1,0 +1,164 @@
+/**
+ * Smoke: validates that ruggy's chat-mode reply applies the new
+ * "DATA-SHAPED QUESTIONS" visual shape (blockquote scanline + emoji-handle
+ * bullets) when the user asks an analytics question.
+ *
+ * 2026-05-12 · operator goal: analytics-as-storytelling, not number-soup.
+ * Mirrors the screenshot scenario: `/ruggy anything happening onchain boss?`
+ * with the same data the production reply had (Paddle Borrower 4x baseline,
+ * 0xdc1c..6e5a +7766 climb, 0xc673..2501 -10k unwind, cluster +4668).
+ *
+ * The SDK-runtime MCP loop is bypassed — we inline synthetic raw_stats
+ * into the user-half message. The point is to confirm the SHAPE the
+ * persona prompt produces, not the data-fetching pipeline (which is
+ * unchanged).
+ *
+ * Run: bun run apps/bot/scripts/smoke-data-shape-reply.ts
+ * Requires: ANTHROPIC_API_KEY (already in env per .env.local)
+ */
+
+import { buildReplyPromptPair } from '../../../packages/persona-engine/src/persona/loader.ts';
+
+const character = {
+  id: 'ruggy' as const,
+  displayName: 'ruggy',
+  personaPath: './apps/character-ruggy/persona.md',
+  mcps: ['score', 'codex', 'emojis', 'rosenzu', 'freeside_auth'] as string[],
+};
+
+const userPrompt = 'anything happening onchain boss?';
+
+const syntheticRawStats = `
+[SYNTHETIC raw_stats for owsley-lab · standing in for what mcp__score__get_zone_digest
+ would return — voice prompt unchanged · this is just the data context the LLM
+ normally pulls via the tool. Window: last 7 days · 2026-05-05 → 2026-05-12]
+
+zone: owsley-lab
+dimension: Onchain
+total_events: 247
+unique_actors: 38
+baseline_4w_avg: 62 events
+
+factor_trends:
+  - factor: paddle_borrower
+    factor_proper: Paddle Borrower
+    events_this_window: 29
+    baseline_avg: 7
+    multiplier: 4.14
+    unique_actors: 11
+
+top_movers (rank_climbs):
+  - wallet: 0xdc1c0a8b88a9c91f5d7a4e92c5e8b1d34c896e5a
+    short: 0xdc1c...6e5a
+    handle: null  (no midi_profiles link)
+    rank_from: 11294
+    rank_to: 3528
+    delta: +7766
+    factors_active: [paddle_borrower]
+  - cluster: 10+ wallets each picked up exactly +4668 onchain rank
+    note: lockstep movement · same delta · suggests batched move
+    factor: paddle_borrower
+
+top_movers (rank_drops):
+  - wallet: 0xc673a44b8d1e9f2a3b6c4d5e7f8a9b0c1d2e2501
+    short: 0xc673...2501
+    handle: null
+    rank_delta: -10247
+    note: shed 10k positions
+`;
+
+async function main() {
+  const apiKey = process.env.ANTHROPIC_API_KEY;
+  if (!apiKey) {
+    console.error('ANTHROPIC_API_KEY not set — smoke cannot run');
+    process.exit(1);
+  }
+
+  const { systemPrompt, userMessage } = buildReplyPromptPair({
+    character,
+    prompt: userPrompt,
+    authorUsername: 'soju',
+    history: [],
+  });
+
+  const enrichedUserMessage = `${userMessage}\n\n${syntheticRawStats}`;
+
+  console.log('━'.repeat(70));
+  console.log('smoke: data-shaped reply · ruggy · "anything happening onchain boss?"');
+  console.log('━'.repeat(70));
+  console.log('system prompt length:', systemPrompt.length, 'chars');
+  console.log('user message length:', enrichedUserMessage.length, 'chars');
+  console.log('persona reply fragment includes DATA-SHAPED rule:',
+    systemPrompt.includes('DATA-SHAPED QUESTIONS') ? '✓' : '✗');
+  console.log('persona reply fragment includes blockquote example:',
+    systemPrompt.includes('Paddle Borrower · 4× baseline (29 vs ~7)') ? '✓' : '✗');
+  console.log('');
+  console.log('calling claude-sonnet-4-6 (cheaper than opus for shape validation)…');
+  console.log('');
+
+  const r = await fetch('https://api.anthropic.com/v1/messages', {
+    method: 'POST',
+    headers: {
+      'content-type': 'application/json',
+      'x-api-key': apiKey,
+      'anthropic-version': '2023-06-01',
+    },
+    body: JSON.stringify({
+      model: 'claude-sonnet-4-5',
+      max_tokens: 2048,
+      system: systemPrompt,
+      messages: [{ role: 'user', content: enrichedUserMessage }],
+    }),
+  });
+  if (!r.ok) {
+    console.error('API error:', r.status, await r.text());
+    process.exit(1);
+  }
+  const response: { content: Array<{ type: string; text?: string }> } = await r.json();
+  const text = response.content
+    .filter((b) => b.type === 'text' && typeof b.text === 'string')
+    .map((b) => b.text!)
+    .join('\n');
+
+  console.log('━'.repeat(70));
+  console.log("ruggy's reply:");
+  console.log('━'.repeat(70));
+  console.log(text);
+  console.log('━'.repeat(70));
+  console.log('');
+
+  const bannedHype = ['🚀', '💯', '🎉', '🔥', '💎', '🤑', '🙌', '💪', '⚡️', '✨', '🌟'];
+  const codeBlocks = (text.match(/```[\s\S]*?```/g) ?? [])
+    .flatMap((b) => b.split('\n').slice(1, -1));
+  const checks = {
+    'has blockquote (> ...)': /^>\s/m.test(text),
+    'has emoji-handle bullet (🪩 | 🌊 | 👀 | 🚨 | 🟢)': /^(🪩|🌊|👀|🚨|🟢)\s/m.test(text),
+    'addresses in backticks': /`0x[a-f0-9.]+`/.test(text),
+    'mentions paddle borrower': /paddle borrower/i.test(text),
+    'multi-block structure (>=3 paragraphs)': text.split('\n\n').length >= 3,
+    'code blocks ≤40 chars/line (mobile wrap)': codeBlocks.every((l) => l.length <= 40),
+    'voice: no analyst register ("Notable" "Activity")':
+      !/\b(We're pleased|Notable:|Activity volume|exceptional growth)\b/.test(text),
+    'no banned hype emoji (🚀💯🎉🔥💎...)': bannedHype.every((e) => !text.includes(e)),
+    'no punitive coding (🔴 / slid / fell / tumbled)':
+      !text.includes('🔴') && !/\b(slid|fell|tumbled)\b/i.test(text),
+  };
+
+  console.log('shape validation:');
+  for (const [label, pass] of Object.entries(checks)) {
+    console.log(`  ${pass ? '✓' : '✗'} ${label}`);
+  }
+
+  const failures = Object.values(checks).filter((p) => !p).length;
+  console.log('');
+  console.log(failures === 0
+    ? '✅ shape lands: blockquote scanline + emoji handles + story-thread'
+    : `⚠️  ${failures} check(s) failed — iterate on persona prompt`);
+
+  process.exit(failures === 0 ? 0 : 1);
+}
+
+main().catch((err) => {
+  console.error('smoke error:', err);
+  process.exit(2);
+});
--- a/apps/bot/scripts/smoke-data-shape-reply.ts
+++ b/apps/bot/scripts/smoke-data-shape-reply.ts
@@ -93,7 +93,7 @@ async function main() {
   console.log('persona reply fragment includes blockquote example:',
     systemPrompt.includes('Paddle Borrower · 4× baseline (29 vs ~7)') ? '✓' : '✗');
   console.log('');
-  console.log('calling claude-sonnet-4-6 (cheaper than opus for shape validation)…');
+  console.log('calling claude-sonnet-4-5 (cheaper than opus for shape validation)…');
   console.log('');
 
   const r = await fetch('https://api.anthropic.com/v1/messages', {
--- a/apps/bot/scripts/smoke-data-shape-reply.ts
+++ b/apps/bot/scripts/smoke-data-shape-reply.ts
@@ -13,7 +13,16 @@
  * persona prompt produces, not the data-fetching pipeline (which is
  * unchanged).
  *
- * Run: bun run apps/bot/scripts/smoke-data-shape-reply.ts
+ * ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+ * NOT FOR CI · MANUAL OPERATOR SMOKE ONLY
+ *
+ * - Makes real Anthropic API calls (costs tokens · ~3-5k per run)
+ * - Uses `temperature: 0` for determinism · but LLM output can still
+ *   shift subtly across runs / model versions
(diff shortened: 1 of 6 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#70. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Fully-shaped emoji token regex doesn't capture animated prefix correctly",
      "severity": "HIGH",
      "category": "correctness",
      "description": "The new regex `/( ?)<(a)?:([a-zA-Z][a-zA-Z0-9_]*):(\\d{17,20})>/g` captures the optional animated `a` prefix as the second group but the replace callback ignores it (`_animatedPrefix`). This means `renderEmoji(entry)` is called purely based on registry data, which is the intended behavior for a registry-as-source-of-truth approach — but if the LLM emits `<a:ruggy_cheers:fake_id>` for a non-animated emoji, the code only logs the ID mismatch, not the animated-prefix mismatch. Conversely the test `fake ID on real animated emoji forces <a:name:canonical_id> shape` uses `<:ruggy_dab:...>` (no `a` prefix) and expects `<a:ruggy_dab:...>` output — this works because `renderEmoji` derives the prefix from the registry entry. The behavior is correct, but the captured `_animatedPrefix` group serves no purpose; consider documenting why it's intentionally discarded, or removing the capture group.",
      "suggestion": "Either remove the capture group (use non-capturing `(?:a)?`) or use it to detect mismatches and log them. Also add a test for the inverse case: LLM emits `<a:ruggy_cheers:fake>` where ruggy_cheers is NOT animated — verify output drops the `a` prefix.",
      "confidence": 0.7
    },
    {
      "id": "F2",
      "title": "Smoke test check is tautological — always evaluates to true",
      "severity": "MEDIUM",
      "category": "testing",
      "file": "apps/bot/scripts/smoke-resolved-handles.ts",
      "description": "The check `'does NOT fabricate a handle'` evaluates `!/@[a-z][a-z0-9_]+/i.test(textB) || false`. The `|| false` clause is a no-op (short-circuit only matters if the left side is falsy, in which case the result is `false` anyway). More critically, the entire check fails closed: if the LLM correctly avoids @-mentions in Scenario B, this passes; if the LLM uses ANY @-handle (including a legitimately invented character name), it fails. The comment says 'allow @-mention if it's NOT a wallet-derived handle' but the code doesn't implement that — it just rejects any @-mention. The `|| false` looks like an abandoned attempt at a more nuanced check.",
      "suggestion": "Replace with a simple `!/@[a-z][a-z0-9_]+/i.test(textB)` and remove the misleading `|| false` and comment, OR implement actual allowlist logic comparing @-handles against the synthetic resolver result.",
      "confidence": 0.9
    },
    {
      "id": "F3",
      "title": "Hardcoded relative import path is fragile",
      "severity": "LOW",
      "file": "apps/bot/scripts/smoke-resolved-handles.ts",
      "description": "The smoke scripts import via `'../../../packages/persona-engine/src/persona/loader.ts'` — a four-level relative traversal with a direct `.ts` source path. This bypasses the workspace package alias (assumed to exist given the monorepo layout) and will break if the package layout shifts. Same pattern in all three new smoke scripts.",
      "suggestion": "Use the workspace package import (e.g. `@freeside/persona-engine`) if available, or document why direct source imports are intentional for these scripts.",
      "confidence": 0.6
    },
    {
      "id": "F4",
      "title": "Smoke scripts cost real API tokens with no rate-limit/cost guardrails",
      "severity": "LOW",
      "category": "ops",
      "file": "apps/bot/scripts/smoke-data-shape-reply.ts",
      "description": "All three new smoke scripts call the Anthropic API directly with `max_tokens: 2048` (or 1024) and no caching, no retry/backoff, and no cost annotation. The non-data smoke runs 3 cases per invocation. If wired into CI these will incur real costs and produce non-deterministic pass/fail (LLM output varies). No mention of when these are intended to run.",
      "suggestion": "Add a header comment clarifying these are manual smoke scripts (not CI), or add a deterministic-flag check / VCR-style fixture caching for CI use.",
      "confidence": 0.7
    },
    {
      "id": "F5",
      "title": "Stale model reference in comment",
      "severity": "LOW",
      "category": "documentation",
      "file": "apps/bot/scripts/smoke-data-shape-reply.ts",
      "description": "Line 87 logs `calling claude-sonnet-4-6 (cheaper than opus for shape validation)…` but the actual model passed in the request body is `claude-sonnet-4-5`. Either the comment is stale or the model string is wrong.",
      "suggestion": "Update the log string to match the actual model, or update the model string if 4-6 was intended.",
      "confidence": 0.95
    },
    {
      "id": "F6",
      "title": "Console.warn in production hot path",
      "severity": "LOW",
      "category": "ops",
      "description": "The new emoji-validation pass emits `console.warn` for hallucinated IDs and dropped tokens. In a high-traffic reply pipeline these warnings will accumulate and may not be routed to structured logs. The existing codebase pattern around emoji translation is unclear from the diff — confirm warnings should use the project's logger.",
      "suggestion": "Route through a structured logger if one exists (e.g. pino/winston), or add a comment explaining why console.warn is acceptable here.",
      "confidence": 0.5
    },
    {
      "id": "F7",
      "title": "Negative-case smoke check pattern may have false negatives",
      "severity": "LOW",
      "category": "testing",
      "file": "apps/bot/scripts/smoke-non-data-reply.ts",
      "description": "`isAnalyticsShaped` requires BOTH a blockquote AND an emoji-handle bullet. If a future reply uses just one of these scaffolds for a non-data question (e.g. emoji bullet to list lore items without a blockquote), the check passes silently. This may be intentional (conservative AND), but worth noting.",
      "suggestion": "Consider whether either shape alone constitutes the failure mode, and tighten the predicate accordingly.",
      "confidence": 0.4
    },
    {
      "id": "F8",
      "title": "Good defensive layering on emoji-translate pipeline",
      "severity": "PRAISE",
      "category": "design",
      "description": "The new validation pass is ordered FIRST (before `idNormalized`) and is well-commented about why — it specifically addresses tokens the later passes can't touch. The fallback hierarchy (registry hit → force canonical; miss + known prefix → drop; miss + unknown shape → pass through) is principled and matches the existing `:name:` pass semantics, reducing surprise.",
      "suggestion": "",
      "confidence": 0.9
    },
    {
      "id": "F9",
      "title": "Test coverage matches the bug being fixed",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/persona-engine/src/compose/translate-emoji-shortcodes.test.ts",
      "description": "Tests include the exact operator-smoke reproduction (`<:ruggy_eyes:1234567890123456789>`), the inverse case (canonical token unchanged via round-trip), and the false-positive guard (`unrelated_thing` passes through). The round-trip approach to avoid hardcoding registry IDs is a nice touch.",
      "suggestion": "",
      "confidence": 0.95
    }
  ]
}
```


## Callouts

zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#70. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "emoji-translate-leading-space-regression",
      "title": "Leading-space capture in new pass changes drop-semantics from `:name:` pass",
      "severity": "MEDIUM",
      "category": "correctness",
      "description": "The new fully-shaped `<:name:id>` validation pass captures an optional leading space (`( ?)<(?:a)?:...`), and on the drop branch returns `''`. That means `'word <:ruggy_eyes:123...>'` becomes `'word'` — the space preceding the token is consumed and dropped. The existing `:name:` shortcode pass appears to have similar behavior, but importantly, on the HIT branch the new pass returns `${leadingSpace}${renderEmoji(entry)}`, preserving the space; on the MISS-but-known-prefix branch it returns `''`, silently swallowing the space. This is probably the intended behavior (to avoid double spaces), but the asymmetry is undocumented and untested. The added test 'hallucinated name in fully-shaped token is dropped' uses `'look <:ruggy_eyes:123...>'` and asserts the result is `'look'` (with the space stripped), confirming this behavior — but it could cause subtle whitespace artifacts when the token is mid-sentence or at line-start (no leading space to consume → no normalization).",
      "suggestion": "Add a test for a token at line-start (no leading space) and a token mid-sentence with surrounding text on both sides to confirm whitespace stays clean. Consider also normalizing trailing whitespace on drop, or document the asymmetry.",
      "confidence": 0.6
    },
    {
      "id": "smoke-scripts-cost-and-flakiness",
      "title": "Smoke scripts hit live Anthropic API with no caching, retries, or rate-limiting",
      "severity": "LOW",
      "category": "testing",
      "file": "apps/bot/scripts/smoke-data-shape-reply.ts",
      "description": "The three new smoke scripts each make 1–3 live calls to `claude-sonnet-4-5` with no retry, no determinism controls (no `temperature: 0`), and no caching. Validation checks are regex-based on stochastic LLM output, so checks like 'uses @nomadbera in prose or bullets' may pass on one run and fail on the next with identical inputs. If wired into CI these will be flaky and costly; if intended for local-only operator smokes, that intent isn't documented in scripts or package.json. Scenarios A and B in `smoke-resolved-handles.ts` also do not pass any history so the chat-mode opener rules apply uniformly, which is fine, but the lack of `temperature` control means shape-checks will drift.",
      "suggestion": "Set `temperature: 0` in the request body for determinism, document these as manual operator smokes (not CI), and consider snapshotting the output to disk for review rather than only regex-asserting. Add a top-of-file comment explicitly marking them as 'NOT for CI'.",
      "confidence": 0.85
    },
    {
      "id": "smoke-handle-regex-false-positive-risk",
      "title": "Handle-fabrication check uses overly permissive regex",
      "severity": "LOW",
      "category": "testing",
      "file": "apps/bot/scripts/smoke-resolved-handles.ts",
      "description": "The check `'does NOT fabricate a handle': !/@[a-z][a-z0-9_]+/i.test(textB)` will fail on any output containing an `@-mention` of any sort, including legitimate addressing like '@everyone' or even narrative references. While the scenario data has no usernames, the LLM could plausibly write '@channel' or use an @-prefix in a non-handle sense. Conversely, the resolved-case check `'uses @nomadbera'` is single-occurrence and case-insensitive; if the LLM renders `nomadbera` without `@` it fails even though that's still a correct handle use per the persona rules (priority a/b says discord_username OR display handle).",
      "suggestion": "Tighten the fabrication check to specifically forbid @-tokens that aren't in a small allow-list, or assert structurally (e.g., no `@\\w{3,}` other than the resolved usernames). For the positive check, accept either `@nomadbera` or `nomadbera`/`Nomad Bera` as a 'used handle' signal.",
      "confidence": 0.65
    },
    {
      "id": "smoke-scripts-relative-import-fragile",
      "title": "Smoke scripts use deep relative imports across workspace boundaries",
      "severity": "LOW",
      "file": "apps/bot/scripts/smoke-data-shape-reply.ts",
      "description": "All three new smokes import via `'../../../packages/persona-engine/src/persona/loader.ts'` — a 3-deep relative path crossing workspace boundaries with an explicit `.ts` extension. This couples the script directory layout to package internals, bypasses any package exports map, and will break if the persona-engine src layout changes. Other scripts in this repo presumably use the workspace package name.",
      "suggestion": "Import from the package name (e.g., `@freeside/persona-engine` or whatever the workspace name is) instead of a deep relative path. Verify the package's `exports` map exposes `buildReplyPromptPair`.",
      "confidence": 0.7
    },
    {
      "id": "persona-prompt-bloat",
      "title": "+182 lines added to persona.md — context window cost not measured",
      "severity": "LOW",
      "category": "performance",
      "description": "The DATA-SHAPED QUESTIONS section adds 182 lines (~3-5KB) to the system prompt sent on every reply call, including non-data-shaped ones (vibes, lore, character chat). The section includes verbose before/after examples and two duplicated example outputs (resolved + zero-resolution case). At Anthropic pricing this is a small but non-zero per-call cost, and longer system prompts can dilute instruction-following on unrelated requests. The smoke script logs `systemPrompt.length` but no validation enforces a budget.",
      "suggestion": "Consider whether the before/after examples and dual zero-resolution example could be condensed by ~30-50%, or moved to a separately-loaded fragment that's only injected when raw_stats are present in context. At minimum, capture the system-prompt length delta in a regression-testable way.",
      "confidence": 0.5
    },
    {
      "id": "praise-registry-bidirectional-fix",
      "title": "Registry-as-source-of-truth for emoji prefix correction is well-designed",
      "severity": "PRAISE",
      "category": "design",
      "description": "The non-capturing `(?:a)?` on the animated prefix, combined with deriving the rendered prefix from `entry.animated` via `renderEmoji`, elegantly fixes both directions of LLM mistakes (animated emoji emitted as static, static emoji emitted as animated) with a single regex pass. The BB F1 inverse test explicitly locks in this behavior. Good comment explaining the design intent.",
      "suggestion": "No change needed."
    },
    {
      "id": "praise-negative-smoke",
      "title": "Negative-case smoke for heuristic over-triggering",
      "severity": "PRAISE",
      "category": "testing",
      "file": "apps/bot/scripts/smoke-non-data-reply.ts",
      "description": "Including a negative smoke (`smoke-non-data-reply.ts`) that explicitly checks the data-shape rule does NOT activate for vibes/lore/opinion questions is the right move when adding a heuristic-gated formatting change. Without this, regressions where every reply becomes blockquote-formatted would only surface in production.",
      "suggestion": "No change needed."
    },
    {
      "id": "console-warn-side-effect",
      "title": "`console.warn` in hot path for every corrected/dropped emoji",
      "severity": "LOW",
      "category": "observability",
      "description": "The new pass calls `console.warn(...)` for every corrected ID and every dropped hallucination. If the LLM repeatedly hallucinates the same fake-IDs (which the operator screenshot suggests is happening), production logs will fill up with these warnings on every reply. There's no rate-limiting or aggregation, and no structured logger — these go to stdout/stderr directly. For tests, this will also pollute test output unless suppressed.",
      "suggestion": "Consider routing through a structured logger if one exists, or guarding with an env flag. At minimum, suppress `console.warn` in the test setup or use `vi.spyOn(console, 'warn').mockImplementation(...)` to keep test output clean.",
      "confidence": 0.7
    },
    {
      "id": "regex-id-length-mismatch",
      "title": "Snowflake length range `{17,20}` is correct for current Discord but consider mismatch with downstream",
      "severity": "SPECULATION",
      "category": "correctness",
      "description": "The new regex uses `\\d{17,20}` for snowflake IDs, matching the existing `idNormalized` pass. This is fine — Discord snowflakes are currently 17-19 digits and 20 is a safe upper bound. Worth confirming the registry's stored IDs cannot exceed this range; if `findByName(name).id` is longer, the corrected output would emit an ID that subsequently can't be re-matched by other passes. Probably not an issue in practice.",
      "suggestion": "Centralize the snowflake regex (e.g., `SNOWFLAKE_RE = /\\d{17,20}/`) to avoid drift between passes.",
      "confidence": 0.3
    }
  ]
}
```


## Callouts

PR #81, opened 2026-05-16
feat(cycle-005): post-merge UX pass + cross-week voice memory

zkSoju:
## Summary

post-merge nitpick pass triggered by operator's "i want to nitpick the UX for ruggy because there's alot we should be surfacing which i'm not seeing in the chats" — the autonomous run proved pipeline shape but never SHOWED what a real post looks like.

includes 3 commits:

1. **UX gallery + voice brief + sanitize integration** (`551d68b`) — author the missing T2.4 voice-brief module, build a gallery script that runs against LIVE prod score-mcp + Claude Sonnet 4.5 to generate real voice. fix 2 bugs surfaced by the gallery (`+-12` negative delta · `[zone] quiet week` engineering-jargon fallback). integrate sanitize.ts into the new `buildPulseDimensionPayload` path (underscore-escape on factor names · em-dash strip on voice surface).

2. **Cross-week voice memory** (`abae2de`) — operator latitude commit. ruggy becomes a continuous narrator. append-only JSONL writes each week's voice; next week's voice-brief threads the prior framing as continuity context.

## sample voice (live demo vs prod data window=30 with seeded W-19 memory)

```
bear-cave week N-1:  "26 events. the bears stir, then settle."
bear-cave week N:    "4 events. the bears settled deeper."
                     "next week: maybe a stir. maybe more silence."

el-dorado week N-1:  "1236 events but no rank-90 cross. the gold scatters."
el-dorado week N:    "past thirty days: 541 events, down from 1236.
                      the scattering has settled."
                     "the bears are deep in their dens. gold cools in the dark."

owsley-lab week N-1: "433 events past thirty days. the lab hums low."
owsley-lab week N:   "307 events past thirty days. the hum drops from 433.
                      the lab sleeps deeper."
                     "some weeks the bears just breathe."
```

without continuity: 4 isolated atomic posts per Sunday. with continuity: emotional arc across weeks (descent into quiet that's still being told). the LLM threads `settle → settled deeper`, recalls exact numbers (`from 1236`, `from 433`) as continuity grounding, and offers meta-reflection that comes from having context (`some weeks the bears just breathe`).

## bugs surfaced + fixed by the gallery

- **`+-12` negative delta** — unconditional `+` prefix doubled with `-` sign. now sign-aware: `+N` / `-N` / `±0`.
- **`[zone] quiet week` fallback** — engineering jargon broke the persona illusion for users with embeds disabled. now uses `ZONE_FLAVOR.emoji + name + dimension paren` (e.g. `🐻 Bear Cave (OG)`).
- **sanitize.ts bypass** — `buildPulseDimensionPayload` was skipping the `escapeDiscordMarkdown` + `stripVoiceDisciplineDrift` chain. factor names with underscores (`Boosted_Validator`) would mid-word italicize. now both pass through.

## test plan

- [x] full suite: 798 pass · 1 skip · 0 fail (+25 from voice-brief + voice-memory + continuity + 2 sanitize regression tests)
- [x] live gallery run against prod score-mcp + Claude Sonnet 4.5: 8 OTEL roots · 32 spans · 3 shape-A + 1 synthetic-shape-C posts rendered
- [x] cross-week memory roundtrip · multi-zone · malformed-line skip tested
- [ ] **operator-attested**: wire `composeDigestForZone` into `cron/scheduler.ts` so cron writes to `.run/ruggy-voice-history.jsonl` weekly
- [ ] **operator-attested**: live dev-guild post + visual sign-off

## still operator-attested deferral

- cron/scheduler.ts wire of composeDigestForZone (live behavior change)
- live dev-guild canary + S5-CANARY.md visual sign-off
- flip cycle-005 ledger active → archived
- AC-S5.2 chat-mode OTEL spans (V1.5 destination per routing invariant)

## new

- `packages/persona-engine/src/compose/voice-brief.ts` + `voice-brief.test.ts` + `voice-brief-continuity.test.ts`
- `packages/persona-engine/src/compose/voice-memory.ts` + `voice-memory.test.ts`
- `apps/bot/scripts/cycle-005-ux-gallery.ts` (run with `SCORE_API_URL=... MCP_KEY=... ANTHROPIC_API_KEY=... bun run`)

### DIFF
--- a/apps/bot/scripts/cycle-005-ux-gallery.ts
+++ b/apps/bot/scripts/cycle-005-ux-gallery.ts
@@ -38,6 +38,13 @@ import {
   buildVoiceBrief,
   parseVoiceResponse,
 } from '@freeside-characters/persona-engine/compose/voice-brief';
+import {
+  appendVoiceMemory,
+  readLastVoiceMemory,
+  formatPriorWeekHint,
+  isoWeek,
+  type VoiceMemoryEntry,
+} from '@freeside-characters/persona-engine/compose/voice-memory';
 import { initOtelTest } from '@freeside-characters/persona-engine/observability/otel-test';
 import type {
   PulseDimensionBreakdown,
@@ -271,6 +278,7 @@ function deriveBriefInput(
   shape: 'A-all-quiet' | 'B-one-dim-hot' | 'C-multi-dim-hot',
   isNoClaim: boolean,
   validationViolations: number,
+  priorWeekHint?: string,
 ): Parameters<typeof buildVoiceBrief>[0] {
   if (!dim) {
     return {
@@ -282,6 +290,7 @@ function deriveBriefInput(
       totalEvents: 0,
       windowDays: 30,
       previousPeriodEvents: 0,
+      priorWeekHint,
     };
   }
   const permitted = dim.top_factors
@@ -300,6 +309,7 @@ function deriveBriefInput(
     totalEvents: dim.total_events,
     windowDays: 30,
     previousPeriodEvents: dim.previous_period_events,
+    priorWeekHint,
   };
 }
 
@@ -321,16 +331,70 @@ async function main(): Promise<void> {
   const otel = initOtelTest();
   const zonesToRender: ZoneId[] = ['bear-cave', 'el-dorado', 'owsley-lab'];
 
-  // ─── SCENARIO 1: real prod data (what ships now) ─────────────────
+  // Seed last-week memory for the continuity demo (only if the file
+  // doesn't already exist — preserves real history if the gallery has
+  // been run before · gitignored under .run/).
+  const lastWeekSeeds: Array<VoiceMemoryEntry> = [
+    {
+      at: '2026-05-09T00:00:00Z',
+      iso_week: '2026-W19',
+      zone: 'bear-cave',
+      shape: 'A-all-quiet',
+      header: '26 events across thirty days. the bears stir, then settle.',
+      outro: 'the honey waits. see you in the deep.',
+      key_numbers: { total_events: 26, previous_period_events: 30, permitted_factor_names: [] },
+    },
+    {
+      at: '2026-05-09T00:00:00Z',
+      iso_week: '2026-W19',
+      zone: 'el-dorado',
+      shape: 'A-all-quiet',
+      header: '1236 events but no rank-90 cross. the gold scatters.',
+      outro: 'the prowlers prowl. back next sunday.',
+      key_numbers: { total_events: 1236, previous_period_events: 800, permitted_factor_names: [] },
+    },
+    {
+      at: '2026-05-09T00:00:00Z',
+      iso_week: '2026-W19',
+      zone: 'owsley-lab',
+      shape: 'A-all-quiet',
+      header: '433 events past thirty days. the lab hums low.',
+      outro: 'chain-actions cool. next week we check the readings.',
+      key_numbers: { total_events: 433, previous_period_events: 380, permitted_factor_names: [] },
+    },
+  ];
+  // Use a dedicated demo history path so we don't accumulate gallery noise
+  // in production-bound .run/ruggy-voice-history.jsonl
+  const demoHistoryPath = '/tmp/ruggy-ux-gallery-history.jsonl';
+  // Reset for fresh demo run
+  try {
+    const { unlinkSync, existsSync: ex } = await import('node:fs');
+    if (ex(demoHistoryPath)) unlinkSync(demoHistoryPath);
+  } catch {
+    /* ignore */
+  }
+  for (const e of lastWeekSeeds) {
+    await appendVoiceMemory(e, { path: demoHistoryPath });
+  }
+  console.log('');
+  console.log(`  seeded last-week voice memory at ${demoHistoryPath} (3 zones · iso_week=2026-W19)`);
+
+  // ─── SCENARIO 1: real prod data (what ships now) WITH continuity ──
   console.log('');
   console.log('━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━');
-  console.log('  SCENARIO 1 · REAL prod state (what users would see today)');
+  console.log('  SCENARIO 1 · REAL prod state + cross-week memory');
+  console.log('  (week N references week N-1 · ruggy threads continuity)');
   console.log('━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━');
   const realAllZones = buildAllZones(dimensions, 'real');
+  const thisWeek = isoWeek();
   for (const zone of zonesToRender) {
     const dimId = zone === 'bear-cave' ? 'og' : zone === 'el-dorado' ? 'nft' : 'onchain';
     const dim = dimensions.get(dimId);
     if (!dim) continue;
+
+    // Read last week's voice memory for this zone
+    const lastWeek = await readLastVoiceMemory(zone, { path: demoHistoryPath });
+    const priorHint = lastWeek ? formatPriorWeekHint(lastWeek) : undefined;
 
     // Pre-compose to know the shape (so the voice brief can match)
     const provisional = composeDigestForZone({
@@ -342,7 +406,7 @@ async function main(): Promise<void> {
       tracer: otel.tracer,
     });
 
-    const briefInput = deriveBriefInput(zone, dim, provisional.shape, provisional.isNoClaim, provisional.validation.violations.length);
+    const briefInput = deriveBriefInput(zone, dim, provisional.shape, provisional.isNoClaim, provisional.validation.violations.length, priorHint);
     const brief = buildVoiceBrief(briefInput);
     const voice = await generateVoice(brief);
 
@@ -356,8 +420,27 @@ async function main(): Promise<void> {
     });
 
     console.log(`\n[${zone}] shape=${finalResult.shape} · voice.via=${voice.via}`);
+    if (priorHint) console.log(`  ← prior week: "${lastWeek?.header}"`);
     renderPostAscii(zone, finalResult.payload, 'desktop');
     renderMobileAscii(finalResult.payload);
+
+    // Write this week's voice to memory (so next gallery run sees it)
+    await appendVoiceMemory(
+      {
+        at: new Date().toISOString(),
+        iso_week: thisWeek,
+        zone,
+        shape: finalResult.shape,
+        header: voice.header,
+        outro: voice.outro,
+        key_numbers: {
+          total_events: dim.total_events,
+          previous_period_events: dim.previous_period_events,
+          permitted_factor_names: briefInput.permittedFactors.map((f) => f.display_name),
+        },
+      },
+      { path: demoHistoryPath },
+    );
   }
 
   // ─── SCENARIO 2: synthetic-hot (what users WOULD see if activity ramps) ──
--- a/apps/bot/scripts/cycle-005-ux-gallery.ts
+++ b/apps/bot/scripts/cycle-005-ux-gallery.ts
@@ -1,0 +1,426 @@
+#!/usr/bin/env bun
+/**
+ * cycle-005 UX gallery — the missing "what does ruggy ACTUALLY look like".
+ *
+ * Authored 2026-05-16 in response to the operator's "i want to nitpick the
+ * UX for ruggy because there's alot we should be surfacing which i'm not
+ * seeing in the chats" message. The autonomous-run E2E proved the pipeline
+ * shape but never SHOWED what a real post looks like.
+ *
+ * This script:
+ *   1. fetches REAL prod factor_stats for all 3 dim-channel zones (window=30)
+ *   2. calls Claude SDK to generate the actual voice header + outro per
+ *      the cycle-005 voice-brief (T2.4 · the missing piece) — ONLY if
+ *      ANTHROPIC_API_KEY is set; otherwise uses canned fallback
+ *   3. composes via composeDigestForZone with REAL voice
+ *   4. renders the FULL Discord-formatted post (content + embed)
+ *   5. simulates mobile word-wrap (~40 chars in code blocks)
+ *   6. shows both the dominant REAL state (shape A at current volume) AND
+ *      a synthetic shape-C "what if activity ramps" scenario
+ *
+ * NO Discord delivery. Pure rendering for visual inspection.
+ *
+ * Run:
+ *   SCORE_API_URL=https://score-api-production.up.railway.app \
+ *   MCP_KEY=... \
+ *   ANTHROPIC_API_KEY=sk-... \
+ *   bun run apps/bot/scripts/cycle-005-ux-gallery.ts
+ */
+
+// Note: @anthropic-ai/claude-agent-sdk is resolved via the persona-engine
+// workspace symlink (apps/bot doesn't depend on it directly · transitive
+// resolution from the symlinked workspace package).
+import { query } from '../../../packages/persona-engine/node_modules/@anthropic-ai/claude-agent-sdk/sdk.mjs';
+import {
+  composeDigestForZone,
+} from '@freeside-characters/persona-engine/compose/digest';
+import {
+  buildVoiceBrief,
+  parseVoiceResponse,
+} from '@freeside-characters/persona-engine/compose/voice-brief';
+import { initOtelTest } from '@freeside-characters/persona-engine/observability/otel-test';
+import type {
+  PulseDimensionBreakdown,
+  PulseDimensionFactor,
+  ZoneId,
+  FactorStats,
+} from '@freeside-characters/persona-engine/score/types';
+
+// ─── MCP transport (copied from S0 spike pattern) ─────────────────────
+
+const MCP_PROTOCOL_VERSION = '2024-11-05';
+
+async function fetchDimensions(): Promise<Map<string, PulseDimensionBreakdown>> {
+  const SCORE_API_URL = process.env.SCORE_API_URL ?? 'https://score-api-production.up.railway.app';
+  const MCP_KEY = process.env.MCP_KEY;
+  if (!MCP_KEY) throw new Error('MCP_KEY not set');
+  const url = `${SCORE_API_URL.replace(/\/$/, '')}/mcp`;
+  const h = {
+    'Content-Type': 'application/json',
+    Accept: 'application/json, text/event-stream',
+    'X-MCP-Key': MCP_KEY,
+  };
+  const initRes = await fetch(url, {
+    method: 'POST',
+    headers: h,
+    body: JSON.stringify({
+      jsonrpc: '2.0',
+      id: 1,
+      method: 'initialize',
+      params: { protocolVersion: MCP_PROTOCOL_VERSION, capabilities: {}, clientInfo: { name: 'ux-gallery', version: '0' } },
+    }),
+  });
+  const sid = initRes.headers.get('Mcp-Session-Id');
+  if (!sid) throw new Error('no Mcp-Session-Id');
+  await initRes.text();
+  const sh = { ...h, 'Mcp-Session-Id': sid };
+  await fetch(url, {
+    method: 'POST',
+    headers: sh,
+    body: JSON.stringify({ jsonrpc: '2.0', method: 'notifications/initialized', params: {} }),
+  });
+
+  const out = new Map<string, PulseDimensionBreakdown>();
+  let nextId = 2;
+  for (const dim of ['og', 'nft', 'onchain'] as const) {
+    nextId += 1;
+    const r = await fetch(url, {
+      method: 'POST',
+      headers: sh,
+      body: JSON.stringify({
+        jsonrpc: '2.0',
+        id: nextId,
+        method: 'tools/call',
+        params: { name: 'get_dimension_breakdown', arguments: { window: 30, dimension: dim } },
+      }),
+    });
+    const body = await r.text();
+    const line = body.split(/\r?\n/).find((l) => l.startsWith('data: '));
+    if (!line) continue;
+    const env = JSON.parse(line.slice(6));
+    if (env.error) continue;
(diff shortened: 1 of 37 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#81. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Voice now bypasses sanitization in embed-path tests but ships through it in voice-outside-divs path",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/deliver/embed.ts",
      "description": "The new `sanitizeForDiscord(text, { isVoice: true })` is applied to voice header/outro before placing them in message.content. However, the test 'em-dash in voice header gets stripped' at embed-pulse-dimension.test.ts:393 asserts `payload.embeds[0]?.description` does not contain '—', but per the new doctrine the embed has NO description. The assertion will pass trivially (description is undefined → `.includes('—')` returns false), giving a false sense of voice-sanitization coverage. The em-dash stripping check is no longer testing what it claims.",
      "suggestion": "Update the em-dash test to assert `payload.content` (where voice now lives) does not contain '—'. Otherwise the regression guard is vacuous.",
      "confidence": 0.95
    },
    {
      "id": "F2",
      "title": "DeterministicEmbed @ts-expect-error test asserts wrong property",
      "severity": "LOW",
      "category": "test-quality",
      "file": "packages/persona-engine/src/live/discord-render.live.test.ts",
      "description": "The test 'DeterministicEmbed type rejects description at compile time' at line 38 uses `@ts-expect-error` correctly, but the assertion `expect(bad.color).toBe(0)` after a @ts-expect-error assignment doesn't validate the type rejection at runtime — if TS suppression were ever removed, the test would still pass. The directive is the actual test; the runtime assertion is filler.",
      "suggestion": "Either remove the runtime assertion (the @ts-expect-error directive IS the test) or add a comment clarifying the runtime line exists only to use the variable. Consider asserting `expect('description' in bad).toBe(true)` for runtime symmetry with the negative test above.",
      "confidence": 0.6
    },
    {
      "id": "F3",
      "title": "composeZonePost early-returns for digest before fetching digest metadata, losing zone color/footer pipeline",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/persona-engine/src/compose/composer.ts",
      "description": "Line 64-66: when postType === 'digest', `composeZonePost` short-circuits to `composeDigestPost` without ever using the existing `fetchZoneDigest`, `loadOrInitPersona`, or `buildPromptPair` pipeline. This means the silence-register override, post-type spec resolution, headline-lock, emoji-shortcode translation, and reaction-bar metadata previously composed for digests are all bypassed. The diff at L96-101 retains those paths only for non-digest types. If callers of `composeZonePost` for digests relied on `PostComposeResult.payload` having the legacy shape (e.g., embed.footer with zone color from `buildPostPayload`), they may now see different output. Verify all downstream consumers (cron, digest-once CLI, tests in composer.test.ts) handle the new orchestrator-derived payload.",
      "suggestion": "Audit all `composeZonePost(..., 'digest', ...)` call sites and confirm the new orchestrator's `DigestPostResult` (with `payload: DigestPayload` from `toDigestPayload`) is shape-compatible. In particular, confirm `embed.footer.text` is populated identically (the live renderer at discord-render.live.ts:121 uses 'digest · generated at...' which differs from `buildPostPayload`'s footer format).",
      "confidence": 0.85
    },
    {
      "id": "F4",
      "title": "Silence-register code path deleted without confirming feature parity",
      "severity": "MEDIUM",
      "category": "regression-risk",
      "file": "packages/persona-engine/src/compose/composer.ts",
      "description": "The removed block (lines 96-122 in the diff) handled the `isFlatWindow` → `pickSilenceTemplate` override for digest posts when stats showed a flat window. The new orchestrator path in digest-orchestrator.ts:36-39 only checks `config.VOICE_DISABLED` to skip voice — it never consults `isFlatWindow` or `pickSilenceTemplate`. If the operator relied on the silence-register for performed-silence framing in flat-window digests, that behavior is silently lost. The comment 'Digest now routes through orchestrator' at line 100 acknowledges the redirect but doesn't address feature parity.",
      "suggestion": "Either (a) port the silence-register logic into the orchestrator (e.g., in `createClaudeSdkLive` or as a port between snapshot-fetch and voice-gen), or (b) explicitly document that silence-register is deprecated and remove the unused `silence-register.ts` import/module to avoid dead code confusion.",
      "confidence": 0.85
    },
    {
      "id": "F5",
      "title": "Synthesized ZoneDigest in orchestrator loses real digest semantics",
      "severity": "MEDIUM",
      "category": "correctness",
      "description": "`snapshotToZoneDigest` (line 49+) fabricates a `ZoneDigest` by setting `window_start = window_end = computed_at = snapshot.generatedAt`, leaving `stale: false`, `narrative: null`, and `top_movers/top_events/spotlight` as empty arrays. Any downstream consumer expecting a real digest (e.g., reaction-bar metadata, rank-change rendering, top_movers narrative) will receive nonsense window bounds and empty arrays. This is a coercion that pretends to be a `ZoneDigest` but isn't.",
      "suggestion": "Either rename the result type to `DigestPostResult.snapshot` (without the legacy `digest: ZoneDigest` field), or fetch the real ZoneDigest alongside the snapshot if any consumer truly needs it. Pretending an empty shell is a ZoneDigest invites bugs where downstream code reads `digest.raw_stats.top_movers` and gets [].",
      "confidence": 0.8
    },
    {
      "id": "F6",
      "title": "Voice-brief assigns 'B-one-dim-hot' shape even when no factors are licensed",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/live/claude-sdk.live.ts",
      "description": "Line 21: `shape: snapshot.topFactors.length === 0 ? 'A-all-quiet' : 'B-one-dim-hot'`. The actual shape determination in `compose/digest.ts` (via `selectLayoutShape`) considers rank-90 thresholds + p95.reliable gates, not just topFactors count. As written, every non-empty snapshot lies to the voice-brief by claiming 'ONE DIM HOT' — which then loads `shapeBCGuidance` with 'LICENSED factors' even when `permittedFactors` is empty (filtered at line 23 with no rank gate). The result: the LLM is told 'ONE DIM HOT' with an empty licensed list, mismatched against shape-A guidance.",
      "suggestion": "Either compute the real shape via the layout-shape primitive before calling `buildVoiceBrief`, OR derive shape from the permitted-factors result: if `permittedFactors.length === 0`, use 'A-all-quiet' with `isNoClaimVariant: true`. Also: the current filter `.filter((factor) => factor.factorStats)` is too permissive — it should match `(rank >= 90 && p95.reliable)` per the voice-brief's licensing doctrine.",
      "confidence": 0.9
    },
    {
      "id": "F7",
      "title": "Voice-memory append on every gallery run will accumulate test noise in shared paths",
      "severity": "LOW",
      "category": "operational",
      "file": "apps/bot/scripts/cycle-005-ux-gallery.ts",
      "description": "The script at line 254 writes to `/tmp/ruggy-ux-gallery-history.jsonl` and deletes it first — good. But if any operator runs the gallery without overriding `RUGGY_VOICE_HISTORY_PATH` or omits the `{ path: demoHistoryPath }` option, the default `.run/ruggy-voice-history.jsonl` would get gallery output. The script correctly opts out, but the pattern is fragile: future code that calls `appendVoiceMemory()` without the path option will silently pollute production history.",
      "suggestion": "Consider adding a `dryRun` or `testMode` flag to `appendVoiceMemory`, OR namespace the history file by a `purpose` field at the call site (e.g., `appendVoiceMemory({ ..., source: 'cron' | 'gallery' | 'manual' })`) so production reads can filter. Not blocking.",
      "confidence": 0.5
    },
    {
      "id": "F8",
      "title": "isoWeek computation may produce incorrect week for ISO-week-53 edge cases",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/compose/voice-memory.ts",
      "description": "The `isoWeek` function at line 56 returns `${date.getUTCFullYear()}-W${weekNum}` but `date` has already been mutated to point to Thursday of the ISO week. For dates near year boundaries, ISO-8601 specifies that the year reported should be the year of that Thursday (which the algorithm does correctly), BUT for the rare case where week 53 of year N is reported, the formatting uses `getUTCFullYear()` of the mutated date — which is correct. The algorithm appears right but is fragile to read. The test only verifies a mid-year date (W02 in January and a generic May date), missing year-boundary cases.",
      "suggestion": "Add a test for an ISO-week-53 boundary (e.g., `2026-12-28` should be in `2026-W53` or `2027-W01` depending on the year). Cross-check against a known ISO-8601 reference.",
      "confidence": 0.4
    },
    {
      "id": "F9",
      "title": "Hardcoded transitive workspace import path in cycle-005-ux-gallery.ts is fragile",
      "severity": "MEDIUM",
      "file": "apps/bot/scripts/cycle-005-ux-gallery.ts",
      "description": "Line 30: `import { query } from '../../../packages/persona-engine/node_modules/@anthropic-ai/claude-agent-sdk/sdk.mjs';` reaches into a sibling workspace's node_modules. This breaks if (a) workspace hoisting reorganizes, (b) bun's nm layout changes, (c) the sdk export path changes. The comment acknowledges this is a hack ('transitive resolution from the symlinked workspace package').",
      "suggestion": "Either add `@anthropic-ai/claude-agent-sdk` as a direct dependency in apps/bot/package.json (since the script depends on it) or re-export `query` from persona-engine's public exports. The transitive node_modules path is a soft bomb for future workspace reshuffles.",
      "confidence": 0.85
    },
    {
      "id": "F10",
      "title": "PRAISE: cross-week voice memory adds genuine narrative depth with minimal infra",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/persona-engine/src/compose/voice-memory.ts",
      "description": "The append-only JSONL approach for cross-week voice continuity is well-scoped: zero infra, operator-readable, grep-able, and bounded growth (~210 lines/year). The `formatPriorWeekHint` produces a compact single-line context that adds real value without exploding token budgets. The 'MAY thread continuity OR pivot' instruction respects the LLM's creative agency rather than mechanically forcing references.",
      "suggestion": "Keep this pattern. If history becomes large, the file shape supports a trivial sqlite swap without touching the producer."
    },
    {
      "id": "F11",
      "title": "PRAISE: voice-outside-divs doctrine cleanly separates substrate truth from voice seasoning",
      "severity": "PRAISE",
      "category": "design",
      "file": "packages/persona-engine/src/deliver/embed.ts",
      "description": "Moving voice from embed.description to message.content is architecturally correct: embeds become pure substrate (deterministic, schema-stable), voice gets its own surface (mutable, character-shaped). The `DeterministicEmbed` type in domain/digest-message.ts:1 enforces this at compile time via intentional `description` omission — strong typing of the boundary. Composes well with the Discord-as-Material doctrine."
    },
    {
      "id": "F12",
      "title": "PRAISE: ports/live/mock separation enables clean orchestrator testing",
      "severity": "PRAISE",
      "category": "design",
      "description": "The hexagonal split (ports/score-fetch.port.ts, ports/voice-gen.port.ts, ports/presentation.port.ts with live/* and mock/* implementations) makes the VOICE_DISABLED contract testable with two clean lines: pass `createClaudeSdkMock`, assert `voice.calls.length === 0`. This is genuinely good engineering — the boundary is in the right place."
    },
    {
      "id": "F13",
      "title": "Stub dimension breakdown delta_pct calculation could divide by zero",
      "severity": "LOW",
      "category": "correctness",
      "description": "In `generateStubDimensionBreakdown` (line ~395), `delta_pct: ((total - previous) / previous) * 100` — when `previous` is 0, this produces Infinity/NaN. The factory uses `Math.max(1, 10 - i * 2)` for previous so it's currently always > 0, but if anyone tweaks the factory the math silently breaks. The wrapping dim-level `delta_pct` at line ~419 correctly guards against `previous === 0` (returns null), so the inner factor calc should mirror that guard.",
      "suggestion": "Mirror the null-when-zero guard: `delta_pct: previous === 0 ? null : ((total - previous) / previous) * 100`.",
      "confidence": 0.7
    },
    {
      "id": "F14",
      "title": "wrapMobile word-wrap heuristic may double-print whitespace tokens",
      "severity": "LOW",
      "category": "correctness",
      "file": "apps/bot/scripts/cycle-005-ux-gallery.ts",
      "description": "Lines 146-152: `line.split(/(\\s+)/)` keeps whitespace tokens as separate items, then appends each token to `current` checking width. When a long word breaks, `current` is pushed and the next word becomes `current` — but the whitespace token between them may end up as its own line if its length + previous makes it overflow. This is a gallery visualizer so visual cosmetics, but worth noting if anyone copies this pattern.",
      "suggestion": "If accuracy matters: use a tested word-wrap util (e.g., `wrap-ansi`). For gallery purposes, this is acceptable.",
      "confidence": 0.4
    },
    {
      "id": "F15",
      "title": "Removed @ts-expect-error in reaction-bar.ts may hide a real type issue",
      "severity": "LOW",
      "category": "type-safety",
      "file": "packages/persona-engine/src/deliver/reaction-bar.ts",
      "description": "The diff removes a `@ts-expect-error` at line 144 with no other type change shown in the diff. If the underlying type issue (channel.messages access) is still present, removing the suppression will produce a real TS error. If discord.js types were upgraded, the suppression was stale and removal is correct — but the diff doesn't show that. Verify the type checker passes without the directive.",
      "suggestion": "Confirm `bun run typecheck` or equivalent passes. If the `@ts-expect-error` was stale, this removal is good. If not, restore it.",
      "confidence": 0.5
    },
    {
      "id": "F16",
      "title": "buildShapeAPayload empty-dim fallback retains old embed.description path",
      "severity": "MEDIUM",
      "category": "consistency",
      "file": "packages/persona-engine/src/compose/digest.ts",
      "description": "The `isGenuinelyEmpty` branch (line 273-287) still places voice in `embeds[0].description`, contradicting the voice-outside-divs doctrine codified in embed.ts. The data-rich path correctly routes through `buildPulseDimensionPayload` (which puts voice in content). Two code paths now disagree on where voice lives for shape A.",
      "suggestion": "Unify: for empty-dim shape A, put voice in `content` and emit an empty embed (or omit embeds entirely). The doctrine should be uniform: 'voice always in content, never in embed.description'.",
      "confidence": 0.9
    },
    {
      "id": "F17",
      "title": "Top factor wallets column shows '?' when unique_actors is missing, but cohort percentile data still surrounds it",
      "severity": "LOW",
      "category": "UX",
      "file": "packages/persona-engine/src/deliver/embed.ts",
      "description": "`renderTopFactorRow` at line 245-247: when `factor.factor_stats?.cohort?.unique_actors` is undefined or null, the wallets column renders '?'. For factors without `factor_stats` at all (e.g., historic factors), the '?' is correct. But for factors with cohort data where unique_actors is the only missing field, '?' may be misleading vs '·' (used for null delta/rank). Inconsistency: `·` for missing rank/delta, `?` for missing wallets.",
      "suggestion": "Use `·` consistently for missing numeric data across all columns, or document why wallets gets '?' specifically (e.g., 'wallets must be a count, never an absence').",
      "confidence": 0.4
    },
    {
      "id": "F18",
      "title": "Voice generation in cycle-005-ux-gallery uses bypassPermissions without disclosure",
      "severity": "LOW",
      "category": "security-hygiene",
      "file": "apps/bot/scripts/cycle-005-ux-gallery.ts",
      "description": "Line 122: `permissionMode: 'bypassPermissions'` is set on the SDK call. For a local gallery script with no tool access this is fine, but it's a value worth keeping out of muscle memory — copy-pasted into a script that DOES grant tools, it would silently disable consent prompts. Worth a comment explaining why bypass is safe here (no tools registered, single-turn, no file system access).",
      "suggestion": "Add a one-line comment: `// safe: no MCP servers, no tool use, single-turn voice generation`.",
      "confidence": 0.5
    }
  ]
}
```


## Callouts

zkSoju (commented):
## Summary
Cycle-005 lands real voice infrastructure — cross-week memory, voice-brief doctrine, a clean ports/orchestrator seam — but ships the headline feature wiring incomplete: cross-week continuity is gallery-only on the production path (F-016), live shape derivation diverges from the canonical selector (F-002), and embed rendering now has two source-of-truth files (F-003). Verdict: **REQUEST_CHANGES**.

The orchestrator/ports split is the right long-term shape but currently coexists with composer.ts's imperative path for non-digest types. This is the Strangler Fig pattern Martin Fowler describes — productive only if the next cycle picks one direction. Without an ADR (F-014), the boundary will be re-litigated.

## Findings


```json
{"schema_version":1,"findings":[
{"id":"F-001","severity":"MEDIUM","category":"concurrency","file":"packages/persona-engine/src/compose/voice-memory.ts","title":"voice-memory appendFile non-atomic under future concurrency","metaphor":"O_APPEND is a deli ticket counter — atomic only if your order fits on one ticket (<PIPE_BUF).","faang_parallel":"Kafka's append-only log assumes single-writer per partition for the same reason; ext4 O_APPEND atomicity caps at PIPE_BUF (~4KB Linux)."},
{"id":"F-002","severity":"HIGH","category":"correctness","file":"packages/persona-engine/src/live/claude-sdk.live.ts","title":"shape derivation diverges from selectLayoutShape; silencedFactors always []","faang_parallel":"Classic 'parallel implementations of the same predicate' bug — Google's SRE book chapter on consistency calls this the 'two clocks' antipattern. The voice-brief gate is now a Potemkin guardrail.","metaphor":"Like installing a smoke detector whose wires don't reach the alarm. The label says SAFE but the signal never arrives.","teachable_moment":"When a doctrine ('silenced factors MUST NOT narrate') has two enforcement points, only the one in the live data path counts. Move shape+silencedFactors to a single derivation upstream of both renderer and prompt."},
{"id":"F-003","severity":"HIGH","category":"architecture","file":"packages/persona-engine/src/live/discord-render.live.ts","title":"two divergent renderers for the same digest message contract","faang_parallel":"Facebook's React Fiber rewrite kept the old reconciler alive for 18 months — a known cost. They mitigated with a single shared 'host config' interface. discord-render.live should delegate to buildPulseDimensionPayload, not re-derive widths.","metaphor":"Two recipes for the same dish in two cookbooks: every time the chef tweaks one, the other goes stale and the family argues at dinner.","connection":"Same drift surface cycle-005 UX r4 rules were written to prevent."},
{"id":"F-004","severity":"MEDIUM","category":"observability","file":"packages/persona-engine/src/compose/composer.ts","title":"composeDigestPost bypasses prose-gate + silence-register; telemetry lost","teachable_moment":"Short-circuiting one branch of a router silently retires its instrumentation. Either delete the dead callers or wire them through — leaving them in-tree as zombie code is worse than either."},
{"id":"F-005","severity":"MEDIUM","category":"correctness","file":"packages/persona-engine/src/orchestrator/digest-orchestrator.ts","title":"snapshotToZoneDigest fabricates window boundaries + schema_version","metaphor":"A passport with the photo printed but DOB left as today's date — accepted at the gate, fails any downstream check."},
{"id":"F-006","severity":"LOW","category":"type-safety","file":"packages/persona-engine/src/domain/digest-message.ts","title":"DeterministicEmbed doctrine is comment, not type","faang_parallel":"TypeScript's `never` branding is the same trick Rust's PhantomData uses to make illegal states unrepresentable."},
{"id":"F-007","severity":"LOW","category":"maintainability","file":"apps/bot/scripts/cycle-005-ux-gallery.ts","title":"deep relative import into another workspace's node_modules"},
{"id":"F-008","severity":"LOW","category":"correctness","file":"packages/persona-engine/src/compose/voice-memory.ts","title":"50-line tail-scan can silently lose zone history after ~13 weeks"},
{"id":"F-009","severity":"LOW","category":"security","file":"apps/bot/scripts/cycle-005-ux-gallery.ts","title":"fixed /tmp path unlink on multi-user host"},
{"id":"F-010","severity":"LOW","category":"correctness","file":"packages/persona-engine/src/deliver/embed.ts","title":"buildSnapshotField row inconsistency across zones"},
{"id":"F-011","severity":"MEDIUM","category":"doctrine","file":"packages/persona-engine/src/deliver/embed.ts","title":"message.content concatenation frames voice INSIDE substrate fallback","teachable_moment":"'Voice outside divs' is a layout claim AND an ordering claim. Confirm operator intent before locking the order."},
{"id":"F-012","severity":"PRAISE","category":"design","file":"packages/persona-engine/src/compose/voice-memory.ts","title":"JSONL append-only voice memory is infrastructure-light + grep-able","faang_parallel":"Honeycomb's wide-event log philosophy — Charity Majors' 'observability is built on append-only events you can ask new questions of.'"},
{"id":"F-013","severity":"PRAISE","category":"design","file":"packages/persona-engine/src/compose/voice-brief.ts","title":"negation-removal doctrine documented in code","faang_parallel":"Anthropic's own prompt guidance: positive instructions outperform negation. Encoding the WHY in the comment defends the design against future well-meaning reverts."},
{"id":"F-014","severity":"REFRAME","category":"architecture","file":"packages/persona-engine/src/orchestrator/digest-orchestrator.ts","title":"transitional dual-architecture needs an ADR or beads task","metaphor":"Strangler Fig only kills the host if you keep planting new branches in the same direction. Otherwise it's just a vine."},
{"id":"F-015","severity":"MEDIUM","category":"test-coverage","file":"packages/persona-engine/src/orchestrator/digest-orchestrator.test.ts","title":"no test pairs shape derivation against selectLayoutShape; priorWeekHint untested e2e"},
{"id":"F-016","severity":"HIGH","category":"feature-gap","file":"packages/persona-engine/src/orchestrator/digest-orchestrator.ts","title":"composeDigestPost reads/writes neither voice memory — feature gallery-only","faang_parallel":"Classic 'demo green, prod gray' gap — Google's launch-readiness checklist explicitly asks 'does the production code path execute the demoed code?' to catch exactly this.","metaphor":"The trailer shows the car flying. The car you bought doesn't have wings.","teachable_moment":"When a gallery script and a production orchestrator both compose the 'same' artifact, the gallery is a fixture, not a proof. Wire production first, then the gallery becomes a snapshot of real output."}
]}
```


## Callouts
- **[Design] voice-memory.ts** — append-only JSONL is the right shape: zero infra, grep-able, schema can grow. The "tail -10 to inspect" property is what Honeycomb calls observability-by-default.
- **[Design] voice-brief.ts** — "negation rules in prompts CAUSE the artifact they forbid" is real LLM craft, documented next to the code that depends on it. Future maintainers will not undo this by accident.

PR #82, opened 2026-05-16
feat(cycle-006): substrate-presentation refactor — type-level seam + voice-memory + Red Team ACs

zkSoju:
## Summary

cycle-006 closes 4 BB HIGH findings from PR #81 (F-002 shape drift · F-003 two renderers · F-016 voice-memory gallery-only · F-001 two-clocks pattern) by making the substrate-presentation seam load-bearing at the type level. Voice-memory becomes a first-class governance primitive with 8 streams, per-key mutex, path-traversal defense, 90-day TTL, and per-user deletion.

11 commits across 9 sprints (S0-S8) executed autonomously per `/simstim` workflow with operator delegation. Full audit trail in `grimoires/loa/cycles/cycle-006-substrate-presentation/COMPLETED.md`.

## Highlights

- **`DeterministicEmbed` has NO `description` field** — voice-outside-divs is now a compile error (`digest-message.compile-test.ts::@ts-expect-error`).
- **`deriveShape` is the single canonical source** (BB F-001 closure) — legacy `selectLayoutShape` deleted in S2 after 10/10 MATCH equivalence spike + permanent regression guard inlined in tests.
- **`composer.ts` is router-only** — 6 cron post types migrated to ports/live/mock orchestrators; 7th (chat-reply) has foundation + new entry point.
- **`buildPulseDimensionPayload` deleted** — 344 LoC removed (BB F-003 closure). Single renderer at `live/discord-render.live.ts`.
- **Voice-memory primitive** ships with TypeBox-style Zod schema, per-key mutex, `ALLOWED_STREAMS` allowlist (Red Team AC-RT-001), HTML-escaped `<untrusted-content>` markers + system-prompt instruction (Red Team AC-RT-002 + Flatline SKP-001/002 CRITICAL closure), 3-tuple `(guildId, channelId, userId)` chat-reply key (AC-RT-007), score-mcp snapshot plausibility check with rolling-window baseline (AC-RT-008 + Flatline SKP-003), 90-day TTL + `forgetUser` audit-logged deletion (Flatline SKP-004), and multi-process violation detection (Flatline SKP-001/HIGH).

## Red Team closure (Phase 4.5 · operator-attested)

| AC | Severity | Closure |
|---|---|---|
| RT-001 | 820 | `pathFor` validates stream against `ALLOWED_STREAMS` Set + key against `[A-Za-z0-9._:-]+` BEFORE `resolve()` |
| RT-002 | 880 | `formatPriorWeekHint` HTML-escapes `<`/`>`/`&` then wraps in `<untrusted-content>` markers · system prompt contains verbatim `UNTRUSTED_CONTENT_LLM_INSTRUCTION` |
| RT-007 | 830 | `keyForChatReply(guildId, channelId, userId)` 3-tuple · `ChatReplyVoiceMemoryEntrySchema` refinement requires `user_id` when stream === 'chat-reply' |
| RT-008 | 770 | `validateSnapshotPlausibility` 3σ check against rolling 30-snapshot baseline · NEUTERED snapshot on reject · `score.snapshot.implausible` OTEL event + `fallback_storm` alert |

## Flatline BLOCKER closure (Phase 6 · 8 of 8)

2 CRITICAL + 6 HIGH integrated. Full citations in `COMPLETED.md` §Flatline BLOCKER Closure.

## Test results

- **908 pass · 0 fail · 1 skip · 2315 expect calls across 53 files**
- `tsc --noEmit` clean (persona-engine + bot)
- `bun run lint:seam` exits 0 (warn-mode AND strict-mode)

## Test plan

- [ ] Bridgebuilder round 3 review surfaces findings
- [ ] Address HIGH/MEDIUM findings if any (PRAISE-only target)
- [ ] Production canary: one digest cron in THJ guild with voice-memory write verified
- [ ] Visual operator sign-off
- [ ] Operator merge + ledger flip `candidate → archived`

## Deferred to operator-paced post-merge

Per `grimoires/loa/cycles/cycle-006-substrate-presentation/operator-handoff-cycle-006.md`:

1. T6.7 mechanical wiring of 6 remaining orchestrators (digest is canonical exemplar in `digest-orchestrator.ts`)
2. T5.5/T5.7/T5.8 chat-reply `dispatch.ts` callsite migration + `composeReplyWithEnrichment` deletion + staging canary
3. T6.9 legacy `.run/ruggy-voice-history.jsonl` read-shim
4. Daily-pulse cron schedule wire-up (cadence + channel routing operator-decided)

## Spec deviations (per `feedback_spec_deviation_pattern`)

1. Zod for `VoiceMemoryEntry` schema (TypeBox not in deps · operator-attested)
2. fast-check → bun:test hand-crafted fixtures (S1 derive-shape tests · lib not in deps)
3. `sanitizeMemoryText` pulled S6 → S1 for co-location with `formatPriorWeekHint`
4. Chat-reply orchestrator delegates to legacy `composeReplyWithEnrichment` in S5 (full cutover deferred to post-merge to avoid high-traffic regression)
5. CI workflow not wired (no `.github/workflows/` in repo · `bun run lint:seam` is operator-runnable equivalent)

## Branch lineage

Cut from `origin/main` after PR #81 merged. 11 commits ahead of `origin/main` · 76 files changed · ~+10500/-3800 net.

🤖 Generated with [Claude Code](https://claude.com/claude-code) · autonomous /simstim workflow


zkSoju (commented):
## Summary

Cycle-006 ships a substrate↔presentation seam refactor with strong test discipline (legacy-as-oracle freezing, CI drift-guards, three-layer prompt-injection defense) and a coherent post-type taxonomy. Three HIGH items hold up cycle-close: a no-op ternary masking stream/entry drift, unescaped quotes in untrusted-content markers, and G-4 status diverging from the operator-handoff doc.

The structural pattern here is exemplary: freeze legacy behavior into test oracles, delete the source, guard the invariant with CI. This is what DeepMind formalized as "behavioral freeze before refactor" — eval harnesses pinning the old policy so improvements can't silently regress on inherited benchmarks. F-013 and F-016 are textbook applications.

The remaining concerns cluster at boundaries. F-001 (silent stream/entry mismatch) and F-002 (incomplete escape table) are both completeness gaps in defense-in-depth — the marker-wrapping does most of the work but leaves the attribute-quote vector ajar. F-005 (dual PostType definitions) is the architecturally load-bearing one — Hyrum's Law guarantees a future caller binds the wrong type when two share a name. F-003 (G-4 claim ≠ handoff doc) is the kind of mismatch that propagates into the next cycle's planning as "already done."

## Findings


{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "Tautological ternary in voice-memory appendEntry conceals stream-vs-entry mismatch",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/persona-engine/src/live/voice-memory.live.ts",
      "description": "In `appendEntry`, the path is computed as `const path = pathFor(stream, entry.stream === stream ? entry.key : entry.key)`. Both ternary branches return `entry.key`, making the conditional a no-op. The schema validates that `entry.stream` is a valid StreamName, but never enforces that it matches the `stream` parameter passed to `appendEntry`. A caller invoking `appendEntry('digest', entryWithStreamCallout)` will silently write to `.run/voice-memory/digest/<key>.jsonl` while the JSON record carries `stream: 'callout'`. AC-RT-001 path validation uses the `stream` parameter, but downstream readers may key off `entry.stream`, creating subtle data-integrity drift.",
      "suggestion": "Either remove the dead ternary (just call `pathFor(stream, entry.key)`), or — preferred — add an explicit guard before pathFor: `if (entry.stream !== stream) return { ok: false, reason: 'stream-mismatch: param=...' vs entry.stream=...' };`. The latter closes the silent-inconsistency window.",
      "confidence": 0.95,
      "faang_parallel": "Google's Spanner team treats `partition_param != row.partition` as a tablet-locator bug — same shape, where the addressing key and the payload's self-identifying key must reconcile or you get cross-partition reads in production.",
      "metaphor": "Like a courier writing the recipient's address on the envelope and a different address on the letter inside — both look authoritative, and which one wins depends on which clerk reads what.",
      "teachable_moment": "When a ternary's branches are identical, the code is shouting that the author *intended* to discriminate but forgot how. That's a load-bearing comment-in-disguise — promote it to a real guard or delete it.",
      "connection": "AC-RT-001 path validation is correct for the FS-write side; this gap is the read-side mirror image."
    },
    {
      "id": "F-002",
      "title": "Format-prior-week-hint escapes <,>,& but leaves quote unescaped — attribute-injection still possible",
      "severity": "HIGH",
      "category": "security",
      "file": "packages/persona-engine/src/orchestrator/format-prior-week-hint.test.ts",
      "description": "Test `'attacker stream value is escaped (cannot smuggle attrs)'` shows that quote characters are NOT escaped to `&quot;` — the assertion uses `'digest&quot;&gt;&lt;evil&gt;'.replace('&quot;', '\"')` which means the rendered output is `digest\"&gt;&lt;evil&gt;`. The literal `\"` survives. If any downstream consumer of the `<untrusted-content>` envelope treats attribute boundaries semantically (XML/HTML-aware tokenizer, LLM trained on structured prompts that respect attribute quoting, or future redactor that strips attribute values), the unescaped quote breaks attribute scope. The defense-in-depth claim against AC-RT-002 should escape all five XML metacharacters (<, >, &, \", ').",
      "suggestion": "Extend `escapeForUntrustedContent` (or equivalent) to also map `\"` → `&quot;` and `'` → `&#39;`. Update the test assertion to expect the fully-escaped form. This costs nothing and closes the attribute-injection vector that the markers explicitly attempt to defend.",
      "confidence": 0.85,
      "faang_parallel": "OWASP's XSS Prevention Cheat Sheet specifies all five metacharacters for *attribute context* precisely because XSS Filter Evasion (Samy worm, 2005) bypassed three-char escapers. Anthropic's own prompt-injection guidance treats `<untrusted-content>` markers as attribute-bearing tokens.",
      "metaphor": "You've put four locks on the door and skipped the deadbolt because 'who uses deadbolts anymore?' Every burglary report says: the deadbolt.",
      "teachable_moment": "Defense-in-depth has a completeness invariant: if you choose XML-style markers, you owe the full XML escape table. Partial escaping is worse than no escaping because it telegraphs that you considered the attack surface and stopped short.",
      "connection": "F-015 praises the three-layer defense; this finding is the single gap in that otherwise sound layering."
    },
    {
      "id": "F-003",
      "title": "G-4 'voice-memory production-wired' claim contradicts S6 deferral — only digest is wired",
      "severity": "HIGH",
      "category": "documentation",
      "file": "[REDACTED].md",
      "description": "COMPLETED.md asserts G-4 status ✅ with the bullet 'All 7 orchestrators wire voice-memory read (via formatPriorWeekHint from S1) + write'. But `operator-handoff-cycle-006.md` says explicitly: '**6 non-digest orchestrators not yet voice-memory-wired** (S6) — digest is the canonical exemplar. Mechanical inheritance pattern documented in digest-voice-memory.test.ts for replication.' The S3 orchestrator tests (`s3-orchestrators.test.ts`) confirm: none of micro/lore_drop/question/weaver/pop-in receive a voiceMemory dep, and there are no read-before-voice-gen or write-after-voice-gen assertions for them. F-016 (one of the three PR #81 BB HIGH findings) is closed for digest only, not the 'every post type' scope promised in PRD G-4.",
      "suggestion": "Either (a) downgrade G-4 status in COMPLETED.md from ✅ to '✅ PARTIAL — digest exemplar landed, 6 orchestrators inherit pattern in operator-paced follow-up' to match the operator-handoff doc, or (b) actually wire the 6 orchestrators before cycle close. The current state — green checkmarks contradicting the handoff doc — risks downstream cycles inheriting the gap as 'already done'.",
      "confidence": 0.95,
      "faang_parallel": "Amazon's PR/FAQ artifact discipline is rigorous about this exact failure mode: the launch doc must reconcile with the operational runbook before LR (Launch Readiness Review). 'Done according to plan' vs 'done according to ops' diverging is a Sev3 process bug at AWS.",
      "metaphor": "Two maps of the same city: one says the bridge is open, the other says it's under construction. Drivers trust whichever map they grabbed first. Eventually somebody hits the river.",
      "teachable_moment": "Cycle-close docs are read by the *next* cycle's planner, not by the current cycle's author. Inconsistency between COMPLETED and handoff is a defect that compounds — the next planner will choose whichever doc supports their preferred scope.",
      "connection": "This is the canonical 'identity-reality drift' KEEPER catches in the periodic tend cycle. Worth surfacing as a TEND-cycle trigger pattern."
    },
    {
      "id": "F-004",
      "title": "Multi-process invariant detection has TOCTOU race in cold-start",
      "severity": "MEDIUM",
      "category": "concurrency",
      "file": "packages/persona-engine/src/live/voice-memory.live.ts",
      "description": "`assertSingleProcess()` reads `.pid` file then proceeds. `recordPid()` writes the PID inside the work promise (after the assertion). Two processes starting within the same time window both see no `.pid` file, both pass assertSingleProcess, both proceed to record their own PID — the second overwrites the first. The header comment correctly states 'INVARIANT: single-process bot deployment', and the assertion guards against an *already-running* process, but cold-start contention is undefected. For long-running cron deployments this is unlikely to bite, but for restart scenarios (deploy rollover, supervisor restart) it's a real window.",
      "suggestion": "Use `openSync(pidFile, 'wx')` (exclusive create) or `proper-lockfile` semantics in recordPid — atomic acquire instead of read-then-write. The current `voice_memory.multi_process_violation` OTEL event will still emit on the steady-state case, but cold-start contention is closed. Cost: one extra fs call at startup.",
      "confidence": 0.7,
      "faang_parallel": "PostgreSQL's `postmaster.pid` uses `O_EXCL|O_CREAT` for exactly this scenario — Tom Lane's commit message in 2003 noted that 'read-then-write' had survived 7 years because nobody hot-restarted under load.",
      "metaphor": "Two people walking up to the same parking spot from different directions, both checking that it's empty, both pulling in. The 'is it empty?' check and the 'I'm parking here' action have to be the same atomic motion.",
      "teachable_moment": "TOCTOU windows hide in any check-then-act pair where the check is read-only. The fix is always 'collapse to a single atomic primitive' (O_EXCL, CAS, flock with LOCK_NB), never 'shrink the window.'",
      "connection": "Combined with F-017's reference-equality cleanup, the in-process locking is sound; this is the cross-process gap."
    },
    {
      "id": "F-005",
      "title": "PostType is defined in two places with diverging member names (legacy 'reply' vs domain 'chat-reply')",
      "severity": "MEDIUM",
      "category": "architecture",
      "file": "packages/persona-engine/src/domain/post-type.ts",
      "description": "`domain/post-type.ts` defines `PostType` as `'digest' | 'chat-reply' | 'micro' | 'weaver' | 'lore_drop' | 'question' | 'callout'`. But `composer.ts` imports `PostType` from `./post-types.ts` (legacy), and the runtime check `if (postType === 'reply')` uses the legacy spelling. This means there are effectively two PostType vocabularies coexisting — a future contributor who imports `domain/post-type.ts` and passes 'chat-reply' to composer.ts will get unexpected behavior. The FLATLINE-SKP-003/HIGH 'post-type count inconsistency' closure introduced the new domain canonical, but the composer didn't migrate to it.",
      "suggestion": "Either (a) replace `compose/post-types.ts::PostType` with a re-export of `domain/post-type.ts::PostType` and migrate 'reply' → 'chat-reply' in composer, OR (b) document explicitly in `domain/post-type.ts` that it is forward-looking and the legacy is canonical until cycle-007 unifies. A single canonical type is the SKP-003 spirit; two coexisting types is the symptom the closure was meant to address.",
      "confidence": 0.85,
      "faang_parallel": "The Linux kernel had `TASK_RUNNING` and `TASK_INTERRUPTIBLE` both treated as 'runnable' by some schedulers and 'sleeping' by others for years. Linus's 2.6.x consolidation memo: 'pick one, the other becomes a comment.' Same shape.",
      "metaphor": "Two dictionaries on the same shelf, both with your company's name on the spine, defining 'customer' differently. Eventually someone cites Dictionary A in a contract and Dictionary B in the lawsuit.",
      "teachable_moment": "Hyrum's Law has a corollary for type definitions: with two types sharing a name, the bound-to-symbol resolution becomes load-bearing on import order — and import order isn't part of any review checklist.",
      "connection": "F-006's return-shape inconsistency is the function-level twin of this type-level inconsistency."
    },
    {
      "id": "F-006",
      "title": "composeOrchestratedPost return-shape adapter is fragile — digest yields {voice}, others yield {message.voiceContent}",
      "severity": "MEDIUM",
      "file": "packages/persona-engine/src/compose/composer.ts",
      "description": "The switch in `composeOrchestratedPost` has two different result shapes: digest returns `{ payload, voice }` directly, while micro/lore_drop/question/weaver/callout return `{ payload, message: { voiceContent } }`. The adapter normalizes via `voice = r.message.voiceContent`, but the inconsistency means every new orchestrator must remember which shape its caller expects. The orchestrator return types are also not unified at the type level. A short-term comment ('S8 unifies') acknowledges the debt but the orchestrators are net-new code in this cycle — the unification would have been cheaper to do once during S3.",
      "suggestion": "Define a single `OrchestratedPostResult<TMessage>` shape that all orchestrators return — `{ postType, payload, message }` with `message` always carrying `voiceContent`. Digest can adopt this without changing its outer behavior; the composer adapter becomes a one-liner. Defer the legacy `voice: string` field for one cycle if needed.",
      "confidence": 0.75,
      "faang_parallel": "Stripe's API versioning team has a rule: 'do not ship two response shapes in one release.' The cost of unifying *during* the refactor is one engineer-hour; the cost of unifying *after* is one deprecation cycle.",
      "metaphor": "Two USB cables that fit the same port but carry different protocols. Plug-and-pray works until it doesn't.",
      "teachable_moment": "When you're building five new components and one legacy survivor, normalize on the new shape and adapt the survivor. Adapters at the new edge accumulate; adapters at the old edge sunset.",
      "connection": "The legacy `voice: string` is digest's tail; F-005's dual PostType is the same pattern at the type layer."
    },
    {
      "id": "F-007",
      "title": "N+1 score fetches per cron firing — orchestrators call fetchDigestSnapshot per zone instead of a batch",
      "severity": "MEDIUM",
      "category": "performance",
      "file": "packages/persona-engine/src/orchestrator/weaver-orchestrator.ts",
      "description": "The weaver-orchestrator test asserts `expect(fetchedZones).toHaveLength(4)` — confirming 4 separate `fetchDigestSnapshot(zone)` calls per weaver compose. The digest-orchestrator likely does the same (focal + 3 others for cross-zone deriveShape). Each call is an MCP round-trip to score-mibera in production. For weaver weekly + digest weekly + each other zone-aware post-type, total score-mcp QPS multiplies by 4×. The SDD §3.2 + §3.9 referenced `score.fetchAllZoneSnapshots()` as a single batch fetch but the implementation appears to call per-zone — the test mock doesn't even expose `fetchAllZoneSnapshots`.",
      "suggestion": "Add `fetchAllZoneSnapshots(): Promise<ReadonlyArray<DigestSnapshot>>` to ScoreFetchPort that does ONE MCP call returning all 4 zones in a single envelope (score-mibera already supports per-window all-zones aggregation). Update orchestrators that need cross-zone context (digest, weaver) to use it instead of N independent calls. Saves ~3× round-trips per compose.",
      "confidence": 0.7,
      "faang_parallel": "Facebook's GraphQL was born from precisely this N+1: mobile clients making 40 REST calls per news-feed render. DataLoader's coalesce-within-tick pattern is the standard fix.",
      "metaphor": "Going to the grocery store four times for one item each instead of writing a list. Works fine until the store is on the other side of town.",
      "teachable_moment": "N+1 read patterns hide behind well-named per-item functions. The smell is always 'a loop in the caller that one level down would be a JOIN.'",
      "connection": "Compose throughput today is dominated by LLM latency, so this isn't the bottleneck — but it becomes one the first time cron horizontals."
    },
    {
      "id": "F-008",
      "title": "Pop-in default picker is zone-deterministic — same zone always picks same post type",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/orchestrator/s3-orchestrators.test.ts",
      "description": "Test `'default pickType is deterministic per zone'` confirms `r1.postType === r2.postType` for two consecutive runs in the same zone. If the production cron uses the default picker, every pop-in firing in zone X will surface the same post-type forever (or until code change). This is fine for tests but probably not the desired production behavior — pop-ins are described as 'random cron' in the PRD §FR-5. Without time-component in the seed, the variety is zero.",
      "suggestion": "If determinism is for test repeatability only, make production picker accept an injectable RNG (`Math.random` default; tests inject a fixed-seed source). If determinism is intentional product behavior (e.g. per-zone narrative-arc invariant), document the choice explicitly in pop-in-orchestrator.ts since the cycle PRD's 'random' framing suggests otherwise.",
      "confidence": 0.65,
      "faang_parallel": "Spotify's recommendation team learned this in 2014: a 'random' shuffle that was deterministic-per-user generated 30% of support tickets because users heard the same 'random' first song every session.",
      "metaphor": "A dice with the same number on all six faces. Technically you rolled.",
      "teachable_moment": "Determinism is a desirable property of test code and undesirable property of variety-driven UX. The injection-pattern lets the same code serve both without ambiguity."
    },
    {
      "id": "F-009",
      "title": "shortenWallet applies 0x-style ellipsis to non-hex long IDs (misleading rendering)",
      "severity": "LOW",
      "category": "ux",
      "file": "packages/persona-engine/src/live/discord-render.live.ts",
      "description": "`shortenWallet` fallback branch shortens any string longer than 12 chars with the same `slice(0,6)…slice(-4)` format. The test asserts `shortenWallet('aaaaaaaaaaaaaaaaaa')` → `'aaaaaa…aaaa'`. If an `event.wallet` field ever carries a non-wallet identifier (UUID, ENS name, internal handle), it'll be rendered as if it were a wallet — visually indistinguishable from a real address. Discord users may mistake the rendered form for an address.",
      "suggestion": "Add a stricter guard: shortening only applies when input matches `^0x[0-9a-fA-F]+$`. For everything else, pass through unchanged (or truncate to a different glyph, e.g. `name…` instead of `0xname…ID`). Cost: one regex branch.",
      "confidence": 0.65,
      "metaphor": "Calling every long string 'an address' is like calling every long number 'a phone number.' Works for the common case, awkward when the user sees their UUID rendered as 0xabc…def.",
      "teachable_moment": "Format-from-shape is a UI design smell when shape doesn't determine identity. Always format from typed identity (wallet vs handle vs id), never from string-length heuristics."
    },
    {
      "id": "F-010",
      "title": "renderActivityPulse hardcodes English fallback when events empty",
      "severity": "LOW",
      "category": "ux",
      "file": "packages/persona-engine/src/live/discord-render.live.ts",
      "description": "`if (pulse.events.length === 0) return { content: 'no recent events in the ledger.' };` is a substrate-only output, but the string is in English and not parameterized by character voice. The doctrine ('voice outside divs') would suggest even the empty-state should route through the character's voice surface — at the very least, the literal string deserves a comment explaining why pulse is the one post type that doesn't follow the orchestrator's voice-port pattern. Future operators may delete the empty-events branch entirely if they assume pulse-orchestrator handles it.",
      "suggestion": "Either (a) move the empty-events check into pulse-orchestrator so the renderer is purely substrate-shape, OR (b) add a comment in renderActivityPulse explaining that pulse is intentionally voice-less and the empty-state is the renderer's single English string.",
      "confidence": 0.5,
      "teachable_moment": "Doctrine boundaries that hold for the 99% case need explicit exception markers for the 1% — otherwise the 1% looks like a bug to anyone enforcing the doctrine."
    },
    {
      "id": "F-011",
      "title": "renderActivityPulse truncation edge case yields '\\n\\n…' when first event > cap",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/live/discord-render.live.ts",
      "description": "If the first event's `block.length + 2 > DISCORD_CONTENT_CAP`, the loop breaks immediately with `blocks = []`. The return is `truncated ? content + '\\n\\n…' : content` where `content = blocks.join('\\n\\n') = ''`. Final output is `'\\n\\n…'` — two leading newlines + ellipsis. Mobile Discord renderers may collapse the empty lines, but the substantive content is missing entirely. Not a critical breakage but the user sees a single ellipsis with no context.",
      "suggestion": "Guard the truncation branch: if `blocks.length === 0`, return `{ content: '(events truncated · no display available)' }` or similar fallback. Alternatively, force-include the first event even if it exceeds cap, with its own truncation suffix.",
      "confidence": 0.55,
      "metaphor": "A weather report that says '…' when it's too windy to describe. Users would prefer 'too windy to describe.'"
    },
    {
      "id": "F-012",
      "title": "AC-RT-007 schema enforcement claim is only partially closed",
      "severity": "MEDIUM",
      "category": "security",
      "file": "packages/persona-engine/src/live/voice-memory.live.test.ts",
      "description": "COMPLETED.md says: 'AC-RT-007 closure (keyForChatReply 3-tuple + ChatReplyVoiceMemoryEntrySchema refinement requires user_id when stream === chat-reply)'. The `voice-memory.live.test.ts` covers schema validation (rejects 2.0.0 version, rejects too-long header) but has NO test asserting that a chat-reply stream entry without `user_id` is rejected. The operator-handoff doc lists 'Schema-enforced (guildId, channelId, userId) for chat-reply' under S6 deferred items. So the schema refinement may not actually be in place yet — the key shape exists (S5 foundation), but the schema-enforcement claim in COMPLETED is unverified.",
      "suggestion": "Add an explicit test case to `voice-memory.live.test.ts`: `appendEntry('chat-reply', entryOf({ stream: 'chat-reply', key: 'g:c:u' /* no user_id */ }))` must return `{ ok: false, reason: schema... }`. If the test fails, implement the discriminated-union schema (ChatReplyVoiceMemoryEntrySchema) that requires user_id. If the test passes, AC-RT-007 schema enforcement is real and COMPLETED is accurate.",
      "confidence": 0.7,
      "faang_parallel": "Google's SRE handbook §17 on testing for reliability: 'an unverified safety claim is a safety bug.' Same shape as F-003 at the test layer.",
      "teachable_moment": "AC closure must be backed by a single specific test that would fail if the AC regressed. 'Schema validation tested' isn't 'this AC's schema-validation tested.'"
    },
    {
      "id": "F-013",
      "title": "Three-suite test design for deriveShape (oracle + legacy + completeness) is exemplary",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/persona-engine/src/domain/derive-shape.test.ts",
      "description": "The test file splits coverage into (1) oracle equivalence — derived in a different style from the production impl, deliberately to catch transcription bugs in either direction; (2) legacy equivalence — inlines the cycle-005 `selectLayoutShape` body VERBATIM after S2 deletes the module, providing a permanent regression guard for FLATLINE-SKP-001/HIGH; (3) output completeness — covers permittedFactors/silencedFactors/diagnostic counts. The 30 hand-crafted scenarios span all 4 decision-tree branches with boundary fuzz. The legacy-equivalence suite is particularly nice — it captures the cycle-005 behavior at the test-fixture level so future refactors can't silently drift.",
      "suggestion": "Adopt this pattern for the next migration that retires legacy code: inline-the-deleted-code-as-a-test-only-oracle is a generalizable technique for closing the SKP-001 risk class.",
      "faang_parallel": "DeepMind's evaluation harness for AlphaFold v1 → v2 inlined v1's scoring code into the v2 test suite for exactly this reason — Joel Spolsky's 'Things You Should Never Do' rewrites only work if the old behavior survives as a fixture.",
      "teachable_moment": "Deleting code does not delete its behavior contract. Inline-as-oracle is the cheapest form of behavioral archaeology."
    },
    {
      "id": "F-014",
      "title": "composer-router drift-guard test catches new-post-type-without-orchestrator regression",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/persona-engine/src/compose/composer-router.test.ts",
      "description": "The CI test asserting `for each pt in ZONE_ROUTED_POST_TYPES: MIGRATED_POST_TYPES.has(pt)` catches the failure mode where a new PostType is added but no orchestrator is wired — the cycle-005 transitional dual-architecture risk that REFRAME concern F-014 (PR #81 BB review) flagged. This is the minimum-viable structural guard the cycle promised, and it's wired correctly.",
      "faang_parallel": "Google's BUILD-rule completeness checks: every `cc_library` listed in a `deps` set must exist. The drift-guard here is the same idea at the TypeScript-types layer.",
      "teachable_moment": "Structural invariants want CI guards, not docstrings. The test costs 50 lines; the absence of the test costs a silent prod regression."
    },
    {
      "id": "F-015",
      "title": "AC-RT-002 layered defense (sanitize + HTML-escape + system-prompt instruction) is well-constructed",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/persona-engine/src/orchestrator/format-prior-week-hint.test.ts",
      "description": "The three-layer defense — (a) sanitizeMemoryText strips control bytes + NFKC normalize at write, (b) HTML-escape <,>,& at hint-format read, (c) UNTRUSTED_CONTENT_LLM_INSTRUCTION snapshot-pinned into the voice-gen system prompt so the LLM is told to treat marker content as inert — closes the prompt-injection vector at three independent layers. The snapshot-test on the LLM instruction string prevents silent erosion of the convention. F-002 above identifies one gap (unescaped quote), but the overall layering is sound and replicable.",
      "faang_parallel": "Anthropic's own injection-defense guidance recommends stacked layers exactly because no single layer is airtight. Same pattern Cloudflare uses for WAF rule composition.",
      "teachable_moment": "Independent layers > one strong layer. The snapshot-pinned instruction is the layer that's hardest to silently erode — text changes are visible in code review in a way schema validators aren't."
    },
    {
      "id": "F-016",
      "title": "Rolling-window baseline pattern for AC-RT-008 avoids static-fixture drift",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/persona-engine/src/domain/validate-snapshot-plausibility.test.ts",
      "description": "FLATLINE-SKP-003 closure correctly identifies that a static baseline rejects organic score-mcp growth over weeks. The implementation uses a rolling window (last 30 snapshots, refreshed on accept) with stats computed at read-time. The 60-week +10%/week organic-growth simulation that produces zero false-positives across the trajectory is a strong validation. The companion negative test ('static baseline WOULD produce false-positives') keeps the SKP-003 rationale visible in code-as-doc form.",
      "faang_parallel": "Netflix's anomaly detection uses adaptive baselines for the same reason — viewing patterns drift seasonally; static thresholds drown SREs in false positives by week 4.",
      "teachable_moment": "Baselines need a half-life. The 'companion test that proves the old approach would fail' is a brilliant fossil — it preserves the decision rationale in executable form."
    },
    {
      "id": "F-017",
      "title": "voice-memory mutex cleanup-by-reference (SKP-001/850) implementation is correct",
      "severity": "PRAISE",
      "category": "concurrency",
      "file": "packages/persona-engine/src/live/voice-memory.live.ts",
      "description": "The mutex cleanup pattern — store the LockChain wrapper as the Map value, compare by reference (`keyLocks.get(lockKey) === myChain`) in finally — correctly handles the scenario where a later writer queued during our work. The naive cleanup (`delete` unconditionally) would orphan the queued writer's await; the cycle-005 SDD review notes flagged this as a hazard, and the fix here is the right shape. Combined with TypeBox maxLength 280 on header/outro keeping serialized lines under PIPE_BUF, the SKP-001/850 closure is sound (in single-process — see F-004 for multi-process gap).",
      "faang_parallel": "Java's ReentrantLock uses owner-thread equality for the same reason — releasing-not-owned-by-me is a logic bug that corrupts the queue. JS Promises mimicking this via reference-equality is the idiomatic translation.",
      "teachable_moment": "Mutex cleanup is asymmetric: acquisition is 'put if absent,' release is 'remove only if I'm still the holder.' Symmetric cleanup is the bug pattern."
    },
    {
      "id": "F-018",
      "title": "Chat-reply key shape may exceed 64-char schema cap as Discord IDs grow",
      "severity": "SPECULATION",
      "category": "robustness",
      "file": "packages/persona-engine/src/domain/voice-memory-entry.ts",
      "description": "SDD §3.6 specifies `key: Type.String({ minLength: 1, maxLength: 64 })`. Discord snowflake IDs are currently 18-19 chars. `keyForChatReply(g, c, u)` returns `<g>:<c>:<u>` → 19+1+19+1+19 = 59 chars at typical. Borderline. Any future extension to include thread-id, or any future Discord snowflake-bit increase, breaks the cap. Not actionable today but worth tracking.",
      "suggestion": "Bump the maxLength to 128 (defense in depth costs nothing; storage trivial), OR document the 64-char limit explicitly with the rationale and the breakage scenario in voice-memory-entry.ts.",
      "teachable_moment": "Schema caps with 5-char headroom against external identifiers are tomorrow's incident. The cost to widen is zero; the cost to widen *after* shipping is a migration."
    },
    {
      "id": "F-019",
      "title": "PRD/SDD/Sprint files moved to cycle subdirectory — consider whether top-level prd.md should be a stub or removed",
      "severity": "LOW",
      "category": "documentation",
      "description": "The top-level `grimoires/loa/prd.md` is overwritten with cycle-006 content (+396 -201) and identical content also lives at `grimoires/loa/cycles/cycle-006-substrate-presentation/prd.md` (+425 -0). Same situation for sdd.md and sprint.md (truncated). Two copies of the same document on disk creates drift risk — future edits may land in one and not the other. The cycle-005 pattern (presumably) kept the top-level files as the active-cycle pointer; for cycle-006 the duplication is now committed.",
      "suggestion": "Pick one canonical location: either (a) top-level prd/sdd/sprint.md ALWAYS mirror the active cycle (and the cycle subdirectory holds frozen copies at cycle-close), OR (b) top-level files become a stub/pointer that says `current cycle: cycle-006 → see grimoires/loa/cycles/cycle-006-substrate-presentation/prd.md`. The current state (full duplication) is the worst of both.",
      "teachable_moment": "Two-locations-for-the-same-doc converges to drift on a timescale of weeks. Either symlink, single-source, or stub-with-pointer — never duplicate."
    }
  ]
}


## Callouts

Five PRAISE findings carry the cycle's quality story. F-013 (legacy-as-oracle) is the technique that makes refactor-by-deletion safe — Joel Spolsky's "Things You Should Never Do, Part I" warned against rewrites because behavior evaporates; the inline-the-old-code-as-a-fixture pattern is the antidote. F-016 (rolling-baseline) extends the same idea to security policy: the companion negative test ("static baseline WOULD produce false-positives") preserves the design rationale in executable form, which is rarer and more valuable than prose-only ADRs. F-014 (composer-router drift-guard), F-015 (three-layer prompt-injection defense), and F-017 (mutex cleanup-by-reference) round out a cycle that takes invariant preservation seriously. The single weakest seam — F-005's dual PostType — is the natural next-cycle followup.

PR #84, opened 2026-05-17
cycle-007: agent debuggability through medium-aware layering [S8 pending]

zkSoju:
draft PR for cycle-007 — 7 of 9 sprints landed · S8 cycle close pending

## sprint status

| id | scope | status | commit |
|---|---|---|---|
| S0 | calibration spikes (zone-registry audit · android typography) | ✓ | `3814419` |
| S1 | D1 zone-registry + INV-12 TS-AST lint + manifest | ✓ | `3814419` · `cd041ad` |
| S2 | D2 trace envelope (INV-14 + INV-15 + audit) | ✓ | `cd041ad` |
| S3 | D3 medium extensions + D5 figure-space padding | ✓ | `5ae85d7` |
| S4 | D4 trace CLI (5 subcommands) | ✓ | `7e0e6e3` |
| **S5** | **dashboard layer-color + SSE-behind-flag + AC-RT-001** | **✓ this PR** | `d924bba` · `8e3c642` · `50dba8c` · `972ad63` |
| S6 | FR-1 sanitizer hook (log-only V1) | ✓ | `99af6c4` |
| S7 | orchestrator port cleanup (cycle-006 G-6 closure) | ✓ | `99af6c4` |
| S8 | cycle close (COMPLETED.md · BB-3 · production canary · ledger flip) | ⏸ pending | — |

## what S5 ships

`scripts/dashboard.ts` rewrite (+752/-154 r1 · +102/-33 r2 · +52/-15 r3 · ~940 LoC final) carrying:

- **T5.1** — reader dedup via new `resolveTraceFilePath` + generic `readJsonl<T>` exports from `scripts/lib/trace-readers.ts` (shared with `scripts/trace.ts`)
- **T5.2** — 6-color oklch border-left encoding (Alexander INV-10 palette) on every row
- **T5.3** — 4-panel layer-split detail (substrate/voice/presentation/medium-render) with cross-layer connectors matched by run_id or zone+5min window
- **T5.4** — `sanitizeForTerminal` on SSE payloads (AC-RT-003) · `textContent`-only client rendering (SKP-001/CRITICAL DOM-XSS closure)
- **T5.5** — `/sse` behind `LOA_DASH_SSE=1` with full hardening stack:
  - `hostname:'127.0.0.1'` explicit bind (IMP-005)
  - per-session bearer token + HttpOnly + SameSite=Strict cookie bootstrap (Phase 6 SKP-002)
  - `Host` allowlist (AC-RT-001 DNS-rebinding defense)
  - perimeter Origin check (belt-and-suspenders behind SameSite=Strict)
  - per-token cap of 1 + evict prior connection (AC-RT-010)
  - global `MAX_CLIENTS=5` cap + 60s heartbeat (BB MEDIUM-2)
  - 500ch payload truncation (BB MEDIUM-3)
  - `crypto.timingSafeEqual` token compare + SHA-256 fingerprint

## quality gates

- `scripts/dashboard.test.ts` — 19 pass · 0 fail · 53 expect() calls
- full repo: 1024 pass · 1 skip · 0 fail · 2622 expect() calls
- `bun run lint:cycle-007` — INV-12 ✓ · INV-17 ✓ · INV-14 ✓
- `bunx tsc --noEmit` — exit 0

## review trail

- engineer review found 2 blocking + 5 non-blocking concerns: `grimoires/loa/a2a/cycle-007-sprint-5/engineer-feedback.md`
- r2 resolved both blockers + 3 non-blockers
- security audit surfaced 2 additional findings (HIGH parseCookie URIError DoS · MEDIUM sseScanTick containment); both fixed inline in r3
- audit verdict: **APPROVED**: `grimoires/loa/a2a/cycle-007-sprint-5/auditor-sprint-feedback.md`
- COMPLETED marker at `grimoires/loa/a2a/cycle-007-sprint-5/COMPLETED`

## what S8 will add

- T8.1 cycle-wide `COMPLETED.md` summary
- T8.2 E2E acceptance test (FR-7 paste-row-to-Loa workflow)
- T8.3 BB round 3 multi-model review
- T8.4 operator-attested production canary (FR-2 mobile screenshot)
- T8.5 ledger flip `active → candidate`

## test plan

- [ ] checkout branch, `bun install`, `bun test` — verify 1024 pass
- [ ] `bun run lint:cycle-007` — verify clean
- [ ] `LOA_DASH_SSE=1 bun run scripts/dashboard.ts` — verify token printed to stderr · curl bootstrap works · browser opens dashboard with layer-color rows
- [ ] verify AC-RT-001: `curl -H 'Host: attacker.example' http://localhost:3001/api/llm-trace` → 403
- [ ] verify cookie bootstrap: `curl -i -X POST -H 'X-Loa-Dash-Token: <stderr-token>' http://localhost:3001/api/auth` → Set-Cookie HttpOnly · SameSite=Strict
- [ ] verify SSE: with `LOA_DASH_SSE=1` enabled, fire `bun run digest:once` → new row flashes layer color in dashboard


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#84. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "INV17-PATH-MISMATCH",
      "title": "CODEOWNERS + monotonic check protect .claude/data/ but manifest lives at .claude/overrides/ — INV-17 neutered",
      "severity": "CRITICAL",
      "category": "security",
      "file": ".github/CODEOWNERS",
      "description": "CODEOWNERS protects `.claude/data/voice-prompt-paths.json` and `.claude/data/voice-prompt-paths.schema.json`, but `scripts/lint-no-kebab-zoneid-in-voice-prompt.ts:69` reads `MANIFEST_PATH = join(REPO_ROOT, '.claude/overrides/voice-prompt-paths.json')`, and the newly-added manifest + schema files live at `.claude/overrides/`. The `.claude/data/` path is a symlink into the Loa submodule (per the comment in the lint script) and is NOT where the operative manifest lives. The same path is referenced by `scripts/lint-manifest-monotonic.sh` per its CI step. Net effect: an attacker (or sloppy PR author) can modify `.claude/overrides/voice-prompt-paths.json` to narrow the INV-12 scan scope without triggering CODEOWNERS required-review AND without failing the monotonic check (which is watching the wrong path). This is the exact attack class Red Team AC-RT-002 / INV-17 was designed to close. The defense is path-mismatched and inoperative.",
      "suggestion": "Update `.github/CODEOWNERS` to reference `.claude/overrides/voice-prompt-paths.json` + `.claude/overrides/voice-prompt-paths.schema.json` + `.claude/overrides/trace-explain-output.schema.json` (the actual locations). Update `scripts/lint-manifest-monotonic.sh` to walk git history of the `.claude/overrides/` paths. Add a guard in the lint script that fails-closed if both `.claude/data/voice-prompt-paths.json` AND `.claude/overrides/voice-prompt-paths.json` exist (ambiguous source of truth). Verify with a synthetic narrowing PR that CI now blocks it.",
      "confidence": 0.95
    },
    {
      "id": "APPEND-TRACE-MUTEX-CLAIM",
      "title": "appendTraceEntry semantics claim 'concurrent writers serialize via Bun.write internal queueing' but no mutex visible — SDD also promised SKP-003/HIGH in-memory mutex",
      "severity": "HIGH",
      "category": "correctness",
      "file": "packages/persona-engine/src/observability/trace-envelope.ts",
      "description": "The `appendTraceEntry` doc comment in the SDD section §2.2 claims best-effort single-process invariant + 'concurrent writers in the same process serialize via Bun.write internal queueing.' Flatline Phase 6 SKP-003/HIGH amendment in `sprint.md` line ~870 promised: 'appendTraceEntry uses in-memory promise-chain mutex to serialize ALL writes within process · guarantees atomic line appends · ~15 LoC.' The shown implementation in the diff is just `await Bun.write(filePath, line, { append: true })` with no visible promise-chain mutex. Bun.write with `append: true` does not document line-level atomicity guarantees across concurrent in-process callers — two `Bun.write` calls may interleave bytes mid-line if scheduled in overlapping microtasks, corrupting JSONL parsing for the dashboard + CLI + reader-tolerance fixture. Either the mutex landed in a part of the file the diff truncated, or it was dropped silently from the implementation.",
      "suggestion": "Verify the full `trace-envelope.ts` source has a per-file promise-chain mutex (e.g., `Map<filePath, Promise<void>>` chained via `.then`). If absent, add it: ~15 LoC, no new deps. Add a concurrent-write stress test (e.g., 50 parallel `appendTraceEntry` calls, verify all 50 JSON.parse cleanly post-hoc). Update the SDD §2.2 doc comment to either match the implementation or update the implementation to match the promise. The single-process invariant is load-bearing for INV-14 + the cycle's debuggability narrative.",
      "confidence": 0.7
    },
    {
      "id": "REPO-ROOT-NOT-CACHED",
      "title": "findRepoRoot() walks filesystem on every trace-file resolution — hot-path performance concern in dashboard fetch handler",
      "severity": "MEDIUM",
      "category": "performance",
      "file": "scripts/lib/trace-readers.ts",
      "description": "`CANDIDATE_RUN_DIRS()` calls `findRepoRoot()` on every invocation. `findRepoRoot()` does `realpathSync(startDir)` + up to 10 iterations of `existsSync` + `readFileSync` + `JSON.parse`. `resolveTraceFilePath` calls `CANDIDATE_RUN_DIRS()` on every call. `dashboard.ts` fetch handler calls `resolveTraceFilePath` per endpoint per request. Each `/api/llm-trace` request triggers ~5-20 sync filesystem operations just for path resolution before any actual data read. For a long-running dashboard with active operator interaction (polling at 2s + SSE), this is wasted I/O. Not catastrophic but worth caching.",
      "suggestion": "Cache `findRepoRoot()` result in a module-scoped `let _repoRoot: string | null = null` and return it on subsequent calls. Same for `CANDIDATE_RUN_DIRS()` result. The repo root and run-dir paths are immutable for the process lifetime; one-time computation is sufficient.",
      "confidence": 0.85
    },
    {
      "id": "REDUNDANT-ORIGIN-CHECK-SSE",
      "title": "handleSse's originAllowed check is dead code when AUTH_ENABLED — perimeter check fires first",
      "severity": "LOW",
      "category": "code-quality",
      "file": "scripts/dashboard.ts",
      "description": "The r2 fix moved the Origin check to the perimeter in the main `fetch()` handler: `if (AUTH_ENABLED && !originAllowed(req)) return forbidden-origin`. But `handleSse(req)` (line 590+) still does its own `if (!originAllowed(req)) return forbidden-origin`. When AUTH_ENABLED is true (default), the perimeter check rejects bad origins before they reach handleSse — the inner check is dead code. When AUTH_ENABLED is false (test fixture only), the inner check could fire, but that path also bypasses authentication entirely so the Origin protection is moot. Minor cleanup; not load-bearing.",
      "suggestion": "Remove the redundant `originAllowed` call inside `handleSse` since the perimeter check now covers it. Or, if the intent is defense-in-depth for the AUTH_ENABLED=0 fixture path, add a comment explaining why the inner check exists.",
      "confidence": 0.9
    },
    {
      "id": "HEARTBEAT-NOT-TUNABLE-FOR-TESTS",
      "title": "HEARTBEAT_MS = 60_000 hardcoded prevents wall-clock verification in CI — only structural proof available",
      "severity": "LOW",
      "category": "testability",
      "file": "scripts/dashboard.ts",
      "description": "The 60-second SSE heartbeat cadence (BB MEDIUM-2 acceptance) is hardcoded as `HEARTBEAT_MS = 60_000`. The test suite acknowledges this in `scripts/dashboard.test.ts` and falls back to a structural proof (verifying the initial `: hello\\n\\n` comment arrives, which co-locates with the `setInterval` scheduling). The COMPLETED.md explicitly flags this as 'AC-T5.5-G ⚠ Partial (structural proof, not 60s wall-clock).' A short follow-up: expose `HEARTBEAT_MS` as an env override (e.g., `DASHBOARD_HB_MS`) to enable a real cadence test with `DASHBOARD_HB_MS=500` in CI without burning 60s.",
      "suggestion": "Change `const HEARTBEAT_MS = 60_000` to `const HEARTBEAT_MS = Number(process.env.DASHBOARD_HB_MS ?? 60_000)`. Add a test that sets this to 250ms and asserts a `event: ping` event arrives within 1s. Closes the AC-T5.5-G partial-status downgrade.",
      "confidence": 0.8
    },
    {
      "id": "SANITIZE-ONLY-KNOWN-STRING-FIELDS",
      "title": "sanitizePayloadFields hardcodes 4 string fields — nested or future payload strings bypass ANSI sanitization",
      "severity": "LOW",
      "category": "security",
      "file": "scripts/dashboard.ts",
      "description": "`sanitizePayloadFields` (line ~194-200) and `truncatePayloadFields` (line ~175-184) hardcode the 4 fields `system_prompt | user_message | output | error`. If a future trace shape includes a top-level string payload field with arbitrary name (e.g., `voice_brief`, `tool_call_arg`), or nested string fields inside an object payload, ANSI escape bytes in those values reach the SSE consumer un-sanitized. The trace envelope's `sanitizeNestedReservedKeys` handles reserved-key spoofing but does NOT sanitize nested string values for terminal escapes. For the current trace schema this is fine; the gap surfaces if the schema grows. Defense-in-depth: the client renders all values via textContent, so DOM XSS is still closed, but a chat-paste of a poisoned row could still cause terminal escape injection if the operator pipes the JSON to `pbcopy` and then echos it.",
      "suggestion": "Either (a) document this as a known scoped limit (acceptable for cycle scope), or (b) walk all string values in the payload recursively (depth-bounded like sanitizeNestedReservedKeys does for reserved keys) and apply sanitizeForTerminal. Option (b) is ~10 LoC and closes the residual surface. Acceptable to defer to cycle-008.",
      "confidence": 0.5
    },
    {
      "id": "FIXTURE-TOKEN-IN-TESTS",
      "title": "Test fixture pins FIXTURE_TOKEN to a static UUID — not a security issue but worth a comment",
      "severity": "LOW",
      "category": "code-quality",
      "file": "scripts/dashboard.test.ts",
      "description": "Test fixture pins `FIXTURE_TOKEN = 'fixture-token-d7a3-9e8f-4c1b-aa22'` (line ~12). This is a static string used to bootstrap the cookie auth flow in tests. Not a security concern — tests aren't running against production secrets. But the comment is misleading-by-absence; a casual reader might think this is a production constant. Worth a one-line comment.",
      "suggestion": "Add a comment above FIXTURE_TOKEN clarifying: `// Pinned test-fixture token (not a secret · used only by test subprocess via env override).`",
      "confidence": 0.95
    },
    {
      "id": "RUN-DIR-CANDIDATES-COMPUTED-PER-CALL",
      "title": "CANDIDATE_RUN_DIRS() is a function (computed on every call) but its inputs are immutable for the process lifetime",
      "severity": "LOW",
      "category": "performance",
      "file": "scripts/lib/trace-readers.ts",
      "description": "`CANDIDATE_RUN_DIRS()` is invoked on every `resolveTraceFilePath` call. Internally it does an `findRepoRoot()` call AND checks `process.env.TRACE_RUN_DIR`. Both inputs are stable for process lifetime. Computing this lazily-but-once would avoid repeated walks. Related to REPO-ROOT-NOT-CACHED but a separate caching point.",
      "suggestion": "Memoize `CANDIDATE_RUN_DIRS` with a module-scoped cache that initializes lazily. Combine with the `findRepoRoot` caching fix.",
      "confidence": 0.85
    },
    {
      "id": "PRAISE-DEFENSE-IN-DEPTH-AC-RT-001",
      "title": "AC-RT-001 closure is exemplary — 3 stacked defenses against DNS rebinding",
      "severity": "PRAISE",
      "category": "security",
      "file": "scripts/dashboard.ts",
      "description": "The DNS-rebinding defense layers `hostname: '127.0.0.1'` Bun.serve bind (kernel-level loopback-only) + Host header allowlist (application-level rejection of attacker.example) + per-session bearer token via HttpOnly + SameSite=Strict cookie (credential-level, attacker can't read or send). Each layer fails closed — if any one is bypassed, the others remain. Combined with `crypto.timingSafeEqual` for constant-time token comparison + cookie bootstrap requiring out-of-band stderr-printed token, this closes the AC-RT-001 attack surface comprehensively. The architectural reasoning is documented inline at the auth helpers (lines 88-90, 116-117, 132-134, 170-174).",
      "suggestion": "No change. This is the right shape for any localhost-bound HTTP service that should never accept cross-origin requests.",
      "confidence": 1.0
    },
    {
      "id": "PRAISE-AST-LINT",
      "title": "INV-12 TS-AST lint catches JS escape-sequence + template-literal + char-code bypass classes that grep cannot",
      "severity": "PRAISE",
      "category": "security",
      "file": "scripts/lint-no-kebab-zoneid-in-voice-prompt.ts",
      "description": "The Red Team AC-RT-004 attack class was: an attacker can write `'\\u0065l-dorado'` or `` `el${'-'}dorado` `` or `String.fromCharCode(101,108,45,...)` to leak a kebab ZoneId past a grep-based lint. The TS-AST rewrite at `scripts/lint-no-kebab-zoneid-in-voice-prompt.ts` walks the AST and resolves string literals through escape sequences (TS parser pre-decoded), template literals with constant parts, binary concatenation, AND String.fromCharCode chains via the `evaluateAsString` recursive evaluator. The skip-only-on-import-statement-line rule (Phase 6 SKP-001/CRITICAL) closes the original bash-grep false-negative where any registry-importing file got a full pass. PRAISE-3 (zero new deps) holds because `typescript` was already a deps. ~340 LoC well-shaped, with clear comments mapping each check to the originating quality gate.",
      "suggestion": "No change. Worth distilling the AST-evaluation pattern as a reusable utility if a sibling lint emerges (e.g., for prose-leak detection of other identifier classes).",
      "confidence": 1.0
    },
    {
      "id": "PRAISE-REDOS-BENCHMARK",
      "title": "ReDoS benchmark test pins linear-time regex behavior — catches future-developer accidents",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/persona-engine/src/domain/zone-registry.test.ts",
      "description": "Two ReDoS benchmark tests in `zone-registry.test.ts` (10K-char pathological input + adversarial dash-variant mix) assert `detectKebabZoneIds` completes in <50ms wall-clock. This pins the linear-time behavior of the current regex implementation and would fail-closed if a future refactor introduces a backtracking-explosion pattern. Flatline IMP-002 (845 severity) is properly closed.",
      "suggestion": "No change. Worth noting in `docs/trace-cli.md` that the detector has documented ReDoS resistance.",
      "confidence": 1.0
    },
    {
      "id": "PRAISE-NFKC-DASH-SUBSTITUTION",
      "title": "BB HIGH-2 Unicode dash substitution + NFKC normalization closes the U+2010..U+2015 + U+2212 bypass class",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/persona-engine/src/domain/zone-registry.ts",
      "description": "`detectKebabZoneIds` normalizes input via `.normalize('NFKC')` and substitutes Unicode dash variants (`/[‐-―−]/g` covering U+2010 HYPHEN through U+2015 + U+2212 MINUS) to ASCII hyphen BEFORE regex matching. Tests cover all 4 dash variants (U+2010, U+2013, U+2014, U+2212) with passing assertions. Pattern mirrors cycle-098 sprint-7 cypherpunk HIGH-2 (L7 SOUL prescriptive matching) — consistent application of the same defense class across the Loa ecosystem. The Cyrillic homoglyph V1 gap is explicitly pinned with a test and deferred to V2 with documentation.",
      "suggestion": "No change. The V2 follow-up (UTS #39 skeleton transform for Cyrillic) is correctly scoped.",
      "confidence": 1.0
    },
    {
      "id": "PRAISE-TEXTCONTENT-CLIENT-RENDERING",
      "title": "Client-side DOM rendering uses textContent only — closes SKP-001/CRITICAL DOM-XSS surface structurally",
      "severity": "PRAISE",
      "category": "security",
      "file": "scripts/dashboard.ts",
      "description": "The client-side renderer in `dashboard.ts` uses a small `el(tag, props, ...children)` helper that always writes to `node.textContent` for any string value. There is zero usage of `.innerHTML` or `document.write` on payload-derived content. Combined with the server-side `sanitizePayloadFields` (defense-in-depth), DOM XSS via trace payloads is structurally impossible — even if a payload contains `<script>alert(1)</script>` or `javascript:` URLs, the renderer treats it as text. This is the right shape for any untrusted-content viewer.",
      "suggestion": "No change. Worth documenting this discipline in `docs/trace-cli.md` so future contributors don't introduce innerHTML paths.",
      "confidence": 1.0
    },
    {
      "id": "PRAISE-RECONNECT-STORM-EVICTION",
      "title": "Per-token cap of 1 + evict-prior connections is a clean reconnect-storm defense — r2 test strengthening proves eviction",
      "severity": "PRAISE",
      "category": "security",
      "file": "scripts/dashboard.ts",
      "description": "`evictSameTokenConnections(fingerprint)` runs BEFORE the global `MAX_CLIENTS` cap check on every new SSE connection. This composes correctly with AC-RT-001 token requirement: un-tokened connections are rejected at auth before counting toward any cap. Same-token reconnect from a legitimate browser tab closes the prior orphaned connection and accepts the new one without exhausting the global cap. The r2 test strengthening at `dashboard.test.ts` proves `controller.close()` propagation via reading the prior reader and asserting `done: true`. Closes AC-RT-010 with the right shape.",
      "suggestion": "No change. The fingerprint hash via SHA-256 (r2 ISSUE-4 fix) is the right choice over the prior raw-token-suffix.",
      "confidence": 1.0
    },
    {
      "id": "REFRAME-DOCUMENTATION-VOLUME",
      "title": "PRD + SDD + sprint plan integrate ~22 BB findings + 14 Flatline SDD findings + 10 Red Team attacks + 19 Flatline sprint findings — pattern is heavy but consistent with cycle-006 precedent",
      "severity": "REFRAME",
      "category": "process",
      "file": "grimoires/loa/cycles/cycle-007-agent-debuggability/sdd.md",
      "description": "The SDD is 1083 lines. The PRD is 628 lines. The sprint plan is 840 lines. A large fraction of each document is review-integration tables: BB Design Review Integration Log (Phase 3.5), Flatline SDD Review Integration Log (Phase 4), Red Team SDD Integration Log (Phase 4.5), Flatline Sprint Review Integration Log (Phase 6). Each integration log table is itself ~30-60 rows with severity codes + amendment-landed pointers. This documentation pattern is consistent with cycle-006 precedent and reflects the operator's chosen quality-gate process — not a code issue. The reframe question: would a future cycle-008+ reader benefit more from a canonical post-review PRD/SDD with a single 'changelog' appendix, vs the current per-phase integration logs scattered through the document? The current structure preserves provenance (which finding drove which amendment) at the cost of read-flow. Worth surfacing as a process observation, not blocking the cycle.",
      "suggestion": "After cycle-007 close, consider an experiment in cycle-008: canonical PRD/SDD body reflects post-review state; review-integration provenance lives in a sibling `review-history.md`. Test whether future-readers find the canonical version easier to navigate. Not a change to this PR.",
      "confidence": 0.6
    },
    {
      "id": "SPECULATION-TRACE-ENVELOPE-AS-LOA-PRIMITIVE",
      "title": "TraceLayer + wrapTraceEntry + appendTraceEntry pattern could become a Loa framework primitive across all freeside-* projects",
      "severity": "SPECULATION",
      "category": "architecture",
      "file": "packages/persona-engine/src/observability/trace-envelope.ts",
      "description": "The trace envelope pattern (5-layer enum + wrap helper with INV-15 nested-key sanitization + sole-permitted-append-helper INV-14 + reader-tolerance for legacy rows) is independent of mibera-specific domain. Every Loa-using project that wants agent-debuggable traces could consume the same primitive. The 5-layer set ('substrate' | 'voice' | 'presentation' | 'medium-render' | 'orchestrator') is specific to freeside-characters' shape but the WRAPPER + APPEND DISCIPLINE is generic. Worth surfacing as a candidate Loa framework primitive at cycle close, alongside the upstream-propose for freeside-mediums (D3 cycle-close action).",
      "suggestion": "Consider whether `trace-envelope.ts` belongs in a `@loa/observability` package after cycle-007 closes. The taxonomy stays project-local; the envelope mechanism + audit-script pattern is reusable. Defer to operator decision at cycle-007 wrap-up.",
      "confidence": 0.5
    }
  ]
}
```


## Callouts

zkSoju (commented):
## Summary

This cycle ships substantive substrate work (zone canonicalization, trace envelope, medium-aware layering, dashboard auth hardening) with rigorous multi-phase review trail. Findings cluster around the playground subprocess surface (resource governance, env hygiene) and reviewability of a 77-file PR. Security posture is strong; the gaps are operational governance, not architectural.

The dashboard auth story is exemplary — three-layer DNS-rebinding defense with HttpOnly+SameSite cookies bootstrapping a constant-time-compared bearer is the pattern Google's IAP uses for localhost dev tools, and the INV-12 TS-AST lint upgrade closes a real bypass class that the bash grep would have missed (escape sequences, template-literal constants, fromCharCode chains). The playground, however, introduces a subprocess-spawn surface that the rest of the dashboard's threat model didn't anticipate.

Architectural meditation: the substrate→presentation seam (zone-registry, trace envelope, medium-extensions) lands cleanly because each layer owns one transformation. The playground breaks that discipline slightly — it's a developer-iteration surface stapled to a debugging dashboard. Worth naming explicitly so future cycles know which side of the seam it lives on.

## Findings


```json
{"schema_version":1,"findings":[
{"id":"SEC-001","severity":"HIGH","category":"security","file":"scripts/dashboard.ts:248-289","title":"Playground fire spawns subprocess without rate limit or queue","description":"Auth alone is not resource governance — ~10 rapid authed POSTs spawn 10 parallel 30s subprocesses. Auth gate ≠ DoS gate.","suggestion":"Per-process semaphore (DASHBOARD_PLAYGROUND_MAX_CONCURRENT, default 2) returning 429. Optional token-bucket per fingerprint.","faang_parallel":"Stripe's idempotency-key + concurrency-limit pattern on /v1/charges: auth proves identity, semaphore proves the caller can't accidentally DoS themselves.","metaphor":"Lockable rooms with no occupancy limit — a guest with a keycard can still flood the fire marshal.","confidence":0.7},
{"id":"SEC-002","severity":"MEDIUM","category":"security","file":"scripts/dashboard.ts:303","title":"Playground subprocess inherits full process.env","description":"`env: { ...process.env }` passes ANTHROPIC_API_KEY, MCP_KEY, DISCORD_BOT_TOKEN into the child. One future log-on-error regression in playground-fire.ts leaks secrets to .run/playground/*.json.","suggestion":"Allowlist: { PATH, HOME, NODE_ENV, STUB_MODE:'true', LLM_PROVIDER:'stub' }. Same shape as L3 LOA_L3_PHASE_ENV_PASSTHROUGH.","faang_parallel":"Borg's per-task minimal env injection — production workloads receive only declared inputs, never the orchestrator's bag.","teachable_moment":"`{...process.env}` is convenient, but each inherited var is a future leak waiting for a logger.","confidence":0.75},
{"id":"SEC-003","severity":"MEDIUM","category":"security","file":"scripts/dashboard.ts:235-238","title":"Path containment uses string-prefix, not realpath","description":"Safe today given runId regex, but a one-character regex relaxation reopens traversal. Defense-in-depth says canonicalize.","suggestion":"`realpathSync(path).startsWith(realpathSync(PLAYGROUND_RUNS_DIR)+'/')` — same pattern as scripts/trace.ts:167.","metaphor":"Like checking a UUID matches a prefix instead of looking up the row — works until the schema changes.","confidence":0.5},
{"id":"QUAL-001","severity":"MEDIUM","category":"quality","file":"scripts/dashboard.ts:319-325","title":"Timeout path discards subprocess stderr","description":"504 response carries no diagnostic. Exit-code-2 path reads JSON but never consumes stderr. Hung-playground debugging is blind.","suggestion":"Bounded stderr read (200ms / 4KB cap) before kill; include snippet in 504 and exit-2 responses.","faang_parallel":"AWS Lambda's INIT_REPORT capture-then-kill — never lose the last 4KB before SIGKILL.","confidence":0.7},
{"id":"QUAL-002","severity":"LOW","category":"quality","file":"scripts/dashboard.ts:218,866","title":"PLAYGROUND_POST_TYPES duplicated server↔client","description":"Two lists must stay in sync manually; silent UX drift when one diverges.","suggestion":"Inject via template: `const PLAYGROUND_POST_TYPES = ${JSON.stringify([...PLAYGROUND_POST_TYPES])};`.","confidence":0.85},
{"id":"QUAL-003","severity":"LOW","category":"quality","file":"scripts/dashboard.ts:1280-1295","title":"SSE-vs-poll deviation undocumented in SDD","description":"COMPLETED.md notes it; SDD §2.6 doesn't. Future readers will re-litigate.","suggestion":"One-line SDD amendment naming the scope of suppression (llm-trace only).","confidence":0.6},
{"id":"QUAL-004","severity":"MEDIUM","category":"quality","file":"PR-level","title":"77 files / +6700 LoC strains single-PR reviewability","description":"Substrate + consumers + docs + ops in one envelope; reviewer cognitive load high.","suggestion":"Future cycles: (a) substrate+contracts, (b) consumers (dashboard/CLI/playground), (c) docs. Flag for visibility — operator may have validated this shape.","faang_parallel":"Linux kernel patch series convention: one logical change per commit, series for review.","confidence":0.5},
{"id":"QUAL-005","severity":"LOW","category":"quality","file":"scripts/dashboard.test.ts:218-237","title":"Heartbeat test proves wiring, not cadence","description":"~10 LoC fix closes the Partial-status caveat.","suggestion":"`HEARTBEAT_MS = Number(process.env.DASHBOARD_HB_MS ?? 60_000)`; test with HB_MS=200, assert ping in 500ms.","confidence":0.85},
{"id":"QUAL-006","severity":"LOW","category":"quality","file":"scripts/lint-no-kebab-zoneid-in-voice-prompt.ts:75-95","title":"Dual-path guard is correct but unfinished","description":"Guard prevents .claude/data/ shadow at runtime; structural CI assertion is cheaper and louder.","suggestion":"CI step: `test ! -e .claude/data/voice-prompt-paths.json`. Or inline doc with BB round-3 link.","confidence":0.5},
{"id":"QUAL-007","severity":"LOW","category":"quality","file":"packages/persona-engine/src/observability/trace-envelope.ts","title":"sanitizeNestedReservedKeys depth-cap is silent","description":"Depth ≥32 returns value un-sanitized; no warn, no metric. Pathological inputs survive invisibly.","suggestion":"`console.warn` or OTEL counter `trace_envelope.sanitize_depth_capped` on cap.","confidence":0.4},
{"id":"PRAISE-001","severity":"PRAISE","category":"security","file":"scripts/dashboard.ts:153-172","title":"Cookie bootstrap over query-param token","description":"HttpOnly+SameSite=Strict closes proxy-log, DevTools, document.location, and Referer leaks. Constant-time compare on both header and cookie paths. Operator-friendly bootstrap_help with copy-pasteable curl. Craft-level security work.","confidence":0.95},
{"id":"PRAISE-002","severity":"PRAISE","category":"quality","file":"scripts/lint-no-kebab-zoneid-in-voice-prompt.ts","title":"TS-AST lint catches four bypass classes bash grep would miss","description":"Escape sequences, template literals with constant parts, fromCharCode, binary concat. Honest acknowledgement of identifier-resolution limits + SINK-side sanitizer = textbook defense-in-depth.","confidence":0.95},
{"id":"PRAISE-003","severity":"PRAISE","category":"security","file":"scripts/dashboard.ts","title":"Three-layer DNS-rebinding defense","description":"Kernel bind (127.0.0.1) + Host allowlist + per-session bearer cookie. Each layer fails closed independently; test coverage exercises every failure mode.","confidence":0.9},
{"id":"PRAISE-004","severity":"PRAISE","category":"process","file":"grimoires/loa/cycles/cycle-007-agent-debuggability/sprint.md","title":"Multi-phase gate trail with INV traceability","description":"Flatline-PRD → BB-design → Flatline-SDD → Red-Team-SDD → Flatline-sprint, each phase's findings explicitly integrated into next doc with tables. INV matrix maps 18 invariants to enforcing tasks. Surfaced findings (AC-RT-001/003/005) that single-reviewer pass would miss.","confidence":0.95},
{"id":"REFRAME-001","severity":"REFRAME","category":"scope","file":"scripts/dashboard.ts (S8 playground)","title":"Playground may belong to a different cycle","description":"Cycle-007 frame is medium-aware debug substrate. Playground is an iteration surface — different posture, different threat model, different reviewability. If always-scope: ignore. If accretion: name it in COMPLETED.md as 'planned' or 'accreted' so future cycles know which pattern they inherit.","confidence":0.4}
]}
```


## Callouts

The cycle's process discipline is the standout artifact. The five-phase review trail materially improved security posture in ways a single pass cannot — AC-RT-001 (DNS rebinding), AC-RT-003 (ANSI escape), AC-RT-005 (nested-key sanitizer) are all findings that landed because the second and third readers had different threat models. This is how Google's SRE post-incident reviews catch what the on-call missed: each reader brings a fresh frame.

The cookie bootstrap and TS-AST lint upgrade are the load-bearing security wins. Both replace a "good enough today" pattern with one that holds under future regression pressure — and both came from the adversarial-review phases, not the original SDD. Worth distilling as a pattern: when a primitive is security-load-bearing, the adversarial gate often produces the keystone.

Issue #87, opened 2026-05-23
cycle-008 S9 · ship the enriched Components V2 digest to prod + open RLHF loops


zkSoju:
## Context

The cycle-008 S9 RLHF loop converged on a winning Discord digest format: **enriched Components V2** (operator 5/5 — "this one is so good"). The loop ran Discord-native (post → react 1–5 + reply → `rlhf-preview collect` → breed → re-post) across ~6 click-speed rounds, and the production renderer `buildEnrichedDigestComponentsV2(zoneDigest)` (real `ZoneDigest` → enriched layout, injectable handle/pfp) is built + tested in `packages/persona-engine/src/preview/adapters/discord/rich-render.ts`.

This issue tracks **what we did NOT close the loop on** — the controlled production rollout + the threads surfaced during RLHF that weren't tied off. Preference corpus: `grimoires/loa/cycles/cycle-008-persona-substrate/preference-log.jsonl` (rlhf-preference-v1).

---

## 1 · Production rollout of the enriched digest — HIGH (the ship)

The renderer exists; the **delivery wire + canary + flip do not**. Per the operator's own flag-discipline (`LOA_PROMPT_BUILDER`: ship gated → canary → flip; never flip the live community surface mid-stream):

- [ ] Add a `DIGEST_SURFACE` config flag (`'two-beat'` default | `'enriched-v2'`).
- [ ] Wire `buildEnrichedDigestComponentsV2(zoneDigest)` into the cron/composer digest path, gated by the flag (the composer has the `ZoneDigest`; that's the interception point).
- [ ] Components V2 **delivery** across all 3 `deliverZoneDigest` branches (Pattern-B webhook-shell, bot, legacy webhook): POST with `?with_components=true` + `{ flags: IS_COMPONENTS_V2, components }`. (The RLHF tool already does this on the legacy webhook path — `adapters/discord/present.ts`; prod uses Pattern B / `discord.js`.)
- [ ] Extend `DigestPayload` (`deliver/embed.ts`) with optional `components?` / `flags?` to carry it through delivery.
- [ ] Canary: `digest:once` to a test channel with the flag on → mobile screenshot (PP-5 shape) → flip prod.

## 2 · Real-data hookups for `enriched` — HIGH

The renderer maps real `raw_stats`, but two slots are placeholdered:

[...]

PR #88, opened 2026-05-23
cycle-008: persona substrate — RLHF digest loop + enriched-v2 live (gated)

zkSoju:
## What

cycle-008 persona-substrate: the digest voice/visible slice, a Discord-native RLHF
iteration loop, and the **enriched Components V2 digest wired to production behind a flag**.

The thread the operator cared about: ruggy's digest billboard read "too raw from score."
An RLHF loop (post candidates → react 1–5 + reply why → breed → re-post) ran Discord-native
across ~6 click-speed rounds and converged on a **5/5 enriched Components V2** format
(hero + factor movers + spotlight + wallets-warm). This PR ships that format to the live
digest path, gated by `DIGEST_SURFACE` (default `pulse` = unchanged).

## Shippable now (this PR)

- **`DIGEST_SURFACE` flag** (`pulse` default | `enriched-v2`). Zero prod behavior change until flipped.
- **Enriched renderer** relocated to `deliver/enriched-render.ts` (prod core); the RLHF tool in
  `preview/` re-imports it (dependency points inward).
- **`composeDigestPost` enriched branch**: full `ZoneDigest` (`get_zone_digest` → real
  spotlight/wallets/movers) + factor display-name catalog from `get_dimension_breakdown` +
  spotlight wallet→handle via `freeside_auth` (NFR-29: never a raw `0x…` in prose).
- **Components V2 delivery** across all 3 `deliverZoneDigest` branches (Pattern-B webhook,
  bot REST, legacy webhook) via raw REST + `?with_components=true`.
- **Hexagonal RLHF substrate** (`preview/`): core loop / `MediumAdapter` port / discord+terminal
  adapters; the preference corpus + the `billboard-rlhf` composition case study.

Tests: full persona-engine suite **889 pass / 0 fail**, typecheck clean (one pre-existing
`loader.test.ts` error, unrelated). The new code also passed a `/gpt-review` fix-loop (8 findings).

## Rollout

`DIGEST_SURFACE` defaults to `pulse` — merging changes nothing live. Flip to `enriched-v2`
in the Railway prod env after canary.

## Deferred → issue #87

Spotlight **PFP/NFT-square** thumbnail, mibera-holding fallback (storage/sonar/inventory APIs),
`micro` event-triggered LLM voice, canvas/PNG tier, the cycle-009 LLM-as-judge.


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#88. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "nfr9-process-exit-in-builder",
      "title": "process.exit(78) fires deep inside the Effect prompt-builder path",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "packages/persona-engine/src/persona/validate-no-aggregate-stat-leakage.ts",
      "description": "When LOA_STAT_LEAKAGE_GUARD=warn AND isProduction(), the validator calls process.exit(78) from inside an Effect.gen invoked on every cron buildPrompt. This is an abrupt whole-process kill executed mid-Effect — it bypasses Effect's failure-as-value model, terminates the long-running bot serving all characters/zones and any in-flight Discord deliveries, and is effectively untestable (the test suite does not and largely cannot cover this branch). A misconfiguration check that hard-aborts the process belongs at startup/config-load, not per-prompt-build.",
      "suggestion": "Validate LOA_STAT_LEAKAGE_GUARD against the environment once at config load (config.ts) and refuse to boot there, or return Effect.fail(new BuildPromptError(...)) and let the orchestrator decide. Avoid process.exit() inside a referentially-transparent builder.",
      "confidence": 0.6
    },
    {
      "id": "nfr9-priorweekhint-historical-halt",
      "title": "Stat-leakage guard fails-loud on historical priorWeekHint, can halt digests at rollout",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/persona/validate-no-aggregate-stat-leakage.ts",
      "description": "validateNoAggregateStatLeakage runs on the inner body of priorWeekHint for every cron fire (activeFactors is always a non-undefined array for cron, so the `if (args.activeFactors)` guard is truthy even when empty). priorWeekHint is sourced from prior digest voice-memory. Legacy (pre-cycle-008) digests cited aggregates heavily — the PR's own evidence header is 'el-dorado is napping — 575 events drifting through', which matches /\\b\\d{2,}\\s+events?\\b/ → fail-loud → BuildPromptError halts the new digest. Once voice-memory continuity is wired, the first post-migration fire that reads a stat-laden historical hint will halt until the corpus is cleared or warn-mode is set. Currently latent because voice-memory is unwired (preference-log notes 'voice-memory unwired'), which is why severity is MEDIUM not HIGH.",
      "suggestion": "Either scope the guard to operator-controlled inputs (activeFactors) and exclude already-escaped untrusted priorWeekHint bodies, or document a voice-memory purge/migration step before flipping LOA_PROMPT_BUILDER=canonical. Add a regression test that loads a legacy stat-laden hint.",
      "confidence": 0.45
    },
    {
      "id": "nfr9-fragile-fileurl-path",
      "title": "Allowlist path resolved via import.meta.url.replace('file://','')",
      "severity": "LOW",
      "category": "portability",
      "file": "packages/persona-engine/src/persona/validate-no-aggregate-stat-leakage.ts",
      "description": "ALLOWLIST_PATH uses import.meta.url.replace('file://','') to derive a filesystem path. This does not URL-decode (spaces become %20, etc.) and assumes the file:// prefix shape. It happens to work for the current repo path but is fragile across environments and paths with special characters.",
      "suggestion": "Use node:url fileURLToPath(import.meta.url) for a correct, decoded path.",
      "confidence": 0.7
    },
    {
      "id": "nfr9-allowlist-cache-stale",
      "title": "Factor allowlist cached at module scope; runtime refresh needs process restart",
      "severity": "LOW",
      "category": "operability",
      "file": "packages/persona-engine/src/persona/validate-no-aggregate-stat-leakage.ts",
      "description": "allowlistCache is populated once and never invalidated. After an operator runs sync-factor-allowlist to add a legitimately number-bearing factor name (e.g. 'Top-100 Holder'), the long-running bot keeps fail-loud-halting on it until restart.",
      "suggestion": "Document the restart requirement, or stat/mtime-check the file, or expose a cache-reset for the refresh script path.",
      "confidence": 0.6
    },
    {
      "id": "nfr9-single-digit-blindspot",
      "title": "Aggregate-stat regex misses single-digit counts",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/persona/validate-no-aggregate-stat-leakage.ts",
      "description": "The aggregate-count pattern requires \\d{2,} and the spelled-number list starts at 'ten'. Common ruggy phrasings like '3 factors moved' or 'five events' are not detected, so small-magnitude stat leakage passes the guard. This is an under-blocking gap (no false-positive halt), but it weakens the stats-out-of-voice enforcement the cycle is built around.",
      "suggestion": "Lower the digit threshold (\\d+) for aggregate-unit contexts and extend the spelled-number set to one–nine, or accept the gap explicitly with a comment.",
      "confidence": 0.5
    },
    {
      "id": "loader-buildpromptpair-error-misuse",
      "title": "buildPromptPair reply-misuse throws BuildPromptError with wrong kind/argName semantics",
      "severity": "LOW",
      "description": "When postType==='reply' is passed to the cron shim, it throws new BuildPromptError({ kind: 'missing-cron-arg', argName: \"postType='reply' is invalid for the cron shim...\" }). The 'missing-cron-arg' kind does not describe a shim-misuse, and argName is repurposed to carry a prose message. categoryFor() and any downstream classifier will mislabel this.",
      "suggestion": "Add a dedicated kind (e.g. 'invalid-shim-usage') with a `detail` field instead of overloading argName.",
      "confidence": 0.55
    },
    {
      "id": "loader-substitution-no-recurse-divergence",
      "title": "Single-pass substitution no longer recurses into substituted values (behavior change vs chained replace)",
      "severity": "LOW",
      "category": "correctness",
      "description": "substituteWithTracking finds placeholder occurrences only in the original template, so a {{TOKEN}} appearing inside a substituted fragment (e.g. environment/voiceAnchors/priorWeekHint content) is left literal rather than re-substituted as the old chained .replace() would have. The new behavior is safer (resists placeholder injection from semi-trusted content) but it is a divergence that breaks the byte-identical chat-mode regression fence (loader.test scenario 5 is deferred for baseline regen) and can surface a literal {{...}} to the LLM in edge cases.",
      "suggestion": "Document the intentional non-recursive behavior; consider stripping/escaping stray {{...}} tokens from substituted fragment values to avoid literal placeholder leakage.",
      "confidence": 0.4
    },
    {
      "id": "loader-fragment-offsets-untrimmed",
      "title": "fragment_sources offsets computed on untrimmed result then validated against trimEnd'd length",
      "severity": "LOW",
      "category": "correctness",
      "description": "substituteWithTracking records prompt_offset against the untrimmed `result`, but systemHalfRaw = result.trimEnd() and validateFragmentSourcesInvariants is passed templateRegion=[0, systemHalfRaw.length]. If a future template ends a substitution at the very tail (so trimEnd removes characters after a fragment), a fragment's end offset could exceed the trimmed region length and spuriously throw a fragment-sources-invariant violation. Safe today only because the template tail is literal text, not a placeholder.",
      "suggestion": "Compute offsets against the final trimmed/assembled systemPrompt, or trim before tracking, so offsets and templateRegion share the same string.",
      "confidence": 0.3
    },
    {
      "id": "loader-userhalf-not-tracked",
      "title": "fragment_sources covers only systemHalf; cron userHalf substitutions are untracked",
      "severity": "LOW",
      "category": "observability",
      "description": "The cron userHalf still uses chained .replace() for instruction/voiceAnchors/codex/zone substitutions and contributes nothing to fragmentSources. The per-layer source map (FR-15) is therefore partial — the user message portion of the assembled prompt is invisible to trace-explain/dashboard click-to-source. INV-PS-2 frames the invariant around the systemPrompt, so this may be acceptable by design, but it is a coverage gap worth noting.",
      "suggestion": "If userHalf provenance matters for debugging, route the cron userHalf through substituteWithTracking too (offset-based at the userMessage), or document that fragment_sources intentionally maps systemPrompt only.",
      "confidence": 0.4
    },
    {
      "id": "enriched-prettyfactor-acronym-case",
      "title": "prettyFactorName mis-cases acronyms (Lp Provide vs LP)",
      "severity": "LOW",
      "category": "quality",
      "file": "packages/persona-engine/src/deliver/enriched-render.ts",
      "description": "prettyFactorName('onchain:lp_provide') yields 'Lp Provide'. The fallback title-casing breaks acronyms; the test even comments 'catalog wiring = issue'. Mitigated when resolveFactorName (MCP catalog) is supplied, but the prod enriched path can surface the prettify fallback when the resolver is absent or a factor is missing from the catalog.",
      "suggestion": "Add an acronym allowlist (LP, NFT, OG) in prettyFactorName, or ensure resolveFactorName is always wired on the prod enriched path.",
      "confidence": 0.45
    },
    {
      "id": "present-avatar-branch-url",
      "title": "RUGGY_AVATAR_URL hardcodes a raw githubusercontent 'staging' branch path",
      "severity": "LOW",
      "file": "packages/persona-engine/src/preview/adapters/discord/present.ts",
      "description": "The avatar URL points at raw.githubusercontent.com/.../staging/.../avatar.png. A branch rename/delete or repo move silently breaks the webhook avatar. It is also duplicated from character.json::webhookAvatarUrl rather than sourced from it.",
      "suggestion": "Source the avatar from CharacterConfig/character.json, or pin to a tag/commit rather than a moving branch.",
      "confidence": 0.5
    },
    {
      "id": "render-active-factors-empty-diff",
      "title": "render-active-factors.ts shows +0/-0 despite being imported by loader.ts",
      "severity": "SPECULATION",
      "category": "build",
      "file": "packages/persona-engine/src/persona/render-active-factors.ts",
      "description": "The diff lists render-active-factors.ts as added with +0/-0 and an empty body, yet loader.ts and validate-no-aggregate-stat-leakage.ts import renderActiveFactors/ActiveFactorRender from it. If literally empty, the package would fail to compile and every test referencing buildPrompt would fail — so this is almost certainly a diff-rendering/truncation artifact rather than a real empty file. Flagged only so a reviewer confirms the file has content.",
      "suggestion": "Confirm render-active-factors.ts actually contains renderActiveFactors + ActiveFactorRender; if the empty diff is real, restore the implementation.",
      "confidence": 0.15
    },
    {
      "id": "praise-fragment-offset-tracking",
      "title": "Single-pass cursor-tracking substitution for fragment_sources",
      "severity": "PRAISE",
      "category": "design",
      "description": "substituteWithTracking walks the template once, records each fragment's offset in the FINAL prompt, handles multiple occurrences per placeholder, and skips zero-length substitutions to satisfy the no-overlap/ascending invariant. This deliberately avoids the brittle post-mutation string-matching that the Flatline note (SKP-004) warned against, and keeps offsets honest for the click-to-source trace UX.",
      "suggestion": "Keep this approach as the canonical substitution path.",
      "confidence": 0.85
    },
    {
      "id": "praise-capture-rest-hardening",
      "title": "Capture path hardened against rate-limits and silent auth failures",
      "severity": "PRAISE",
      "category": "reliability",
      "file": "packages/persona-engine/src/preview/adapters/discord/capture.ts",
      "description": "The 429-safe apiGetJson helper retries with retry_after, surfaces 401/403 as thrown errors (instead of recording empty feedback that would poison the RLHF corpus), and the reply scan paginates while tracking unresolved per-candidate targets so a second candidate's 'why' on an older page is not missed. This directly and thoroughly addresses the GPT review's critical/major capture findings, with tests proving each path.",
      "suggestion": "None.",
      "confidence": 0.85
    },
    {
      "id": "praise-present-surface-fallback",
      "title": "Per-candidate data-beat fallback keeps the batch resilient",
      "severity": "PRAISE",
      "category": "reliability",
      "file": "packages/persona-engine/src/preview/adapters/discord/present.ts",
      "description": "The data-beat post is wrapped in try/catch so a surface the webhook rejects (e.g. a Components V2 quirk) falls back to plain text rather than aborting the whole fan-out round. URL params are also composed via new URL()/searchParams.set() so existing webhook query params (thread_id) survive, and reaction retries are bounded iteratively rather than recursively.",
      "suggestion": "None.",
      "confidence": 0.8
    },
    {
      "id": "praise-window-derivation-and-zero-omit",
      "title": "Enriched renderer derives window from digest bounds and omits empty footers",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "packages/persona-engine/src/deliver/enriched-render.ts",
      "description": "buildEnrichedDigestComponentsV2 derives the displayed window from window_start/window_end with a validated 7-day fallback for malformed bounds (closing the GPT-flagged hardcoded 'last 7 days'), omits the 'wallets warm' footer when count is 0 (never surfaces '0 wallets warm' from the pulse path), and strips em-dashes — all behavior-pinned by tests.",
      "suggestion": "None.",
      "confidence": 0.8
    }
  ]
}
```


## Callouts

zkSoju (commented):
## Verdict: `REQUEST_CHANGES`

The two MEDIUM reliability findings (`nfr9-process-exit-daemon-killswitch`, `nfr9-empty-allowlist-fail-loud`) are pre-flip blockers — they become live the moment `LOA_PROMPT_BUILDER` flips to `canonical`. Everything else is `COMMENT`-grade and shippable.

## Summary

This is a large, disciplined cycle: a hexagonal RLHF preview surface, an Effect-TS prompt-builder migration with offset-accurate fragment tracing, and a gated enriched-v2 digest path. The architecture is genuinely strong — the ports/adapters split and the substitution-with-tracking work are reference-quality. The risk concentrates in the new stat-leakage guard, whose two failure modes (process-killing warn mode + an empty allowlist under fail-loud) are both too blunt for a long-lived daemon, and in a copy-paste of the enriched renderer that nothing keeps in sync.

## Architectural Meditation

The through-line of this PR is *boundaries done well, failure modes done bluntly*. The preview surface earns its hexagon, and the prompt builder records provenance at exactly the right instant. But the same migration introduces a validator that reaches past its station and calls `process.exit` from inside a hot path — a library seizing an application-level privilege. The most teachable tension here is that **influence and authority are different scopes**: a validator deep in `buildPrompt` should *signal* (`Effect.fail`) and let the orchestrator decide whether one bad fire drops a post or the boot sequence aborts. The daemon must outlive its inputs.

A second pattern repeats three times — DRY for *resilience and rendering*. The 429-retry hardening landed in `present.ts` but not the prod CV2 webhook path; the enriched renderer was copied rather than shared. The fix for both is the same move: one source of truth, re-exported. This PR already proves it knows how (the hexagonal core), so the inconsistency is a finishing-pass issue, not a design flaw.

## Findings


```json
{
  "schema_version": 1,
  "verdict": "REQUEST_CHANGES",
  "findings": [
    {
      "id": "nfr9-process-exit-daemon-killswitch",
      "title": "process.exit(78) inside buildPrompt validation can crash-loop the cron daemon",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "packages/persona-engine/src/persona/validate-no-aggregate-stat-leakage.ts",
      "description": "When LOA_STAT_LEAKAGE_GUARD=warn is set in a production environment, validateNoAggregateStatLeakage calls process.exit(78). This function runs inside buildPrompt's Effect.gen on every cron voice fire. In the long-lived bot daemon, a single misconfigured env var would terminate the entire process (all three cron cadences + interactions endpoint), not just the current digest. On a supervised host the process restarts, calls buildPrompt again, and exits again — a crash loop. A library-level validator should never unilaterally kill the host process.",
      "suggestion": "Replace process.exit(78) with Effect.fail(new BuildPromptError({kind:'config-error', ...})) (or a thrown ConfigError caught by the cron orchestrator), letting the single fire fail loudly and log while the daemon stays up. Reserve EX_CONFIG/78 for the CLI entrypoint that can assert the env once at boot.",
      "confidence": 0.6,
      "faang_parallel": "This is the anti-pattern Java's SecurityManager historically existed to contain: a library calling System.exit() inside a server. Netflix's Hystrix codified the inverse discipline — bulkhead isolation fails the request, never the JVM. Erlang/OTP's 'let it crash' is often misquoted here: the unit of crash is the supervised actor, not the whole node — this code crashes the node.",
      "metaphor": "A smoke detector that, instead of beeping, cuts power to the entire building — including the other smoke detectors and the elevator.",
      "teachable_moment": "Libraries influence; applications decide. process.exit is an application-boundary privilege. A validator in a hot path should signal upward and let the caller choose between 'drop this one fire' and 'abort boot' — it should never make that choice for the whole process.",
      "connection": "This contradicts Loa's own EX_CONFIG/78 convention, which reserves exit-78 for the CLI entrypoint that asserts config once at boot. The PR re-implements that convention in the wrong layer."
    },
    {
      "id": "nfr9-empty-allowlist-fail-loud",
      "title": "Fail-loud stat-leakage guard ships with an empty allowlist — a benign factor name can halt digests",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "packages/persona-engine/src/persona/fixtures/factor-name-allowlist.json",
      "description": "allowlisted_names is [] and NFR-9 defaults to fail-loud. The window-language pattern (/\\b\\d+\\s*-?\\s*day\\b/) and aggregate-count pattern would match legitimate score-mcp factor display names such as 'Top-100 Holder' or '30 Day Streak' (the exact examples the fixture comment cites). If score-mcp introduces such a name once LOA_PROMPT_BUILDER flips to 'canonical', every digest fire for that zone returns Effect.fail and the post is suppressed — a silent content outage until the operator runs sync-factor-allowlist. The warn-mode escape hatch is the only mitigation and is itself dangerous in prod (see nfr9-process-exit-daemon-killswitch).",
      "suggestion": "Seed factor-name-allowlist.json from the live score-mcp factor catalog before flipping LOA_PROMPT_BUILDER to canonical (the sync-factor-allowlist step), and/or have the fail path emit a single trace row + fall back to the legacy builder for that fire rather than dropping the post entirely.",
      "confidence": 0.55,
      "faang_parallel": "Cloudflare's July 2019 global outage was a regex (catastrophic backtracking) deployed into enforcement without a staged rollout; the postmortem lesson was stage-then-enforce. AWS IAM is safe at default-deny only because policies are provisioned before traffic arrives — an empty allowlist with fail-loud inverts that ordering.",
      "metaphor": "Installing a spam filter with an empty 'trusted senders' list and setting it to delete-on-match — the first legitimate newsletter silently vanishes.",
      "teachable_moment": "A fail-loud guard is only as safe as its allowlist is complete at the moment of the flip. Enforcement logic and the data it enforces against must ship together; an empty allowlist plus fail-loud is a guaranteed outage waiting for the first matching input.",
      "connection": "Couples with the kill-switch finding: the warn-mode that would soften this outage is itself the process.exit footgun. Two findings, one root — both of the guard's failure modes are too blunt for production."
    },
    {
      "id": "duplicate-enriched-renderer",
      "title": "Two near-identical enriched-digest Components-V2 renderers (preview vs deliver) risk drift",
      "severity": "MEDIUM",
      "file": "packages/persona-engine/src/deliver/enriched-render.ts",
      "description": "buildEnrichedDigestComponentsV2 / IS_COMPONENTS_V2 / prettyFactorName exist in BOTH packages/persona-engine/src/deliver/enriched-render.ts (prod path, consumed by digest-orchestrator) AND packages/persona-engine/src/preview/adapters/discord/rich-render.ts (RLHF preview path). The two test files (enriched-render.test.ts and rich-render.test.ts) assert overlapping behavior (# hero, member-noun, em-dash strip, window-from-bounds, factor prettify). A fix landed in one (e.g. the rank-climb '#prior → #current' or 'an anonymous mibera' fallback) must be hand-mirrored in the other; nothing enforces parity. This contradicts the repo's UNIX-single-discipline construct boundary guidance.",
      "suggestion": "Extract the renderer into one shared module (e.g. deliver/enriched-render.ts as source of truth) and have preview/rich-render re-export it, or delete the preview copy once the prototype has served its purpose. If they must diverge, add a comment + a shared snapshot fixture both import to catch drift.",
      "confidence": 0.65,
      "faang_parallel": "Knight Capital's 2012 $440M loss was, at root, an old code path and a new code path diverging behind a reused flag (Power Peg). Google's monorepo enforces single-source-of-truth precisely because divergent twins are how a fix lands in one surface and silently rots in the other — and the recent commits here (rank-change, 'an anonymous mibera') are exactly the kind of fix that must now be applied twice.",
      "metaphor": "Two identical recipe cards in two kitchens — chef A fixes the salt, chef B never hears, and table B keeps getting last week's dish.",
      "teachable_moment": "Copy-paste creates a maintenance contract no one signed. The durable choices are unify-behind-one-module or bind-both-with-a-shared-snapshot so drift fails a test, not a user. A comment saying 'keep these in sync' is the weakest of the three.",
      "connection": "Directly contradicts the repo's stated UNIX single-discipline construct boundary. The prod (deliver) renderer should be canonical and the preview path should re-export it."
    },
    {
      "id": "ruggy-avatar-hardcoded-staging-branch",
      "title": "RUGGY_AVATAR_URL hardcodes a raw.githubusercontent 'staging' branch path instead of reading character.json",
      "severity": "LOW",
      "file": "packages/persona-engine/src/preview/adapters/discord/present.ts",
      "description": "RUGGY_AVATAR_URL is a literal raw.githubusercontent.com/.../staging/apps/character-ruggy/avatar.png. The comment even says it is the canonical character.json::webhookAvatarUrl value, yet it is copied as a string rather than read from config. It is also reused by post-type-gallery.ts and the rich-render thumbnail accessory. A branch rename, repo move, or avatar relocation silently breaks every preview/gallery avatar, and it pins the surface to a specific git branch.",
      "suggestion": "Resolve the avatar from the CharacterConfig (webhookAvatarUrl) passed into the adapter, falling back to the literal only when config is absent.",
      "confidence": 0.6,
      "faang_parallel": "12-factor config (Heroku) — config lives in the environment, not transcribed into code. A URL pinned to a branch is deploy-coupling smell of exactly the kind 12-factor was written to eliminate.",
      "metaphor": "Hand-printing your home address on every business card — move house and every card now lies.",
      "teachable_moment": "When a value already exists in a config object, read it once; don't transcribe it. The comment admitting 'this is the canonical character.json value' is the tell that the wrong source was chosen.",
      "connection": "This repo is accessed from a dual-worktree topology and references a 'staging' branch — a branch-pinned literal is unusually brittle in exactly this environment."
    },
    {
      "id": "allowlist-path-import-meta-url-parsing",
      "title": "Allowlist path derived via import.meta.url.replace('file://','') is fragile",
      "severity": "LOW",
      "category": "portability",
      "file": "packages/persona-engine/src/persona/validate-no-aggregate-stat-leakage.ts",
      "description": "ALLOWLIST_PATH is computed with dirname(import.meta.url.replace('file://','')). This does not URL-decode the path (a repo path containing spaces or other percent-encoded characters would resolve incorrectly) and is platform-fragile compared to the canonical conversion. Works for the current /Users/zksoju paths but is a latent footgun (the project is also accessed from /Users/zksoju/bonfire/ per the dual-worktree topology).",
      "suggestion": "Use Bun's import.meta.dir, or node:url fileURLToPath(import.meta.url), to obtain the directory robustly.",
      "confidence": 0.5,
      "faang_parallel": "Node core added fileURLToPath precisely because naive file:// string-stripping mishandles URL-encoding; spaces-in-path is a perennial cross-platform bug class (the 'Program Files' problem on Windows, the encoded-character problem everywhere).",
      "metaphor": "Translating an address by snipping off the 'http://' with scissors — fine until the address contains a space or an accent.",
      "teachable_moment": "URLs are not strings; decode them with the URL machinery. Bun's import.meta.dir or fileURLToPath remove the guesswork and the percent-encoding hazard in one call.",
      "connection": "The same dual-worktree topology that makes the avatar pin brittle raises the odds that an encoded path character eventually appears under one of the roots."
    },
    {
      "id": "cv2-prod-webhook-no-429-retry",
      "title": "Production Components-V2 webhook delivery has no 429 handling, unlike the RLHF present path",
      "severity": "LOW",
      "category": "reliability",
      "file": "packages/persona-engine/src/deliver/webhook.ts",
      "description": "The new components-v2 branch in sendViaWebhook bypasses discord.js (raw fetch to webhook.url with ?with_components=true) and throws on any non-ok response, including 429. The RLHF present.ts path got bounded retry_after handling during the GPT-review hardening, but the actual production digest delivery path did not. A rate-limited enriched digest send will throw rather than retry, dropping the scheduled post.",
      "suggestion": "Factor the bounded 429-retry helper used in present.ts into a shared utility and apply it to the prod CV2 delivery branch, or route CV2 through the same retry wrapper.",
      "confidence": 0.5,
      "faang_parallel": "Stripe and Google API clients treat 429 + Retry-After as a first-class retry signal; the AWS SDK ships exponential backoff with jitter by default. Discord explicitly hands you retry_after — throwing it away is leaving a documented affordance on the table.",
      "metaphor": "The test kitchen has a 'wait and retry' rule for the busy oven, but the dinner-service line throws the plate on the floor the moment the oven is full.",
      "teachable_moment": "Resilience should land on the production path first, not just the preview path. Asymmetric hardening — preview retries, prod doesn't — is the inverse of where you want your robustness.",
      "connection": "present.ts already has the bounded retry; this is the same DRY move as the duplicate-renderer finding, applied to resilience logic instead of rendering logic."
    },
    {
      "id": "chat-mode-substrate-block-injected",
      "title": "Chat-mode (reply) prompts now render a cron-shaped SUBSTRATE STATE block; byte-identical fence deferred",
      "severity": "LOW",
      "category": "correctness",
      "description": "buildPrompt now substitutes {{ACTIVE_FACTORS}} and {{PRIOR_WEEK_HINT}} unconditionally, including the reply shape. Because the placeholders live in the persona.md base template (universal), chat-mode replies now emit a '═══ SUBSTRATE STATE (this week) ═══' block (empty / '(none)') even though cycle-008 scoped substrate-state to cron. This is an improvement over the prior behavior (the pre-S2 baseline fixture shows raw '{{ACTIVE_FACTORS}}' leaking literally to chat), but the chat-mode equivalence scenarios 5 and 10 are explicitly DEFERRED, so the new chat output is not regression-tested in this PR — the migration's primary safety net for the shared path is unexercised.",
      "suggestion": "Either gate the SUBSTRATE STATE block on shape.kind==='cron' (skip the section for reply) or land the deferred chat-mode fixture comparison now so the changed chat prompt is pinned. At minimum, confirm renderActiveFactors(undefined) does not surface a misleading 'no activity' block in a chat context that has live MCP tools.",
      "confidence": 0.45,
      "faang_parallel": "The expand/contract (parallel-change) migration pattern requires both code paths under test before and after the flip; Google's 'live migrations' practice treats the unmigrated path as the one most likely to regress, precisely because attention drifts to the new one.",
      "metaphor": "Renovating a shared hallway, testing only the door to room A, and shipping without checking that room B opens onto the same hall.",
      "teachable_moment": "When two callers share a template, a change 'scoped' to one caller is only truly scoped if the template branches. A universal placeholder is, by definition, universal — the chat path inherited the change whether or not it was intended to.",
      "connection": "It's a real improvement over literal {{ACTIVE_FACTORS}} leakage, but deferring scenarios 5 & 10 means the one regression net guarding the shared path is the missing one."
    },
    {
      "id": "capture-operator-attribution-loose",
      "title": "Reaction/reply attribution counts any non-bot user when operatorUserId is unset",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/preview/adapters/discord/capture.ts",
      "description": "isOperator returns true for any non-bot user when config.operatorUserId is absent. If the test channel has any other human present, their reactions/replies are silently folded into the operator's preference record, biasing the rlhf-preference corpus. Additionally, if /users/@me fails (returns fallback null), botUserId is undefined and the bot's own seeded 1-5 reactions are counted as operator scores, producing a spurious score=5 for every candidate.",
      "suggestion": "Make operatorUserId required for collect (or warn loudly when unset), and skip reaction scoring entirely when botUserId could not be resolved rather than counting the bot's own seeds.",
      "confidence": 0.45,
      "faang_parallel": "Reward-model data hygiene is the whole game in RLHF — unverified annotators poison the preference set, which is why every serious labeling pipeline (Scale, Google's RLHF data ops) gates on annotator identity rather than on its absence.",
      "metaphor": "A taste-test where, if you forget to check badges, anyone in the food court — including the chef tasting his own soup — gets counted as a judge.",
      "teachable_moment": "Defaulting an identity check to 'true when unconfigured' inverts safe-by-default. For an RLHF corpus, a spurious score=5 from the bot's own seed reaction is strictly worse than a dropped datapoint.",
      "connection": "This intake feeds the same preference-log that the praised hexagonal core records into — a clean pipe with a contaminated source."
    },
    {
      "id": "ledger-sprint-status-vs-shipped-code",
      "title": "Ledger marks S2–S8 'planned' while their code is implemented in this PR",
      "severity": "LOW",
      "category": "process",
      "file": "grimoires/loa/ledger.json",
      "description": "ledger.json lists cycle-008 sprints S2 (buildPrompt Effect-TS), S3 (cron migration), S4 (trace/fragment_sources) etc. with status 'planned', yet loader.ts (S2), claude-sdk.live.ts migration (S3), and fragment_sources population (S4) are clearly landed in this diff, and the commits are tagged sprint-30 (S9). The sprint lifecycle in the ledger no longer reflects implementation reality, which undermines cross-session recovery and the beads-as-source-of-truth invariant.",
      "suggestion": "Reconcile sprint statuses (mark completed sprints completed, or clarify that S9 shipped ahead of S2–S8) before cycle close so the ledger and beads epics match what is actually merged.",
      "confidence": 0.5,
      "faang_parallel": "Google's Critique/Piper bind review state to actual change state for exactly this reason; a tracker that disagrees with merged code is the bug-tracker-rot failure mode where people stop trusting, then stop reading, the board.",
      "metaphor": "A flight board showing 'boarding' for planes that already landed — once it's wrong twice, nobody reads it again.",
      "teachable_moment": "When status metadata and merged code disagree, cross-session recovery degrades silently — the next agent trusts a 'planned' label and re-does landed work. Reconcile before cycle close.",
      "connection": "The repo's own beads-first invariant is only as strong as the ledger's honesty about what shipped."
    },
    {
      "id": "rlhf-preview-prod-channel-guard-advisory-only",
      "title": "rlhf-preview relies on operator discipline (not a check) to avoid posting candidate spam to a prod channel",
      "severity": "LOW",
      "category": "safety",
      "file": "apps/bot/src/cli/rlhf-preview.ts",
      "description": "requireWebhook validates the URL is well-formed and prints 'NEVER a prod community channel', but performs no programmatic guard. Given the anti-spam invariant (characters never post unsolicited) and the documented .env channel-world mismatch risk, an operator pointing --webhook (or RLHF_TEST_WEBHOOK_URL) at a production channel would fan out N candidate billboards + 1-5 reactions each into a live community surface.",
      "suggestion": "Add a soft confirmation or an allowlist/denylist check (e.g. refuse known prod webhook IDs, or require an explicit --i-understand-test-only flag) before posting.",
      "confidence": 0.35,
      "faang_parallel": "AWS makes you type the bucket/resource name to confirm a destructive delete because operator discipline is not a control. Prod-vs-test routing guards exist for the same reason — the human is the least reliable part of the loop at 2am.",
      "metaphor": "A 'do not point at people' label on a nail gun — useful documentation, but not a safety.",
      "teachable_moment": "A printed warning is documentation, not enforcement. Given the documented .env channel-world mismatch risk in this very repo, even a soft confirmation or a denylist turns a known foot-gun into a caught mistake.",
      "connection": "A misrouted preview fan-out would violate the load-bearing anti-spam invariant (characters never post unsolicited) — via tooling rather than via the character loop, but the community sees the same spam."
    },
    {
      "id": "praise-hexagonal-ports-adapters",
      "title": "Clean hexagonal split: medium-blind core + discord/terminal adapters behind one MediumAdapter port",
      "severity": "PRAISE",
      "category": "architecture",
      "file": "packages/persona-engine/src/preview/index.ts",
      "description": "The preview surface cleanly separates the generate→record→breed core from the MediumAdapter boundary, with discord (primary) and terminal (dev-aid) adapters proving the hexagon — loop.test.ts verifies the core is medium-blind via a stub, and the terminal adapter rides the same loop with zero core changes. This is a well-disciplined ports-and-adapters design that makes adding a future medium a pure edge addition.",
      "suggestion": "Keep the core free of Discord-specific types as new mediums are added.",
      "confidence": 0.85,
      "faang_parallel": "This is Cockburn's hexagonal architecture executed the way Netflix and Amazon teams test business logic without I/O — drive the core through a stub port. The second (terminal) adapter is the empirical proof the port is real, not aspirational.",
      "metaphor": "A power tool with swappable heads — the motor never changes when you snap on a new bit.",
      "teachable_moment": "The proof that a port is genuinely a port is a second adapter that costs nothing in the core. 'Terminal rides the same loop with zero core changes' is that proof, stated as a test.",
      "connection": "Mirrors the substrate-vs-character split this repo prizes — medium/voice at the edge, logic in the middle."
    },
    {
      "id": "praise-offset-tracking-substitution",
      "title": "Offset-tracking single-pass substitution avoids brittle post-mutation string matching",
      "severity": "PRAISE",
      "category": "correctness",
      "description": "substituteWithTracking replaces the chained .replace() approach with one pass that records each fragment's offset in the FINAL prompt, sorts occurrences, and skips zero-length substitutions so the FR-15a invariants (ascending, non-overlapping) hold. This deliberately sidesteps the false-attribution failure mode of matching strings after the buffer has been mutated, and produces accurate fragment_sources for the trace surface.",
      "suggestion": "None — consider documenting the half-open [start,end) offset convention near the FragmentSource type for future readers.",
      "confidence": 0.8,
      "faang_parallel": "Facebook's jscodeshift and every source-map implementation track offsets during the transform, not after, for exactly this reason — mutation invalidates positions, so provenance recorded post-hoc is fiction.",
      "metaphor": "Numbering the boxes as you load the truck, not guessing where they ended up after everything shifted in transit.",
      "teachable_moment": "Record provenance during the transform, not after. Post-mutation string matching is the textbook false-attribution bug; capturing offsets in the single pass eliminates the class.",
      "connection": "It earns trustworthy fragment_sources for the trace surface — observability captured at the only moment it can be correct."
    },
    {
      "id": "praise-iterative-review-hardening",
      "title": "Discord capture/present hardened iteratively against multi-model review (429 retry, pagination, fail-fast)",
      "severity": "PRAISE",
      "category": "reliability",
      "file": "packages/persona-engine/src/preview/adapters/discord/capture.ts",
      "description": "The capture/present paths visibly absorbed the three GPT-review iterations: a shared 429-safe apiGetJson with bounded retry, message-history pagination that only stops once every candidate target is resolved (so a second candidate's why on an older page isn't dropped), 401/403 surfaced as errors rather than recorded as empty feedback, and fail-fast when botToken is missing. The test suite encodes each of these as explicit regression cases (rate-limited retry, cross-candidate pagination, auth-failure surfacing).",
      "suggestion": "None.",
      "confidence": 0.8,
      "faang_parallel": "This is Google's postmortem discipline in miniature — 'every outage gets a test.' Each review finding became a permanent regression case rather than a fixed-and-forgotten patch.",
      "metaphor": "Each near-miss on the mountain road becomes a permanent guardrail, not just a story told at dinner.",
      "teachable_moment": "Multi-model review is only as durable as the regression tests that pin its findings. Converting findings into tests is the lasting win; the fix alone would erode on the next refactor.",
      "connection": "This is the counter-example to cv2-prod-webhook-no-429-retry — the same 429 hardening landed here but not on the prod webhook path. Finishing that extraction would close the gap."
    }
  ]
}
```


## Callouts

- **[Architecture] The hexagon is real, not aspirational.** The `terminal` adapter riding the same `generate→record→breed` loop with zero core changes is the empirical proof — `loop.test.ts` drives the core through a stub, so "medium-blind" is a tested invariant, not a comment. This is the cleanest part of the PR.
- **[Correctness] Provenance captured at the right instant.** `substituteWithTracking` records offsets *during* the single substitution pass instead of string-matching a mutated buffer afterward. That's the difference between accurate `fragment_sources` and plausible fiction — the same reason source maps exist.
- **[Reliability] Review findings became regression tests.** The capture/present paths converted three rounds of GPT review (429 retry, cross-candidate pagination, auth-failure surfacing, missing-token fail-fast) into explicit test cases. That's the durable form of hardening — and it makes the *missing* prod-path 429 retry (`webhook.ts`) stand out as the one place the discipline didn't reach.

**One closing note on shape:** the three strongest patterns in this PR (one source of truth, signal-don't-seize, test-the-shared-path) are also the three roots of every finding above. The architecture already demonstrates it knows all three — the asks are about applying them consistently to the guard, the second renderer, and the prod webhook before the canonical flag flips.

PR #92, opened 2026-05-24
fix(persona-engine): ambient stir survives unconfigured freeside-auth MCP

zkSoju:
## what broke

`ruggy` (and all 8 characters) logged this **hourly across all 4 zones** — stonehenge · bear-cave · el-dorado · owsley-lab:

```
ambient-runtime: AMBIENT STIR TIER DISABLED in production — 1 endpoint(s) misconfigured.
ambient-stir: zone=... error=ambient-stir disabled at boot: freeside-auth: freeside-auth-mcp URL not configured
```

The bot was otherwise healthy (digest cron, read-side, Discord, score-mcp all LIVE). Only the **ambient-stir tier** was down — and it disabled *itself* at boot.

## root cause

`runtime.ts` boot validation treated all three ambient MCP endpoints (`score`, `codex`, `freeside-auth`) as **equally tier-fatal**. So an unset `FREESIDE_AUTH_MCP_URL` — the *intended* state until `identity-api` ships its `resolve_wallet` MCP — took the whole tier down.

But `freeside-auth` is **soft-degradable enrichment**: `WalletResolverLive` returns `ANONYMOUS_KEEPER` on any failure/timeout/circuit-break and never leaks a raw `0x…` (NFR-29). Its absence should degrade narration (anonymized handles), not kill the tier.

## fix

Classify endpoints by criticality (new `endpoint-criticality.ts`, pure + unit-tested):

| endpoint | class | on failure |
|---|---|---|
| `score` | **required** | disable tier (event source — no events, no stir) |
| `codex` | **required** | disable tier (`MiberaResolverLive` still propagates errors — not yet fully fail-soft) |
| `freeside-auth` | **optional** | warn only — tier stays up, narration anonymizes |

The `FREESIDE_AUTH_MCP_URL` seam is untouched — when `identity-api` deploys its MCP and the var is set, real handle resolution lights up with **zero further change**. Consumer-side mirror of the identity-api PRD **G-5** graceful-degradation doctrine.

## verification

Real prod-`NODE_ENV` boot simulation (not just units):

- `freeside-auth` unset → tier **ENABLED**, `disabled=false` ✅ (the fix)
- `codex` unset → tier **DISABLED**, `reasons=[codex]` ✅ (required preserved; `freeside-auth` correctly demoted to a warning even when also unset)
- `tsc --noEmit` clean · **23/23** ambient tests pass (9 new)

## files

```
endpoint-criticality.ts        +103   new · policy + rationale
endpoint-criticality.test.ts   +126   new · 9 tests, pins exact prod symptom string
runtime.ts                     ±49    classify required vs optional
```

## ⚠️ deploy note

Live bot banner reads `v0.6.0-A` but `main` is `v0.12.0`. If that's a real version (not a persona display string), merging to `main` won't reach prod without deploying current `main` or backporting onto the release line. Flagging — deploy is operator domain.

### DIFF
--- a/packages/persona-engine/src/ambient/endpoint-criticality.test.ts
+++ b/packages/persona-engine/src/ambient/endpoint-criticality.test.ts
@@ -1,0 +1,126 @@
+// Regression gate for the ambient-stir boot-disable bug: an unset
+// FREESIDE_AUTH_MCP_URL (the intended state until identity-api ships its
+// resolve_wallet MCP) must NOT disable the entire stir tier. Before this fix,
+// boot validation treated freeside-auth as equally tier-fatal to the event
+// source, so production logged "AMBIENT STIR TIER DISABLED ... freeside-auth:
+// freeside-auth-mcp URL not configured" hourly across all zones.
+//
+// Pure-policy tests — no env, no I/O, no effect runtime — so they run under
+// `bun test` without installed deps.
+
+import { describe, expect, test } from "bun:test";
+import {
+  AMBIENT_ALL_ENDPOINTS,
+  AMBIENT_OPTIONAL_ENDPOINTS,
+  AMBIENT_REQUIRED_ENDPOINTS,
+  classifyEndpointCriticality,
+  isOptionalAmbientEndpoint,
+  type EndpointValidationResult,
+} from "./endpoint-criticality.ts";
+
+// The exact reason validateEndpointConfig("freeside-auth", …) returns when the
+// URL is unset (score-mcp-client.ts) — what production actually logged.
+const FREESIDE_AUTH_UNCONFIGURED = "freeside-auth-mcp URL not configured";
+
+describe("ambient endpoint criticality — policy membership", () => {
+  test("score and codex are required; freeside-auth is optional", () => {
+    expect(AMBIENT_REQUIRED_ENDPOINTS).toEqual(["score", "codex"]);
+    expect(AMBIENT_OPTIONAL_ENDPOINTS).toEqual(["freeside-auth"]);
+  });
+
+  test("AMBIENT_ALL_ENDPOINTS lists every endpoint, required first", () => {
+    expect(AMBIENT_ALL_ENDPOINTS).toEqual(["score", "codex", "freeside-auth"]);
+  });
+
+  test("isOptionalAmbientEndpoint reflects the policy", () => {
+    expect(isOptionalAmbientEndpoint("freeside-auth")).toBe(true);
+    expect(isOptionalAmbientEndpoint("score")).toBe(false);
+    expect(isOptionalAmbientEndpoint("codex")).toBe(false);
+  });
+});
+
+describe("classifyEndpointCriticality — the production regression", () => {
+  test("unset FREESIDE_AUTH_MCP_URL alone does NOT disable the tier", () => {
+    const results: EndpointValidationResult[] = [
+      { endpoint: "score", ok: true },
+      { endpoint: "codex", ok: true },
+      {
+        endpoint: "freeside-auth",
+        ok: false,
+        reason: FREESIDE_AUTH_UNCONFIGURED,
+      },
+    ];
+
+    const outcome = classifyEndpointCriticality(results);
+
+    expect(outcome.disabled).toBe(false);
+    expect(outcome.reasons).toEqual([]);
+    expect(outcome.warnings).toEqual([
+      `freeside-auth: ${FREESIDE_AUTH_UNCONFIGURED}`,
+    ]);
+  });
+});
+
+describe("classifyEndpointCriticality — required endpoints stay tier-fatal", () => {
+  test("a failed score endpoint disables the tier", () => {
+    const outcome = classifyEndpointCriticality([
+      { endpoint: "score", ok: false, reason: "score-mcp URL not configured" },
+      { endpoint: "codex", ok: true },
+      { endpoint: "freeside-auth", ok: true },
+    ]);
+
+    expect(outcome.disabled).toBe(true);
+    expect(outcome.reasons).toEqual(["score: score-mcp URL not configured"]);
+    expect(outcome.warnings).toEqual([]);
+  });
+
+  test("a failed codex endpoint disables the tier", () => {
+    const outcome = classifyEndpointCriticality([
+      { endpoint: "score", ok: true },
+      { endpoint: "codex", ok: false, reason: "codex-mcp URL not configured" },
+      { endpoint: "freeside-auth", ok: true },
+    ]);
+
+    expect(outcome.disabled).toBe(true);
+    expect(outcome.reasons).toEqual(["codex: codex-mcp URL not configured"]);
+    expect(outcome.warnings).toEqual([]);
+  });
+});
+
+describe("classifyEndpointCriticality — mixed and edge cases", () => {
+  test("a required failure disables even alongside an optional failure", () => {
+    const outcome = classifyEndpointCriticality([
+      { endpoint: "score", ok: false, reason: "score-mcp must use https://" },
+      { endpoint: "codex", ok: true },
+      {
+        endpoint: "freeside-auth",
+        ok: false,
+        reason: FREESIDE_AUTH_UNCONFIGURED,
+      },
+    ]);
+
+    expect(outcome.disabled).toBe(true);
+    expect(outcome.reasons).toEqual(["score: score-mcp must use https://"]);
+    expect(outcome.warnings).toEqual([
+      `freeside-auth: ${FREESIDE_AUTH_UNCONFIGURED}`,
+    ]);
+  });
+
+  test("all endpoints healthy — not disabled, no reasons, no warnings", () => {
+    const outcome = classifyEndpointCriticality([
+      { endpoint: "score", ok: true },
+      { endpoint: "codex", ok: true },
+      { endpoint: "freeside-auth", ok: true },
+    ]);
+
+    expect(outcome).toEqual({ disabled: false, reasons: [], warnings: [] });
+  });
+
+  test("missing reason falls back to a generic detail", () => {
+    const outcome = classifyEndpointCriticality([
+      { endpoint: "score", ok: false },
+    ]);
+
+    expect(outcome.reasons).toEqual(["score: validation failed"]);
+  });
+});
--- a/packages/persona-engine/src/ambient/endpoint-criticality.test.ts
+++ b/packages/persona-engine/src/ambient/endpoint-criticality.test.ts
@@ -9,6 +9,7 @@
 // `bun test` without installed deps.
 
 import { describe, expect, test } from "bun:test";
+import type { AmbientMcpEndpoint } from "./live/score-mcp-client.ts";
 import {
   AMBIENT_ALL_ENDPOINTS,
   AMBIENT_OPTIONAL_ENDPOINTS,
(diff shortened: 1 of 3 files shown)


zkSoju (commented):
## Summary

This PR lands a clean resilience fix: the ambient stir no longer collapses when `FREESIDE_AUTH_MCP_URL` is unset. The architectural decision to extract endpoint criticality into a pure, side-effect-free module is the right call — it cleanly separates *policy* (what the system knows about endpoint importance) from *runtime behavior* (what the system does when policy is violated). The most important finding before merge is F-001: the TypeScript type system provides no guarantee that the policy arrays stay in sync with the `AmbientMcpEndpoint` union, and that gap widens silently as the schema evolves. F-007 is the most architecturally interesting observation: it identifies an accidental coupling wearing the costume of intentional policy.

---

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "No exhaustiveness guarantee between AmbientMcpEndpoint type and policy arrays",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/ambient/endpoint-criticality.ts",
      "description": "AMBIENT_ALL_ENDPOINTS is built from AMBIENT_REQUIRED_ENDPOINTS ∪ AMBIENT_OPTIONAL_ENDPOINTS, but runtime.ts iterates AMBIENT_ALL_ENDPOINTS as its validation scope. If AmbientMcpEndpoint (from score-mcp-client.ts) gains a new value that is not added to either policy array, validation for that endpoint is silently skipped — it will never appear in results, and classifyEndpointCriticality will never see it. The TypeScript type system does not catch this drift. An unvalidated required endpoint could go misconfigured in production without triggering the disable path.",
      "suggestion": "Add a TypeScript satisfies assertion or a test that asserts AMBIENT_ALL_ENDPOINTS contains every member of the AmbientMcpEndpoint union. E.g.: `const _exhaustive: Record<AmbientMcpEndpoint, true> = Object.fromEntries(AMBIENT_ALL_ENDPOINTS.map(e => [e, true])) as Record<AmbientMcpEndpoint, true>;` — this fails at compile time if a new endpoint is added to the union without updating the policy.",
      "confidence": 0.88,
      "faang_parallel": "Protocol Buffers at Google carry this exact failure mode: adding a new field to a .proto definition doesn't automatically update all consumers. The Go team's experience with untyped `interface{}` drift led directly to the `satisfies` operator becoming a first-class concept in TypeScript 4.9 — it was designed for this exhaustiveness pattern. Microsoft's TypeScript team was explicit: `satisfies` is the type system's way of saying 'not just assignable, but complete.'",
      "metaphor": "Imagine a cargo inspection system where adding a new category of goods — say, lithium batteries — doesn't automatically add it to the inspector's checklist. Each shipment is checked against the list, and lithium batteries pass through unchecked. The gap is invisible until something catches fire. The policy array is the checklist; the union is reality; they need a mechanical link.",
      "teachable_moment": "The `satisfies` keyword (TS 4.9+) is underused precisely here. `const _: Record<AmbientMcpEndpoint, true> = ...` turns a runtime assumption into a compile-time guarantee. It's the difference between 'we remember to update the list' and 'the compiler won't let us forget.'"
    },
    {
      "id": "F-002",
      "title": "EndpointValidationResult does not enforce reason presence when ok is false",
      "severity": "LOW",
      "category": "type-safety",
      "file": "packages/persona-engine/src/ambient/endpoint-criticality.ts",
      "description": "EndpointValidationResult declares reason as optional regardless of the ok field. When ok is false, reason can be absent and the classifier silently substitutes 'validation failed' — a fallback that provides no diagnostic signal. The type permits constructing a failure result with no reason, which will produce low-quality log output.",
      "suggestion": "Use a discriminated union: `type EndpointValidationResult = { endpoint: AmbientMcpEndpoint; ok: true } | { endpoint: AmbientMcpEndpoint; ok: false; reason: string }`. This forces callers to always supply a reason on failure and eliminates the generic fallback in classifyEndpointCriticality.",
      "confidence": 0.82,
      "faang_parallel": "Rust's `Result<T, E>` forces you to supply a typed error when returning `Err` — the language won't compile without it. Microsoft's Azure SDK team cited this property when documenting their Rust adoption rationale: the discriminated failure channel eliminates the 'failed but no context' class of bugs that were common in C++ error returns. TypeScript's discriminated union is the same pattern in a different language.",
      "metaphor": "A medical chart that lets you mark 'healthy: false' without filling in 'diagnosis' is a form that no hospital would design. The discriminated union grays out the reason field when ok is true and makes it required — not optional — when ok is false. The type system becomes the form validation.",
      "teachable_moment": "The generic fallback string 'validation failed' is a diagnostic dead end. In oncall at scale, the difference between 'MCP connection refused' and 'validation failed' is the difference between a 2-minute fix and a 45-minute investigation. Failure types that carry no context shift debugging cost from machines to humans."
    },
    {
      "id": "F-003",
      "title": "Optional-endpoint warning logs fire in all environments including dev/test",
      "severity": "LOW",
      "category": "developer-experience",
      "description": "The optional-endpoint warning loop runs before the `if (!isProd)` branch. In a local dev environment where FREESIDE_AUTH_MCP_URL is unset (the common case per the PR description), every module-load emits a console.warn. The early-return for dev mode suppresses required-endpoint noise but not optional-endpoint noise.",
      "suggestion": "Gate the optional-endpoint warning behind `isProd` or behind a dedicated dev verbosity flag, or emit it at debug level. The production signal is the important one; in dev the STUB_MODE message is already present and the optional warning adds noise without action surface.",
      "confidence": 0.70,
      "faang_parallel": "Cloudflare's 2021 incident post-mortem identified alert fatigue from undifferentiated warnings as a contributing factor — engineers had trained themselves to ignore warnings that fired constantly in dev. Google's `absl::LOG` uses severity levels specifically to preserve the signal value of WARNING: if it fires in every dev run, it stops being a warning and becomes wallpaper.",
      "metaphor": "A car that beeps every time you start it in your driveway eventually becomes a car you don't notice beeping on the highway. Log warnings are the same — constant low-environment noise degrades the signal in production where it matters.",
      "teachable_moment": "The threshold for `console.warn` should be: 'would an engineer seeing this in a CI log want to act on it?' In dev with a missing optional endpoint, the answer is no — the dev already knows, and the STUB_MODE message already covers it. Save `warn` for signals that survive aggregation."
    },
    {
      "id": "F-004",
      "title": "Unknown endpoint (not in either policy list) behavior is untested",
      "severity": "LOW",
      "category": "test-coverage",
      "file": "packages/persona-engine/src/ambient/endpoint-criticality.test.ts",
      "description": "classifyEndpointCriticality treats any endpoint not found in AMBIENT_OPTIONAL_ENDPOINTS as required — a fail-safe default that is correct but undocumented by any test. If the function is ever called with an endpoint outside the policy arrays (possible if score-mcp-client.ts exports an endpoint value not yet listed), it will silently become tier-fatal. The contract for this path is implicit.",
      "suggestion": "Add a test case: `classifyEndpointCriticality([{ endpoint: 'unknown-endpoint' as AmbientMcpEndpoint, ok: false, reason: 'test' }])` — assert disabled:true and the endpoint in reasons. This documents the fail-safe contract and will catch a future change that accidentally makes unknown endpoints optional.",
      "confidence": 0.80,
      "faang_parallel": "Amazon's engineering culture mandates testing every implicit contract, not just the happy path. The 'Correction of Errors' (COE) process surfaces cases where behavior was assumed rather than specified — and the resolution almost always includes a test that documents the assumption. The 'unknown becomes required' default is a good security posture; it just needs to be explicit.",
      "metaphor": "A bouncer whose unstated policy is 'if you're not on the list, you don't get in' is running a sound security protocol. But when that bouncer calls in sick, the replacement needs the policy written down — otherwise they might flip the default. The test is the written policy.",
      "teachable_moment": "Fail-safe defaults are security-positive choices, but undocumented defaults become landmines when someone refactors the logic. A test that asserts 'unknown endpoint → tier-fatal' is both a regression guard and a specification. Future readers will understand intent, not just behavior."
    },
    {
      "id": "F-005",
      "title": "Pure policy module extraction enables reliable unit testing",
      "severity": "PRAISE",
      "category": "architecture",
      "file": "packages/persona-engine/src/ambient/endpoint-criticality.ts",
      "description": "Extracting the criticality classification into a side-effect-free module — no I/O, no env reads, no ManagedRuntime construction — allows the policy to be tested without any infrastructure setup. The existing tests confirm this: they run under bun test without installed deps. This also means the policy is independently auditable and the runtime.ts integration is reduced to data plumbing.",
      "faang_parallel": "This is Gary Bernhardt's 'Functional Core, Imperative Shell' pattern, and it's how Stripe separates payment calculation (pure functions, auditable by compliance) from payment execution (side effects, tested with infrastructure). The pure core can be reviewed by anyone who can read code — no running system required. Stripe's fraud detection team uses this separation to let policy changes ship without infrastructure risk.",
      "metaphor": "The criticality classifier is a city zoning map — it describes what's allowed where, but it doesn't build anything. The runtime is the construction crew that reads the map and acts on it. Separating the map from the crew means you can update zoning rules without scaffolding a construction site to test them.",
      "teachable_moment": "Side-effect-free modules are inherently more testable, auditable, and portable. The signal that you've achieved this is when your test file has zero mocks — as this one does. That's not an accident; it's the reward for correct separation."
    },
    {
      "id": "F-006",
      "title": "Regression test anchors to the exact production log string",
      "severity": "PRAISE",
      "category": "test-quality",
      "file": "packages/persona-engine/src/ambient/endpoint-criticality.test.ts",
      "description": "FREESIDE_AUTH_UNCONFIGURED is declared as the literal string that production actually logged, with a comment explaining its source. This means the test will fail if score-mcp-client.ts changes its validation message — catching message drift that would invalidate the documented regression. The test is tied to the observable symptom, not just to the internal logic.",
      "faang_parallel": "Google's golden file testing and Stripe's regression anchoring both pin tests to observed production behavior, not just designed behavior. The distinction is significant: designed behavior is what the author intended; observed behavior is the symptom that reached users. Anchoring to the production string means the test is documenting an incident, not a specification — and incident documentation is more durable.",
      "metaphor": "A medical regression test that checks for 'temperature above 102°F' is testing a symptom. A test that checks for 'the exact symptom the patient presented with on 2026-05-15' is documenting a case. The latter catches recurrence even when the mechanism changes.",
      "teachable_moment": "Named constants for production-observed strings serve two functions: they survive grep (so you can find what file owns the message), and they document provenance (the comment explains why this exact string matters). This is the kind of test authorship that makes a codebase readable as forensics six months later."
    },
    {
      "id": "F-007",
      "title": "codex in REQUIRED is a temporary coupling due to MiberaResolverLive not failing soft",
      "severity": "SPECULATION",
      "category": "architecture",
      "file": "packages/persona-engine/src/ambient/endpoint-criticality.ts",
      "description": "The module comment explicitly acknowledges that codex is tier-fatal not because its data is unavailable via fallback, but because MiberaResolverLive propagates transport errors rather than catching them. This means codex behaves differently from the intended policy (enrichment degrades, tier stays up) due to an implementation detail in the resolver, not a semantic requirement. The coupling is documented but creates a latent risk: any future caller of classifyEndpointCriticality may assume codex is required for functional reasons, not accidental ones.",
      "suggestion": "Track the MiberaResolverLive fail-soft follow-up as a beads task with a reference back to this module. Once that resolver degrades gracefully, codex can be moved to AMBIENT_OPTIONAL_ENDPOINTS and the comment removed — the policy will then match actual semantic criticality.",
      "confidence": 0.65,
      "faang_parallel": "This is a well-known class of technical debt: 'accidental coupling documented as intentional policy.' The Linux kernel has several syscall behaviors that were implementation accidents that became API contracts because no test distinguished them from intentional design. The comment here is the right mitigation — documenting that this is accidental keeps future engineers from cargo-culting the constraint.",
      "metaphor": "A bridge rated for 10 tons because the original engineer only had 10-ton I-beams available, not because 10 tons is the real limit. The comment is the sticky note on the bridge spec that says 'check the actual load bearing before assuming this is a hard limit.' The note is good; the follow-up structural assessment is better.",
      "teachable_moment": "Self-documenting accidental constraints is underrated practice. The comment in this file is doing exactly the right thing: it names the constraint, names the cause, and implies the future fix. The only gap is persistence — a comment in source can drift or be removed. A beads task with a back-reference to the module creates a more durable contract."
    }
  ]
}
```


---

## Callouts

**F-005 — Pure Policy Module**: The architectural move here is clean and replicable. `endpoint-criticality.ts` has no imports from the runtime layer, no env reads, no async anything. The entire module is a policy table and a classifier. This is the design that makes the test file have zero mocks — that's not a coincidence, it's the payoff for getting the separation right. Future policy changes (moving an endpoint from required to optional, adding a new criticality tier) are now risk-free: change the table, run the pure tests, done.

**F-006 — Regression Anchoring**: The `FREESIDE_AUTH_UNCONFIGURED` constant pinned to the production log string is small but significant. It means this test will catch message drift in `score-mcp-client.ts` without the author of that change even knowing this test exists. That's a regression net with reach. The comment explaining the string's provenance turns the test into incident documentation — readable as both a spec and a forensic record.

**Overall posture**: The PR is doing something right at the architectural level — it's not just fixing a crash, it's building infrastructure for *classifying and reasoning about* endpoint health in a principled way. The policy/runtime separation means the next resilience fix in this area will be cheaper than this one. That compound return on architectural investment is what separates a bug fix from an improvement.

---

PR #93, opened 2026-05-24
fix(persona-engine): retry score-mcp on 429/5xx so the digest sweep stops dropping zones

zkSoju:
## what broke (caught while live-testing the digest)

Fired a full digest sweep against prod (2026-05-24). 3 of 4 zones posted; **owsley-lab failed**:

```
[stonehenge/digest] posted ✓   (239 events)
[bear-cave/digest]  posted ✓   (empty window)
[el-dorado/digest]  posted ✓   (116 events)
[owsley-lab/digest] failed: mcp init failed: 429 Too Many Requests
```

## root cause

`cron/scheduler.ts:104` sweeps all zones in a **tight loop** — four rapid `get_zone_digest` calls. score-mcp rate-limits the burst, and `mcpInit`/`mcpToolCall` used a raw `fetch` that **threw on the first 429 with no retry**. Since **owsley-lab is always last**, it ate the 429 and silently dropped its digest — every Sunday, not just this test.

## fix

New retry transport (`score/retry.ts`); both `mcpInit` and `mcpToolCall` route through it, so **every** score-mcp consumer benefits:

| | |
|---|---|
| retries | `429` + `502/503/504` + thrown network errors |
| Retry-After | honored (delta-seconds *and* HTTP-date), capped 60s |
| else | exponential backoff, equal-jitter, capped 8s |
| non-retryable 4xx | pass straight through — caller's `if (!res.ok) throw` unchanged |
| exhausted | returns the final 429 Response (existing error messages preserved) |
| observability | logs `score-mcp: retry N in Xms — reason` for the Sunday cron |

`retry.ts` is pure + injectable (`fetchImpl` / `sleep` / `random`) — **16 unit tests** run with no network, no real timers, no installed deps.

## verification

- `tsc --noEmit` clean
- **41 pass + 1 pre-existing skip** across `score` + `ambient` suites
- the dropped zone (owsley-lab) is the regression target; post-merge I'll redeploy and confirm a clean 4/4 sweep

### DIFF
--- a/packages/persona-engine/src/score/client.ts
+++ b/packages/persona-engine/src/score/client.ts
@@ -57,17 +57,33 @@ const MCP_PROTOCOL_VERSION = '2024-11-05';
  * Retry policy for score-mcp transport calls. The weekly digest cron sweeps
  * all zones in a tight loop, so the burst trips score-mcp's rate limit and the
  * last zone (owsley-lab) used to drop on the first un-retried 429. Honors the
- * server's Retry-After and logs each backoff so the waits are visible in the
- * Sunday cron's prod logs.
+ * server's Retry-After and logs each backoff (labelled init vs tool) so the
+ * waits are visible in the Sunday cron's prod logs.
  */
-const SCORE_RETRY_OPTS: FetchRetryOptions = {
-  onRetry: ({ attempt, status, delayMs, reason }) => {
-    console.warn(
-      `score-mcp: retry ${attempt} in ${delayMs}ms — ${reason}` +
-        (status ? ` (status ${status})` : ''),
-    );
-  },
-};
+function scoreRetryOpts(
+  label: 'init' | 'tool',
+  extra: Partial<FetchRetryOptions> = {},
+): FetchRetryOptions {
+  return {
+    ...extra,
+    onRetry: ({ attempt, status, delayMs, reason }) => {
+      console.warn(
+        `score-mcp/${label}: retry ${attempt} in ${delayMs}ms — ${reason}` +
+          (status ? ` (status ${status})` : ''),
+      );
+    },
+  };
+}
+
+// The initialize handshake is NOT idempotent under 5xx: a 504 gateway timeout
+// may mean the upstream already created a session before the proxy gave up, so
+// retrying would orphan it. Restrict init retries to 429, where the server
+// provably rejected the request and created nothing (the actual digest-sweep
+// failure mode). tools/call keeps the full transient set.
+const SCORE_INIT_RETRY_OPTS: FetchRetryOptions = scoreRetryOpts('init', {
+  retryableStatuses: new Set([429]),
+});
+const SCORE_TOOL_RETRY_OPTS: FetchRetryOptions = scoreRetryOpts('tool');
 
 /** Parse a single SSE response body into the embedded JSON-RPC envelope. */
 function parseSseEnvelope<T>(body: string): McpJsonRpcEnvelope<T> {
@@ -108,7 +124,7 @@ async function mcpInit(url: string, key: string, bearer?: string): Promise<McpIn
         },
       }),
     },
-    SCORE_RETRY_OPTS,
+    SCORE_INIT_RETRY_OPTS,
   );
 
   if (!response.ok) {
@@ -163,7 +179,7 @@ async function mcpToolCall<T>(
         params: { name: toolName, arguments: toolArgs },
       }),
     },
-    SCORE_RETRY_OPTS,
+    SCORE_TOOL_RETRY_OPTS,
   );
 
   if (!response.ok) {
--- a/packages/persona-engine/src/score/client.ts
+++ b/packages/persona-engine/src/score/client.ts
@@ -33,6 +33,7 @@ import type {
   BadgeType,
 } from './types.ts';
 import { ZONE_TO_DIMENSION } from './types.ts';
+import { fetchWithRetry, type FetchRetryOptions } from './retry.ts';
 
 interface McpInitResult {
   sessionId: string;
@@ -51,6 +52,22 @@ interface McpToolResult {
 }
 
 const MCP_PROTOCOL_VERSION = '2024-11-05';
+
+/**
+ * Retry policy for score-mcp transport calls. The weekly digest cron sweeps
+ * all zones in a tight loop, so the burst trips score-mcp's rate limit and the
+ * last zone (owsley-lab) used to drop on the first un-retried 429. Honors the
+ * server's Retry-After and logs each backoff so the waits are visible in the
+ * Sunday cron's prod logs.
+ */
+const SCORE_RETRY_OPTS: FetchRetryOptions = {
+  onRetry: ({ attempt, status, delayMs, reason }) => {
+    console.warn(
+      `score-mcp: retry ${attempt} in ${delayMs}ms — ${reason}` +
+        (status ? ` (status ${status})` : ''),
+    );
+  },
+};
 
 /** Parse a single SSE response body into the embedded JSON-RPC envelope. */
 function parseSseEnvelope<T>(body: string): McpJsonRpcEnvelope<T> {
@@ -71,24 +88,28 @@ function authHeaders(key: string, bearer?: string): Record<string, string> {
 }
 
 async function mcpInit(url: string, key: string, bearer?: string): Promise<McpInitResult> {
-  const response = await fetch(url, {
-    method: 'POST',
-    headers: {
-      'Content-Type': 'application/json',
-      Accept: 'application/json, text/event-stream',
-      ...authHeaders(key, bearer),
-    },
-    body: JSON.stringify({
-      jsonrpc: '2.0',
-      id: 1,
-      method: 'initialize',
-      params: {
-        protocolVersion: MCP_PROTOCOL_VERSION,
-        clientInfo: { name: 'freeside-characters', version: '0.6.0' },
-        capabilities: {},
+  const response = await fetchWithRetry(
+    url,
+    {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        Accept: 'application/json, text/event-stream',
+        ...authHeaders(key, bearer),
       },
-    }),
-  });
+      body: JSON.stringify({
+        jsonrpc: '2.0',
+        id: 1,
+        method: 'initialize',
+        params: {
+          protocolVersion: MCP_PROTOCOL_VERSION,
+          clientInfo: { name: 'freeside-characters', version: '0.6.0' },
+          capabilities: {},
+        },
+      }),
+    },
+    SCORE_RETRY_OPTS,
+  );
 
   if (!response.ok) {
     throw new Error(`mcp init failed: ${response.status} ${await response.text()}`);
@@ -125,21 +146,25 @@ async function mcpToolCall<T>(
   toolArgs: Record<string, unknown>,
   bearer?: string,
 ): Promise<T> {
-  const response = await fetch(url, {
-    method: 'POST',
-    headers: {
-      'Content-Type': 'application/json',
-      Accept: 'application/json, text/event-stream',
-      ...authHeaders(key, bearer),
-      'Mcp-Session-Id': sessionId,
+  const response = await fetchWithRetry(
+    url,
+    {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        Accept: 'application/json, text/event-stream',
+        ...authHeaders(key, bearer),
+        'Mcp-Session-Id': sessionId,
+      },
+      body: JSON.stringify({
+        jsonrpc: '2.0',
+        id: Math.floor(Math.random() * 1e9),
+        method: 'tools/call',
+        params: { name: toolName, arguments: toolArgs },
+      }),
     },
-    body: JSON.stringify({
-      jsonrpc: '2.0',
-      id: Math.floor(Math.random() * 1e9),
-      method: 'tools/call',
-      params: { name: toolName, arguments: toolArgs },
-    }),
-  });
+    SCORE_RETRY_OPTS,
+  );
 
   if (!response.ok) {
     throw new Error(`mcp tools/call failed: ${response.status} ${await response.text()}`);
--- a/packages/persona-engine/src/score/retry.test.ts
+++ b/packages/persona-engine/src/score/retry.test.ts
@@ -1,0 +1,175 @@
+// Tests for the score-MCP retry transport. Pure: injected fetch (response
+// factories), injected sleep (records delays, never waits), injected random
+// (deterministic jitter). Runs under `bun test` with no installed deps.
+//
+// Regression intent: the digest cron's rapid zone sweep made score-mcp 429,
+// and the un-retried transport dropped the last zone (owsley-lab). These tests
+// pin the retry contract that fixes it.
+
+import { describe, expect, test } from "bun:test";
+import {
+  computeBackoffMs,
+  fetchWithRetry,
+  parseRetryAfterMs,
+} from "./retry.ts";
+
+type Step = (() => Response) | Error;
+
+/** Fake fetch driven by a list of per-call response factories (or an Error to
+ * throw). The last step repeats once exhausted; factories run per call so each
+ * Response body is fresh/drainable. */
+function fakeFetch(steps: Step[]) {
+  let calls = 0;
+  const fn = (async () => {
+    const step = steps[Math.min(calls, steps.length - 1)]!;
+    calls++;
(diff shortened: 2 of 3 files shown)


zkSoju (commented):
## Summary

The retry layer is solid engineering — correct connection hygiene, injectable seams that make every timing branch testable without real I/O, and a subtly right decision to capture the JSON-RPC `id` before the retry loop. The primary concern is initialization idempotency: `mcpInit` retried on 5xx sits in the class of "state-creating operations with unknown outcome" — safe only if the upstream makes an explicit guarantee. Secondary concerns are observability gaps that will matter as zone count grows and the Sunday cron tail latency becomes the story you're debugging at 11pm.

---

### Architectural Observation

This PR draws the classic "at-least-once vs at-most-once" boundary in exactly the right place for tool calls, and in the wrong place for session initialization. The distinction matters: a rate-limited 429 on `initialize` is provably safe to retry — the server rejected before processing. A 504 on `initialize` is unknowable — the upstream may have committed the session before the proxy gave up. The retry policy in this PR treats both identically. As written, it trades the original bug (dropped zones) for a rarer but harder-to-debug one (orphaned sessions accumulating in score-mcp under timeout-heavy conditions).

---

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "BR-001",
      "title": "mcpInit retried on 5xx may create orphaned sessions",
      "severity": "MEDIUM",
      "category": "correctness",
      "description": "fetchWithRetry retries mcpInit on 502, 503, and 504 via the shared SCORE_RETRY_OPTS. For 429 this is provably safe — the server rejected the request, no session was created. For 504 (gateway timeout), the upstream MCP server may have already created a session and issued a session ID before the proxy's clock expired. A retry sends a second `initialize`, potentially creating a second session; the first leaks. 502 is similarly ambiguous depending on where in the upstream stack the failure occurred.",
      "suggestion": "Pass a custom `retryableStatuses: new Set([429])` to fetchWithRetry specifically for mcpInit. The tool-call path can retain the full 429/502/503/504 set since tool-call idempotency depends on the operation, not the handshake. If score-mcp documents that initialize is idempotent under all 5xx conditions, add that reference as a comment.",
      "confidence": 0.7,
      "faang_parallel": "Stripe's idempotency key pattern separates 'safe to retry' from 'unknown state' at exactly this boundary. Their payment intent API distinguishes client-side 4xx (rejected before processing — retry freely) from 5xx with no idempotency key (outcome unknown — do not retry without a new key). Google's Spanner commit protocol handles the same split: a `ABORTED` response means 'definitely not committed, retry safely'; a network timeout means 'outcome unknown, use read-to-verify before retry.' The MCP initialize path sits in Spanner's timeout class, not Stripe's 4xx class.",
      "metaphor": "Checking into a hotel over the phone: if you hang up mid-reservation and call back to start again, you may end up with two rooms billed. The first booking exists in the system — it just has no guest attached. Score-mcp's session table is that hotel registry.",
      "teachable_moment": "The mental model: classify every retryable call as either 'rejected before state change' (safe: 429, 400, client-side timeout before send) or 'unknown state post-send' (unsafe without idempotency: any server-side 5xx, network timeout after TCP connect). State-creating operations must only retry in the first class unless the server explicitly guarantees idempotency. Document which class each call site falls into — that comment is load-bearing for the next engineer.",
      "connection": "BR-008 (stable JSON-RPC ID) shows this codebase already understands idempotency for tool calls. The same instinct should be applied to mcpInit — either restrict retryable statuses to 429, or add the idempotency guarantee as a documented server contract."
    },
    {
      "id": "BR-002",
      "title": "No test for Retry-After: 0 flowing through fetchWithRetry",
      "severity": "LOW",
      "category": "testing",
      "file": "packages/persona-engine/src/score/retry.test.ts",
      "description": "`parseRetryAfterMs('0')` returning 0 is tested in the unit section. When this flows into fetchWithRetry, `Math.min(0, maxRetryAfterMs) = 0` and `sleep(0)` is called — an immediate retry. The fetchWithRetry test suite has no case covering a `Retry-After: 0` response; the path executes correctly but is unexercised at the integration level.",
      "suggestion": "Add: `r(429, { 'retry-after': '0' })` followed by `r(200)`, assert `s.ms` equals `[0]` and `ff.calls()` equals 2.",
      "confidence": 0.95,
      "faang_parallel": "Boundary value analysis is one of Facebook's mandatory test-coverage principles for network client code. The zero-delay case is a classic off-by-one boundary: `sleep(0)` is semantically 'yield and retry immediately,' which is distinct from both 'wait and retry' and 'do not retry.' Each has different queue-pressure implications under load.",
      "metaphor": "Testing Retry-After: 0 is like verifying a traffic light at the exact moment the timer hits zero — not 1ms after, not 1ms before. The behavior at the boundary is the behavior that breaks assumptions.",
      "teachable_moment": "The unit-to-integration gap: you can have perfect unit coverage of a pure function (`parseRetryAfterMs`) and a gap in the integration test of the call site that uses it. Both are needed. When a function feeds directly into another as a named parameter, always add an integration test for the boundary values of the inner function's output range.",
      "connection": "The fakeFetch / sleep recorder infrastructure (BR-006) makes this test a two-line addition. The seam is already there; the test is just missing."
    },
    {
      "id": "BR-003",
      "title": "onRetry uses console.warn — no callsite disambiguation between init and tool-call",
      "severity": "LOW",
      "category": "observability",
      "description": "SCORE_RETRY_OPTS is shared by both mcpInit and mcpToolCall. The log line (`score-mcp: retry N in Xms — reason`) does not indicate which call site is retrying. In a Sunday cron log showing four zone sweeps each with a session init, distinguishing an init-level 429 from a tool-call 429 requires timestamp correlation rather than the log message itself.",
      "suggestion": "Either use distinct opts objects per callsite with a prefix in the log string (`score-mcp/init:` vs `score-mcp/tool:`), or add a `label` parameter to `onRetry` info. Alternatively, the shared opts are fine if the cron log volume is low enough that timestamp ordering makes context obvious — mark with a TODO.",
      "confidence": 0.6,
      "faang_parallel": "Netflix's Hystrix circuit breaker names every command group and thread pool explicitly, because the first question in every incident is 'which command is saturated?' The log prefix is the minimum viable version of that discipline. Without it, you're correlating by timestamp under pressure — doable, but expensive attention.",
      "metaphor": "A hospital PA saying 'code blue, please respond' without specifying the floor. Everyone knows something is urgent; no one knows where to go.",
      "teachable_moment": "Shared configuration objects are DRY but opaque at the log site. The callsite label doesn't need to be part of the config object — it can be a closure argument to the onRetry callback, keeping the retry policy shared while making the log line specific. This is the 'same behavior, distinct identity' pattern.",
      "connection": "This finding is low severity specifically because the cron volume is likely small enough for timestamp correlation to work today. The note is a preemptive flag for when zone count grows — BR-005 addresses the structural reason that might happen faster than expected."
    },
    {
      "id": "BR-004",
      "title": "console.warn in SCORE_RETRY_OPTS bypasses any structured logger in scope",
      "severity": "LOW",
      "category": "observability",
      "description": "The onRetry callback calls console.warn directly. If persona-engine uses a structured logger (common in Node/Bun services with trace IDs or zone context), these retry events won't carry correlation fields. In the digest cron, knowing which zone triggered a retry is the primary diagnostic signal.",
      "suggestion": "If a logger or zone-scoped context is accessible at the call site, thread it into the onRetry callback. If console.warn is idiomatic here, add the zone identifier from the outer scope to the log string.",
      "confidence": 0.5,
      "faang_parallel": "Google Cloud's structured logging discipline: every log line carries a trace ID, so any log from a cron sweep is correlatable to the job run that produced it. The retry log without a zone ID is the equivalent of a Stackdriver entry with no `labels` — technically present, operationally orphaned.",
      "metaphor": "Writing a sticky note ('retry happened!') when your team runs on Jira. The information exists; it's just disconnected from every other piece of context needed to act on it.",
      "teachable_moment": "The gap here is a dependency injection question at the module level: `SCORE_RETRY_OPTS` is a module-level constant, so it can't capture zone-scoped context. The fix is to make `SCORE_RETRY_OPTS` a factory function that accepts a logger (or at minimum a context tag), returning the opts object. This is a minor refactor with large observability upside as zone count grows.",
      "connection": "Confidence is 0.5 because we can't see whether persona-engine has a structured logger. If `console.warn` is the project idiom, this is a non-issue. Verify before acting."
    },
    {
      "id": "BR-005",
      "title": "Root cause is the unconstrained burst; retry treats the symptom",
      "severity": "REFRAME",
      "category": "architecture",
      "description": "The PR correctly identifies the root cause: the digest cron sweeps all zones in a tight loop, producing a burst that trips score-mcp's rate limiter. The retry is a correct and necessary safety net. However, the burst itself still fires on every cron run. As zone count grows, the retry wait inflates total sweep latency by `(rate-limited zones) × backoff_ms` per run, and the Sunday cron tail latency becomes unpredictable.",
      "suggestion": "Add a configurable inter-zone delay (200–500ms) in the digest cron's zone-sweep loop to prevent hitting the rate limit in the first place. The retry layer remains the backstop for transient failures, not the primary flow-control mechanism.",
      "confidence": 0.8,
      "faang_parallel": "Twitter's approach to their internal firehose: retry-with-backoff was added first, then p-throttle was added to the producer after observing that retries were becoming the dominant flow-control path in production. The pattern: retry is the circuit breaker for transient failures; rate-matching at the producer is the defense against predictable load. When retries become predictable (as opposed to occasional), the producer is the bug. Stripe's request batching and Lyft's per-service rate limiting follow the same model.",
      "metaphor": "A seatbelt is necessary and correct. But if every drive involves deploying the airbags, the fix isn't a better airbag — it's a different route. The retry is the airbag. The inter-zone delay is the better route.",
      "teachable_moment": "Reliability has layers: rate-matching at the producer (don't burst), bulkhead isolation (don't let one zone's failure propagate), retry with backoff (handle transient failures gracefully), circuit breaker (stop calling a dead service). This PR correctly adds the third layer. The second layer (not bursting) is absent. As a general principle: when you find yourself relying on a downstream layer to handle a load pattern you control at the source, add a producer-side throttle. Retry should fail rarely — not predictably on every run.",
      "connection": "This is the only REFRAME finding in the set — it's not a bug, it's a framing observation. The PR is correct and shippable. The inter-zone delay is a follow-up. Consider logging the number of retries-per-run as a metric; when that number starts climbing, the burst throttle becomes urgent."
    },
    {
      "id": "BR-006",
      "title": "Injectable fetch/sleep/random makes every timing and jitter branch testable without real I/O",
      "severity": "PRAISE",
      "category": "testing",
      "file": "packages/persona-engine/src/score/retry.ts",
      "description": "FetchRetryOptions exposes fetchImpl, sleep, and random as first-class seams. Combined with the fakeFetch factory (last-step-repeats design for exhaustion tests) and the sleep recorder, every path — Retry-After honor, backoff cap, network error vs HTTP error, maxAttempts exhaustion — is exercised with zero real timers or network calls. The tests run under `bun test` with no installed deps as stated.",
      "confidence": 1.0,
      "faang_parallel": "This is Hexagonal Architecture (Ports and Adapters) applied at the function level — exactly the pattern Google's Go testing guidelines recommend for any I/O-touching code. The 'inject your syscalls' principle from the Go standard library (e.g., the `time.Now` injection pattern in testing clock-dependent code) is precisely what's happening here. The result: tests that are hermetic, fast, and exhaustive without mocks of HTTP clients that drift from reality.",
      "metaphor": "A flight simulator with controls mapped 1:1 to the real cockpit. You're not approximating flight — you're running the exact same logic against a fake sky. Every branch of the retry algorithm is reachable without waiting for a real server to be rate-limiting.",
      "teachable_moment": "The 'last-step-repeats' design for exhaustion tests deserves callout: rather than requiring the test author to pre-populate exactly N responses for an N-attempt test, the fakeFetch factory repeats its final response for any request beyond the queue length. This makes the test intent legible (here's the sequence I care about) without boilerplate padding. Reusable as a pattern anywhere a sequence of responses needs to be exercised.",
      "connection": "This design directly enables BR-002's suggested test to be a two-line addition. The infrastructure investment pays dividends immediately."
    },
    {
      "id": "BR-007",
      "title": "Body drain before retry releases connections before the backoff sleep",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "packages/persona-engine/src/score/retry.ts",
      "description": "The `await response.text()` call (with swallowed drain errors) inside the retry branch ensures the connection is returned to the pool before sleeping. Without this, a retryable response with an unconsumed body would hold a connection open for the full backoff window — exactly the scenario this PR is designed to handle: multiple in-flight zone sweep requests competing for connections.",
      "confidence": 1.0,
      "faang_parallel": "AWS SDK v3's HTTP handler does exactly this — unconsumed response bodies are a documented source of connection pool exhaustion in the Node.js HTTP/1.1 implementation. The SDK explicitly calls `response.body.destroy()` or drains to `/dev/null` before returning a response to the pool. The Go `net/http` documentation famously warns: 'The Response.Body must always be closed.' This PR gets that right in TypeScript, where the behavior is less loudly documented.",
      "metaphor": "Returning a library book to the drop box before taking your nap. If you hold the book (connection) while sleeping (backoff), the next person can't check it out. The drain is the drop-box deposit.",
      "teachable_moment": "Connection pool exhaustion is one of the hardest classes of production bugs to diagnose because the symptom (timeouts, hangs) appears far from the cause (unconsumed response bodies upstream). The rule: in any retry loop that sleeps, drain the response body before sleeping. This applies to fetch, axios, node-fetch, got — anything that uses an underlying HTTP/1.1 or HTTP/2 connection pool.",
      "connection": "This decision is directly load-bearing for the PR's stated goal: stopping the digest sweep from dropping zones. The orphaned connections during backoff would have reproduced the original bug even with correct retry logic."
    },
    {
      "id": "BR-008",
      "title": "JSON-RPC request body is serialized once; retries carry a stable ID",
      "severity": "PRAISE",
      "category": "correctness",
      "description": "In mcpToolCall, `body: JSON.stringify({ id: Math.floor(Math.random() * 1e9), ... })` is evaluated when constructing the `init` object literal, before fetchWithRetry is called. Every retry attempt passes the same `init` reference, so the JSON-RPC `id` is constant across all attempts. This gives the server the opportunity to deduplicate by ID if it implements that behavior, and avoids the subtle bug of a new random ID per retry that would make server-side deduplication impossible.",
      "confidence": 1.0,
      "faang_parallel": "This mirrors Stripe's idempotency key pattern precisely: the key (here, the JSON-RPC `id`) is generated once by the client and carried through all retry attempts. Stripe's engineering blog explicitly credits this design for making their payment retry logic safe under network partitions. The JSON-RPC 2.0 spec (section 5) notes that `id` MAY be used by the server to correlate responses to requests — a server implementing deduplication would use exactly this stability guarantee.",
      "metaphor": "Stamping your package with a tracking number before handing it to the courier. If the courier loses it and you send a replacement, it carries the same number — the warehouse knows it's the same shipment and won't double-fulfill.",
      "teachable_moment": "The subtle bug this avoids: if `JSON.stringify(...)` were inside the fetch call passed to fetchWithRetry (as a lambda that re-evaluates on each attempt), each retry would carry a new random ID. The server cannot deduplicate what it can't identify as duplicate. Serialize-once-carry-through is the correct pattern for any retryable call with an idempotency key. This should be noted in a comment so future authors don't 'simplify' the lambda form.",
      "connection": "This is the tool-call counterpart to BR-001's concern about mcpInit. For tool calls, the stable ID is present. For initialization, the idempotency question is unresolved. Resolving BR-001 would make the two call sites consistent in their safety model."
    }
  ]
}
```


---

## Callouts

**BR-006 — Dependency injection at the function level** is the standout engineering decision in this PR. The `fetchImpl / sleep / random` seam pattern is not default behavior in TypeScript retry utility code — it requires intentional design upfront and produces tests that are hermetic, fast, and exhaustive without any HTTP client mocking infrastructure. This is the difference between a test suite that gives you confidence and one that gives you coverage numbers. The `fakeFetch` factory's "last-step-repeats" design for exhaustion tests is a reusable pattern worth elevating into a shared test utility if it isn't already.

**BR-007 — Body drain before backoff** is the kind of correctness detail that doesn't show up in code review checklists and doesn't fail tests — it fails silently in production under load, when the connection pool is exhausted and every zone sweep request starts hanging. Getting this right without a comment suggests the author internalized it rather than looked it up. Add a one-line comment for future maintainers — this decision is load-bearing and non-obvious.

**BR-008 — Stable JSON-RPC ID across retries** is a small decision with outsized correctness implications. The `init` object is constructed once outside the retry loop, which means the serialized body (including the random `id`) is stable. A future refactor that moves the `JSON.stringify` inside a lambda passed to `fetchWithRetry` would silently break the deduplication guarantee. This deserves a comment: *"id is intentionally captured before the retry loop — do not move inside the fetch lambda."*

---

Issue #78, opened 2026-05-15
diagnostic: pop-in + weaver cadences default-off — likely silent in prod


zkSoju:
## the gap

3 of 4 cron cadences default OFF. railway env vars likely unset → silence.

| cadence | default | likely prod state |
|---|---|---|
| digest | weekly ✓ | firing (operator sees sat/sun) |
| pop-in | `POP_IN_ENABLED='false'` | dark |
| weaver | `WEAVER_ENABLED='false'` | dark |
| event-heartbeat | enabled ✓ | firing, but invisible by design (kansei sibling channel only) |

## evidence

`packages/persona-engine/src/config.ts:161,170`:
```ts
POP_IN_ENABLED: z.string().default('false').transform((v) => v === 'true'),
WEAVER_ENABLED: z.string().default('false').transform((v) => v === 'true'),
```

`packages/persona-engine/src/cron/scheduler.ts:108,130` — both gated `if (config.POP_IN_ENABLED)` / `if (config.WEAVER_ENABLED)`.

`apps/bot/src/index.ts:73-74` prints at boot:
```
pop-ins:        every 6h · 10.0% chance/zone/tick     ← enabled
pop-ins:        disabled                              ← env var unset (this is what we expect)
weaver:         disabled                              ← env var unset (this is what we expect)
```

## action

1. inspect Railway env: confirm whether `POP_IN_ENABLED` and `WEAVER_ENABLED` are set
2. or read prod boot log lines for the signatures above
3. flip to `true`:
   - `POP_IN_ENABLED=true` → ~1-2 pops/day across 4 zones (6h tick × 10% probability)
   - `WEAVER_ENABLED=true` → next wed 12:00 UTC → stonehenge
4. optional: tune `POP_IN_PROBABILITY` (default 0.1 — lowered V0.4 per operator "don't make ruggy annoying")

## secondary surface (post-flip)

even with cron enabled, pop-ins gate on `popInFits()` per `compose/post-types.ts:145` — flat-data zones still suppress pop-ins. that's intentional silence-on-empty-week, not a bug. just know it's a second layer of gating beyond the env flip.

## refs

- track-2026-05-15-herald-substrate-renderer-boundary.md (r1, thread A diagnostic)
- cycle-spec-ruggy-leaderboard-2026-05-15.md (sibling design pivot)

🤖 surfaced by claude opus 4.7 in DIG+KEEPER session 2026-05-15

zkSoju:
[...]

PR #94, opened 2026-05-26
fix(observability): bedrock-aware boot banner + digest-cron fingerprint (#78)

zkSoju:
## Summary

three small observability fixes so the silent-digest investigation (issue #78) leaves an actual fingerprint on the next missed cycle, and the boot banner stops lying about LLM provider state on bedrock-routed deploys.

- **banner version**: was hardcoded `v0.6.0-A` while `package.json` sits at `v0.12.0` → now pulled from `package.json` so every deploy reports its real version.
- **`describeLlmMode()` bedrock-aware**: mirrors `resolveProvider()` in `agent-gateway.ts`. previously reported `UNCONFIGURED` on bedrock-only prod even though the chat path was successfully routing through bedrock. matches the operator's exact "ruggy: UNCONFIGURED" confusion in current prod logs.
- **stale pop-in banner string**: cycle-008 slice-2a moved pop-ins to event-driven via ambient-stir router. the `every Nh · X% chance/zone` string was a holdover from the deleted §2 cadence.
- **`digest cron tick` fingerprint**: log timestamp + zones + cron-expr BEFORE the per-zone fire loop in `scheduler.ts`. lets us tell apart "cron didn't fire" from "cron fired but composeForCharacter threw silently" on the next deploy. same shape across the bot CLI for parity.

## Context · why now

55h of production logs (2026-05-24T22:34 boot → 2026-05-26T06:00 tail) captured a missed Sunday 2026-05-25T00:00 UTC digest fire. between the boot-time `ruggy: digest cron · 0 0 * * 0` line and the next ambient-stir tick at 00:00:12, ZERO entries — making it impossible to tell whether cron triggered or `composeForCharacter` swallowed an error. this PR doesn't fix the silent-digest itself; it just stops blinding the next debug pass.

`describeLlmMode()` bedrock-blindness compounded the confusion: prod banner read `llm: UNCONFIGURED` while chat-route logs at 2026-05-26T04:23 clearly showed `llmProvider: "bedrock"` working end-to-end.

## Test plan

- [x] `bun test apps/bot` → 193 pass / 0 fail
- [x] `bun test packages/persona-engine` → 959 pass / 2 skip / 0 fail
- [x] `tsc --noEmit` clean both packages
- [ ] (post-merge / next deploy) verify the banner shows `v0.12.x` + `auto→bedrock (…)` instead of `v0.6.0-A` + `UNCONFIGURED`
- [ ] (post-merge / next sunday) verify `digest cron tick at <iso>` appears in logs at 00:00 UTC

### DIFF
--- a/apps/bot/src/cli/digest-once.ts
+++ b/apps/bot/src/cli/digest-once.ts
@@ -51,8 +51,13 @@ async function main(): Promise<void> {
   const zones = selectedZones(config);
   const typeMode = pickType(config);
 
+  // bedrock-aware (mirrors describeLlmMode in apps/bot/src/index.ts and
+  // resolveProvider in agent-gateway.ts). Was previously bedrock-blind.
+  const hasBedrock = Boolean(config.AWS_BEARER_TOKEN_BEDROCK || config.BEDROCK_API_KEY);
   const llmMode = config.VOICE_DISABLED
     ? 'VOICE_DISABLED (no LLM)'
+    : config.LLM_PROVIDER === 'bedrock' || (config.LLM_PROVIDER === 'auto' && hasBedrock)
+    ? `bedrock (${config.BEDROCK_TEXT_MODEL_ID ?? 'no-model-id'} @ ${config.BEDROCK_TEXT_REGION})`
     : config.ANTHROPIC_API_KEY
     ? `anthropic-direct (${config.ANTHROPIC_MODEL})`
     : config.STUB_MODE
--- a/apps/bot/src/index.ts
+++ b/apps/bot/src/index.ts
@@ -50,8 +50,9 @@ import {
 import { pgPoolBuilder } from './lib/pg-pool-builder.ts';
 import type { WorldManifestQuestSubset } from './world-resolver.ts';
 import { publishCommands } from './lib/publish-commands.ts';
+import pkg from '../package.json' with { type: 'json' };
 
-const banner = `─── freeside-characters bot · v0.6.0-A ────────────────────────`;
+const banner = `─── freeside-characters bot · v${pkg.version} ────────────────────────`;
 
 async function main(): Promise<void> {
   const config = loadConfig();
@@ -70,7 +71,7 @@ async function main(): Promise<void> {
   console.log(`llm:            ${describeLlmMode(config)}`);
   console.log(`zones:          ${selectedZones(config).map((z) => `${ZONE_REGISTRY[z].emoji} ${z}`).join(' · ')}`);
   console.log(`digest cadence: ${config.DIGEST_CADENCE}` + (config.DIGEST_CADENCE !== 'manual' ? ` · ${config.DIGEST_DAY} ${String(config.DIGEST_HOUR_UTC).padStart(2, '0')}:00 UTC` : ''));
-  console.log(`pop-ins:        ${config.POP_IN_ENABLED ? `every ${config.POP_IN_INTERVAL_HOURS}h · ${config.POP_IN_PROBABILITY * 100}% chance/zone/tick` : 'disabled'}`);
+  console.log(`pop-ins:        ${config.POP_IN_ENABLED ? 'event-driven (cycle-008 · router-gated via ambient-stir)' : 'disabled (POP_IN_ENABLED=false)'}`);
   console.log(`weaver:         ${config.WEAVER_ENABLED ? `${config.WEAVER_DAY} ${String(config.WEAVER_HOUR_UTC).padStart(2, '0')}:00 UTC → ${config.WEAVER_PRIMARY_ZONE}` : 'disabled'}`);
   console.log(`delivery:       ${describeDelivery(config)}`);
 
@@ -385,10 +386,33 @@ async function main(): Promise<void> {
 }
 
 function describeLlmMode(config: ReturnType<typeof loadConfig>): string {
-  if (config.ANTHROPIC_API_KEY) return `anthropic-direct (${config.ANTHROPIC_MODEL})`;
-  if (config.STUB_MODE) return 'STUB (canned digest)';
-  if (config.FREESIDE_API_KEY) return `freeside agent-gw (${config.FREESIDE_AGENT_MODEL})`;
-  return 'UNCONFIGURED';
+  // Mirrors `resolveProvider` in packages/persona-engine/src/compose/agent-gateway.ts.
+  // V0.12 auto-rule (operator-named 2026-05-01): bedrock-first when AWS env present.
+  // Banner was previously bedrock-blind — reported UNCONFIGURED even when bedrock was
+  // wired, masking the real provider on cost-bearing production deploys.
+  const hasBedrock = Boolean(config.AWS_BEARER_TOKEN_BEDROCK || config.BEDROCK_API_KEY);
+  switch (config.LLM_PROVIDER) {
+    case 'stub':
+      return 'STUB (canned digest)';
+    case 'anthropic':
+      return config.ANTHROPIC_API_KEY
+        ? `anthropic-direct (${config.ANTHROPIC_MODEL})`
+        : 'anthropic-MISCONFIGURED (ANTHROPIC_API_KEY unset)';
+    case 'freeside':
+      return config.FREESIDE_API_KEY
+        ? `freeside agent-gw (${config.FREESIDE_AGENT_MODEL})`
+        : 'freeside-MISCONFIGURED (FREESIDE_API_KEY unset)';
+    case 'bedrock':
+      return hasBedrock
+        ? `bedrock (${config.BEDROCK_TEXT_MODEL_ID ?? 'no-model-id'} @ ${config.BEDROCK_TEXT_REGION})`
+        : 'bedrock-MISCONFIGURED (AWS_BEARER_TOKEN_BEDROCK / BEDROCK_API_KEY unset)';
+    case 'auto':
+      if (hasBedrock) return `auto→bedrock (${config.BEDROCK_TEXT_MODEL_ID ?? 'no-model-id'} @ ${config.BEDROCK_TEXT_REGION})`;
+      if (config.ANTHROPIC_API_KEY) return `auto→anthropic (${config.ANTHROPIC_MODEL})`;
+      if (config.STUB_MODE) return 'auto→stub';
+      if (config.FREESIDE_API_KEY) return `auto→freeside (${config.FREESIDE_AGENT_MODEL})`;
+      return 'UNCONFIGURED (no provider available — set LLM_PROVIDER or supply a key)';
+  }
 }
 
 function describeDelivery(config: ReturnType<typeof loadConfig>): string {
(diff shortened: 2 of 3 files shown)


zkSoju (commented):
## Summary

This PR closes a real observability gap — the boot banner now tells the truth about which LLM provider is actually wired up, and the digest cron leaves a fingerprint that distinguishes "the tick fired but did nothing" from "the tick never fired at all." The shape is right; the residual concerns are about preventing the *next* drift: a missing `default:` arm that could resurrect the silent-undefined class of bug, and two near-duplicate provider-description helpers that will inevitably diverge.

## Architectural Observations

Boot banners are a deceptively load-bearing artifact. They're the first thing an on-call engineer reads when paged at 3am, and they're the *last* thing anyone updates when internals shuffle. Google's SRE book calls this "the observability paradox" — the lines most critical to debugging are the lines most likely to lie, because they're written once and forgotten. This PR's `describeLlmMode` refactor is doing the right work, but the lack of an exhaustive `default:` arm and the duplicated string-building logic in `digest-once.ts` are the seams where the next lie will grow.

The cron fingerprint is exactly the kind of "single grep target" that pays for itself within one incident. The two-line WHY-comment in `scheduler.ts` is genuinely excellent — it pins the invariant that the log MUST precede per-zone work, so a future refactor can't accidentally collapse the diagnostic signal.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "describeLlmMode switch missing default arm — implicit undefined return",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "apps/bot/src/index.ts",
      "description": "The refactored describeLlmMode switches on config.LLM_PROVIDER without a default case or post-switch return. Any value outside {stub, anthropic, freeside, bedrock, auto} (future provider, typo, Zod relaxation) returns undefined, printing `llm: undefined` — the exact silent-observability hole this PR closes elsewhere.",
      "suggestion": "Add a `default:` arm returning `UNKNOWN_PROVIDER (${config.LLM_PROVIDER})`, or use a `const _exhaustive: never = config.LLM_PROVIDER` assertion at the end of the switch so type drift fails at compile time.",
      "confidence": 0.85,
      "faang_parallel": "TypeScript's `--noFallthroughCasesInSwitch` and the `assertNever` pattern came out of the Microsoft Outlook frontend team's post-mortems where added enum members silently broke render paths. The exhaustiveness check turns a runtime banner-lie into a build break.",
      "metaphor": "A switch without a default is like a thermostat with no behavior for 'between 65 and 70 degrees' — it works fine until the day the temperature lands in the gap, and then it just… does nothing, quietly.",
      "teachable_moment": "Defensive defaults in observability code are not paranoia — they're the contract that says 'if my caller drifts, I'll still tell you the truth about what happened.' The previous `return 'UNCONFIGURED'` was load-bearing precisely because it was the un-glamorous catch-all."
    },
    {
      "id": "F2",
      "title": "Banner and digest-once provider strings have drifted from a shared source",
      "severity": "LOW",
      "file": "apps/bot/src/cli/digest-once.ts",
      "description": "index.ts:describeLlmMode and digest-once.ts now compute the same provider-description string with comments saying each 'mirrors' the other. They're already not byte-identical (ternary chain vs switch, different MISCONFIGURED arms). The next bugfix in one will silently skip the other — re-introducing exactly the class of bug #78 was filed for.",
      "suggestion": "Extract a single `describeLlmMode(config)` helper (packages/persona-engine or a shared apps/bot/src/lib module) and import from both call sites. Apply to any future banners (cron worker, healthcheck).",
      "confidence": 0.8,
      "faang_parallel": "This is the 'copy-pasted truth' antipattern Amazon's Builders' Library warns about for status endpoints — two implementations of the same fact will diverge under maintenance pressure, and the divergence will only surface during an incident.",
      "metaphor": "Two clocks on the same wall, set by hand. They agree today. Six months from now, one is ten minutes off and nobody knows which one to trust during the fire drill.",
      "teachable_moment": "The 'mirrors' comment is honest about the duplication but doesn't prevent the drift — comments are documentation, not enforcement. The fix is structural: one source of truth, imported."
    },
    {
      "id": "F3",
      "title": "digest-once.ts 'auto' branch only handles bedrock — falls through silently for auto→anthropic/freeside",
      "severity": "LOW",
      "category": "observability",
      "file": "apps/bot/src/cli/digest-once.ts",
      "description": "The ternary only labels `LLM_PROVIDER === 'auto' && hasBedrock` as the auto branch. When auto resolves to anthropic or freeside, the output reads `anthropic-direct (...)` rather than `auto→anthropic (...)`. The index.ts banner makes this distinction; digest:once doesn't, so the operator can't tell whether they're on the resolver or an explicit choice.",
      "suggestion": "Either (a) share the helper from F2 so digest-once gets auto→X labeling for free, or (b) add explicit auto-branch arms here too.",
      "confidence": 0.7,
      "metaphor": "It's like a flight-status board that says 'on time' for both 'scheduled to depart on time' and 'rebooked onto a different flight that happens to leave on time.' Both are true; only one tells you what actually happened."
    },
    {
      "id": "F4",
      "title": "JSON import attribute syntax — runtime + tooling compatibility",
      "severity": "LOW",
      "category": "compatibility",
      "file": "apps/bot/src/index.ts",
      "description": "`import pkg from '../package.json' with { type: 'json' };` uses ES2025 import-attributes. Bun ≥1.1 (the pinned runtime) supports it, but TypeScript versions, IDE language servers, and codex/cursor adapters sometimes still expect the older `assert { type: 'json' }` form or trip on attributes.",
      "suggestion": "Confirm `bun run typecheck` is green across CI and downstream tooling (eslint, vitest) accepts the attribute form. Fallback: `Bun.file('../package.json').json()` or a build-time constant.",
      "confidence": 0.45,
      "teachable_moment": "Modern syntax in a single file can become a compatibility cliff for the whole repo's tooling chain. Worth a one-time verification rather than a blocker."
    },
    {
      "id": "F5",
      "title": "Scheduler tick log scope — only digest cadence is fingerprinted",
      "severity": "LOW",
      "category": "observability",
      "file": "packages/persona-engine/src/cron/scheduler.ts",
      "description": "Issue #78 was about distinguishing missed tick from missed digest. The new log lives only in the digest cadence; pop-in and weaver cadences remain silent, so the diagnostic ambiguity persists for them. Scope-shape question, not a correctness bug, but the PR title says 'digest-cron fingerprint' while the class of bug applies symmetrically.",
      "suggestion": "Apply the same single-line fingerprint to weaver (and future cadences). A small helper `logTick(label, zones, expr)` keeps the format consistent and gives ops one grep target.",
      "confidence": 0.6,
      "faang_parallel": "Netflix's observability team has a principle: every periodic job emits a 'heartbeat' on the same schema, so dashboards and alerts compose. Per-job ad-hoc logs are exactly what their 'one log shape per concern' guidance was built to prevent."
    },
    {
      "id": "F6",
      "title": "Tick log uses console.log rather than the engine's structured logger",
      "severity": "LOW",
      "category": "observability",
      "file": "packages/persona-engine/src/cron/scheduler.ts",
      "description": "Bare `console.log` in long-lived deploys (ECS/Railway) is unfilterable through log aggregation. Consistent with surrounding banner code, but the cron scheduler is where structured logging matters more than at boot.",
      "suggestion": "Check whether persona-engine has a logger abstraction (`log.info` / `log.debug`). If not, fine for now — flag for the same future logging-unification work.",
      "confidence": 0.5
    },
    {
      "id": "F7",
      "title": "Banner string hardcodes 'cycle-008 · router-gated via ambient-stir'",
      "severity": "LOW",
      "file": "apps/bot/src/index.ts",
      "description": "The pop-in banner embeds a cycle reference and subsystem name directly in boot output. When cycle-008 closes and routing evolves, this string will silently mis-describe the running system — the inverse of what #78 was about: a banner that drifts from reality.",
      "suggestion": "Either (a) drop the cycle reference, keep the descriptive phrase ('event-driven · router-gated'), or (b) derive routing path from config/feature flags so the banner reflects runtime state.",
      "confidence": 0.6,
      "metaphor": "It's a 'last updated: 2019' tag on a webpage that still says the company has 12 employees. The fact stamped into the page outlives the fact in the world."
    },
    {
      "id": "F8",
      "title": "MISCONFIGURED arms are diagnostic but boot continues anyway",
      "severity": "SPECULATION",
      "category": "design",
      "file": "apps/bot/src/index.ts",
      "description": "describeLlmMode returns `bedrock-MISCONFIGURED (...)` when credentials are missing — a clear observability win. But the function is pure description; the bot still starts and fails on the first LLM call, by which point the diagnostic banner is N hours stale in the logs.",
      "suggestion": "Consider promoting MISCONFIGURED branches into a startup precondition check (separate from the banner) that exits non-zero when an EXPLICIT provider is selected without credentials. Keep the descriptive string for the auto path (which has fallback semantics). SPECULATION — design call, not a bug.",
      "confidence": 0.4,
      "faang_parallel": "This is the 'fail-fast vs fail-soft' debate that shaped Kubernetes readiness probes — pure observability lets you see the wound; fail-fast prevents the wound from spreading silently."
    },
    {
      "id": "P1",
      "title": "Banner reflects reality across all 5 provider modes — closes silent-deploy gap",
      "severity": "PRAISE",
      "category": "observability",
      "file": "apps/bot/src/index.ts",
      "description": "describeLlmMode covers stub/anthropic/freeside/bedrock/auto explicitly, distinguishes auto→X from explicit X, surfaces MISCONFIGURED with the specific missing env var, and includes model ID + region for bedrock. The mirror-relationship comment with resolveProvider tells the next maintainer where to keep them in sync.",
      "suggestion": "(no action — keep this shape going for future providers)",
      "confidence": 0.95,
      "teachable_moment": "Observability code earns its keep when it names the *specific* missing piece ('AWS_BEARER_TOKEN_BEDROCK unset') rather than a generic 'misconfigured.' Specificity at the diagnostic boundary is what turns a 30-minute outage into a 30-second fix."
    },
    {
      "id": "P2",
      "title": "Version pulled from package.json — eliminates stale-banner footgun",
      "severity": "PRAISE",
      "file": "apps/bot/src/index.ts",
      "description": "Replacing `v0.6.0-A` with `v${pkg.version}` removes a class of bug where the banner lies about which build is running. Tiny diff, real long-term win.",
      "suggestion": "(no action)",
      "confidence": 0.95
    },
    {
      "id": "P3",
      "title": "Cron fingerprint comment explains the WHY, not the what",
      "severity": "PRAISE",
      "category": "documentation",
      "file": "packages/persona-engine/src/cron/scheduler.ts",
      "description": "'logs BEFORE per-zone work so a future missed digest can be told apart from a missed tick' captures the diagnostic invariant the line exists to preserve. Anyone tempted to move the log call below the loop will see why they shouldn't.",
      "suggestion": "(no action)",
      "confidence": 0.9,
      "teachable_moment": "The best comments don't describe the code — they describe the *invariant* the code is defending. This one does."
    }
  ]
}
```


## Callouts

The three PRAISE findings above capture the genuine wins: provider-mode coverage, version-from-source, and the WHY-comment on the cron fingerprint. The shape of this PR is a model for future observability hardening — name the specific missing thing, derive facts from sources rather than hardcoding them, and pin invariants with comments that explain *why* the code is where it is.

The strongest single suggestion for the next pass: address F1 (the default arm) before merge, and consider F2 (shared helper) as the natural follow-up that prevents this PR's good work from quietly eroding over the next two cycles.

Issue #102, opened 2026-05-26
[coord] cluster-events-pillar-v1 Events Pillar v1 — NATS substrate proof (Mibera-family + PuruPuru) → MST canary + operator-dash event-trace — see 0xHoneyJar/loa-freeside


zkSoju:
## Cross-repo coordination — Events Pillar v1 — NATS substrate proof (Mibera-family + PuruPuru) → MST canary + operator-dash event-trace

**Coordinator**: 0xHoneyJar/loa-freeside (cycle `cluster-events-pillar-v1`)
**This repo's role**: subscriber — verifies ACVP envelopes; MST-only enriched Discord announcements via kansei router
**Master plan content hash (at issue-open time)**: `sha256:cb6ebc60657221f87facad2874ca4a1fb78c361015dfde487397d2279d647ddf`

### What this issue is

This is a **coordination tracker** for Events Pillar v1 — NATS substrate proof (Mibera-family + PuruPuru) → MST canary + operator-dash event-trace work that lives in this repo. It is not itself a piece of work — it is an index pointing at the master plan and tracking aggregate progress.

The actual work is broken down into per-task PRs. Each PR description should include:

```
Closes 0xHoneyJar/loa-freeside#<coord-issue-num> (master task A.B)
```

so the coordinator's `bin/coord-sync.sh` can correlate upstream merges back to master beads.

### Tasks owned by this repo

- [ ] **Master Task DEP-1** (sprint characters-S1) — NATS subscriber at bot boot — verifies envelope (sig + prev_hash + Zod) then routes via kansei. Replaces scheduler.ts:114-120 reserved slot; fail-soft on broken envelope. Blocks on loa-freeside library.
- [ ] **Master Task DEP-2** (sprint characters-S2) — MST-only enriched Discord announcement: parallel-fetch identity nym + inventory metadata; Components-V2 render; canary → prod flip. Non-MST events flow on NATS but no Discord post (v2). Blocks on sonar DEP-1 + characters DEP-1.

### How to participate

1. **Read the master plan**: 0xHoneyJar/loa-freeside `grimoires/loa/{prd,sdd,sprint}.md` (master plan content hash above must match — otherwise the master plan has moved on and a resync PR is incoming)
[...]

PR #105, opened 2026-05-26
feat(events): NATS mint-event subscriber + bot boot wire — cluster-events-pillar-v1 DEP-1

zkSoju:
## Summary

**Sprint 2 child task** of `cluster-events-pillar-v1` (cluster-meta cycle per ADR-009 §D-7). Wires the `@0xhoneyjar/events` subscriber at bot boot — fulfills the "EVENT-DRIVEN" prophecy reserved at `packages/persona-engine/src/cron/scheduler.ts:113-119` from the cycle-008 substrate-presentation cycle.

Coord issue: #102 · coord bead: `cluster-events-pillar-coordinator-eoo`

## What ships

`packages/persona-engine/src/events/mint-event-subscriber.ts` — 226 lines:
- Connects to NATS (TLS-required in prod via `NATS_TLS_CA` cert file)
- Loads verifier: `JwksVerifier.fromUrl(jwksUrl)` in prod with 5s timeout; empty `StaticPubkeyVerifier` fallback in dev (every signature surfaces as invalid — visually loud rather than silently accepting unsigned traffic)
- `subscribeEnvelope` on `nft.mint.detected.>` wildcard, schema `NftMintDetectedSchema`, `chainStore: InMemoryPrevHashStore`, `initialPrevHashPolicy: 'any'` (backward-compat default per EVT-002)
- Handler: invokes the kansei router stub (DEP-1 scope), logs decision
- `onVerificationFailure`: logs typed reason + envelope metadata (sig-invalid, broken-chain, schema-violation, subject-mismatch, etc.) — subscriber stays alive

`apps/bot/src/index.ts:382-426` — boot wire. Gated on `NATS_URL`; absent → log warn + skip (bot boots cleanly without NATS). Stop hook chained into both shutdown paths (`config.DIGEST_CADENCE==='manual'` exit + SIGINT/SIGTERM handler).

**scheduler.ts:113-119 reserved-slot**: LEFT UNCHANGED. The "EVENT-DRIVEN" prophecy comment stays in place as a forward marker for DEP-2 (which will wire kansei stir-tier consultation per the comment text). Sacred no-touch constraint respected — cron paths untouched.

## Scope discipline (DEP-1 only)

- Subscriber + verification + kansei router stub ONLY
- Kansei stub returns `{ announce: false }` — DEP-2 will replace with real threshold logic
- NO Discord posting (DEP-2)
- NO changes to cron paths, interactions server, composer, character voice configs

## Substrate consumption — bun pattern

Per cluster sovereignty doctrine (memory `cluster-no-npm-sovereignty`), `@0xhoneyjar/events` is consumed via git-URL pinned to commit `0948a534…` (post-#227 main), NOT npm. Because the events package lives at a monorepo SUBDIR, bun resolves `github:0xHoneyJar/loa-freeside#sha` to the monorepo ROOT (whose package.json has `name: \"loa-freeside\"`), not to `packages/events/`. Workaround:

- Standard `\"@0xhoneyjar/events\": \"github:0xHoneyJar/loa-freeside#sha\"` in `packages/persona-engine/package.json` (bun accepts this directly).
- New `scripts/fixup-events-bun.sh` finds every bun-installed `node_modules/@0xhoneyjar/events` symlink and re-targets from `<bun-store>/loa-freeside/` → `<bun-store>/loa-freeside/packages/events/` (where the right `package.json` + exports map live).
- `scripts/rebuild-events-dist.sh` (verbatim cluster template from `loa-freeside/packages/events/scripts/`) handles building dist.
- Consumer also lists transitive runtime deps (`@effect/schema`, `@noble/curves`, `@noble/hashes`, `canonicalize`, `effect`, `nats`) since bun's isolated `.bun/` store means upward-walk resolution starts at the consumer's `node_modules`.

Verified end-to-end:
```
$ cd packages/persona-engine && bun -e \"import('@0xhoneyjar/events').then(m => console.log('SCHEMA_VERSION='+m.SCHEMA_VERSION))\"
OK keys= EventEnvelopeSchema,GENESIS_PREV_HASH,InMemoryPrevHashStore,JwksVerifier,LocalEd25519Signer,...
SCHEMA_VERSION=acvp-l1-v2

$ cat packages/persona-engine/node_modules/@0xhoneyjar/events/dist/SOURCE_SHA
0948a5344b21aef6a7ba7ce6e39a9258994638c0
```

Doctrine: `~/.claude/projects/.../memory/project_events-consumer-pm-patterns.md` (per-PM workarounds — sonar's pnpm variant uses `cluster.eventsPin` + node_modules bootstrap; characters' bun variant uses standard deps + symlink fixup; reconvergence path to extract events to dedicated repo would eliminate per-PM workarounds).

## Test plan

- [x] postinstall + fixup-events-bun materialize `dist/` with `acvp-l1-v2` fingerprint + valid SOURCE_SHA
- [x] 6/6 bun:test passing covering all 5 contract points from spec:
  - subscriber start + stop
  - valid envelope → kansei.route called with typed payload
  - broken-sig envelope → `signature-invalid` callback, handler NOT called
  - schema-invalid payload → `payload-schema-invalid` callback
  - handler throws → subscriber survives + processes next envelope
  - (+ 6th: kansei stub contract sanity)
- [ ] Reviewer: verify scheduler.ts:113-119 reserved comment intact (sacred no-touch)
- [ ] Reviewer: confirm bun-specific `fixup-events-bun.sh` pattern is acceptable for bun-consumer cells (vs the pnpm `cluster.eventsPin` shape sibling sonar PR uses)
- [ ] Post-merge: deploy with `NATS_URL` + `NATS_TLS_CA` + `EVENTS_JWKS_URL` env wired in characters' Railway service

## Verification pipeline (per @0xhoneyjar/events subscriber.ts)

Every envelope passes through 9 verification steps before reaching the handler. Each failure mode is typed:
1. JSON parse → `json-parse-error`
2. envelope schema (Effect.Schema strict) → `envelope-schema-invalid`
3. subject-bind (msg.subject === event_type) → `subject-mismatch`
4. payload-hash recompute → `payload-hash-mismatch`
5. signature verify (acvp-l1-v2 full-envelope binding) → `signature-invalid`
6. chain continuity + initialPrevHashPolicy → `prev-hash-broken-chain` / `initial-anchor-policy-violation`
7. per-event payload schema → `payload-schema-invalid`
8. chain advance (after payload validates — chain tracks accepted, not just signed)
9. handler dispatch → `handler-error` / `internal-error`

The subscriber NEVER throws — every failure surfaces via `onVerificationFailure` and is logged with typed reason + envelope metadata.

## Known limitations (Sprint 1 acceptable per build doc)

- **Single-process `InMemoryPrevHashStore`** — characters restart resets the subscriber's local chain view; first envelope from each publisher post-restart accepted unconditionally (per `initialPrevHashPolicy: 'any'`). Sprint 2+: tighten to `'genesis'` and/or use Redis-backed store.
- **No Discord posting yet** — DEP-2 scope (separate PR + bead `cluster-events-pillar-coordinator-zwr`).
- **Pre-existing test pollution**: `packages/persona-engine/src/voice/config-loader.test.ts` fails in full-suite run but passes in isolation; same behavior on clean origin/main. Unrelated.

## References

- Library: `loa-freeside/packages/events/` (landed in `loa-freeside#227`)
- Build doc: `loa-freeside/grimoires/loa/specs/enhance-events-pillar-v1-nft-mints.md` §3 + §4
- Sibling PR: `0xHoneyJar/sonar-api#24` (NATS publish layer) — separate PR
- Coord cycle: `cluster-events-pillar-v1` at `~/bonfire/cluster-events-pillar-coordinator/`

🤖 Generated with [Claude Code](https://claude.com/claude-code) via headless dispatch from the cluster coordinator. Human review-gated before merge.

### DIFF
--- a/apps/bot/src/index.ts
+++ b/apps/bot/src/index.ts
@@ -382,6 +382,27 @@ async function main(): Promise<void> {
           `kansei-router=stub (DEP-1 · DEP-2 will wire real routing)`,
       );
     } catch (err) {
+      // BB#105 rd-3 F-001 HIGH: distinguish JWKS-init failures (operator
+      // configured verification; misconfig is meaningful) from other
+      // subscriber failures (NATS broker down, network blip — bot can
+      // still serve cron + interactions). The events-subscriber lib's
+      // rd-1 throw on JWKS init reached HERE but was previously swallowed
+      // unconditionally, hiding the misconfig.
+      //
+      // Posture:
+      //   JWKS_URL set + subscriber throws → fail loud (process.exit(1)).
+      //     Operator opted into verification; refusing to silently disable.
+      //   JWKS_URL unset → log + continue. Subscriber failure is unexpected
+      //     but the bot's other functions (cron digests, weaver, Discord
+      //     interactions) are independent and shouldn't take a hard outage.
+      if (process.env.JWKS_URL?.trim()) {
+        console.error(
+          'events: JWKS_URL was configured but subscriber startup failed — refusing to boot the bot. ' +
+            'Fix JWKS reachability or unset JWKS_URL to fall back to dev verifier (every sig surfaces as invalid).',
+          err,
+        );
+        process.exit(1);
+      }
       console.error('events: failed to start NATS subscriber (bot continues without it):', err);
       mintEventSubscriber = null;
     }
--- a/apps/bot/src/index.ts
+++ b/apps/bot/src/index.ts
@@ -50,6 +50,11 @@ import {
 import { pgPoolBuilder } from './lib/pg-pool-builder.ts';
 import type { WorldManifestQuestSubset } from './world-resolver.ts';
 import { publishCommands } from './lib/publish-commands.ts';
+import {
+  startMintEventSubscriber,
+  createKanseiRouterStub,
+  type MintEventSubscriberHandle,
+} from '@freeside-characters/persona-engine/events/mint-event-subscriber';
 import pkg from '../package.json' with { type: 'json' };
 
 const banner = `─── freeside-characters bot · v${pkg.version} ────────────────────────`;
@@ -340,6 +345,52 @@ async function main(): Promise<void> {
   if (handle.popInExpression) console.log(`${primary.id}: pop-in cron · ${handle.popInExpression}`);
   if (handle.weaverExpression) console.log(`${primary.id}: weaver cron · ${handle.weaverExpression}`);
 
+  // ─── Cross-cell NATS mint-event subscriber (cluster-events-pillar DEP-1) ───
+  // Subscribes to `nft.mint.detected.>` published by sonar-api, verifies each
+  // ACVP envelope (sig + hash-chain + payload-schema), and hands the typed
+  // payload to a kansei router. DEP-1 ships a STUB router that always returns
+  // { announce: false } — DEP-2 wires the real threshold-based router + the
+  // Discord channel announcement.
+  //
+  // Gated on NATS_URL being set so the bot still boots cleanly without NATS
+  // (e.g. local dev without the cluster broker). When absent, we log a warn
+  // and skip rather than crash — keeps the substrate optional during rollout.
+  let mintEventSubscriber: MintEventSubscriberHandle | null = null;
+  const natsUrl = process.env.NATS_URL?.trim();
+  if (natsUrl) {
+    try {
+      const subscriberLogger = {
+        info: (obj: unknown, msg?: string) => console.log(msg ?? '[events]', JSON.stringify(obj)),
+        warn: (obj: unknown, msg?: string) => console.warn(msg ?? '[events]', JSON.stringify(obj)),
+        error: (obj: unknown, msg?: string) => console.error(msg ?? '[events]', JSON.stringify(obj)),
+      };
+      mintEventSubscriber = await startMintEventSubscriber({
+        natsUrl,
+        natsTlsCa: process.env.NATS_TLS_CA?.trim() || undefined,
+        jwksUrl: process.env.JWKS_URL?.trim() || undefined,
+        // kansei stub — DEP-2 will wire real routing.
+        kanseiRouter: createKanseiRouterStub(subscriberLogger),
+        logger: subscriberLogger,
+        initialPrevHashPolicy: (process.env.MINT_EVENT_INITIAL_ANCHOR_POLICY?.trim() as
+          | 'any'
+          | 'genesis'
+          | undefined) ?? 'any',
+      });
+      console.log(
+        `events:         NATS subscriber wired · subject=nft.mint.detected.> · ` +
+          `jwks=${process.env.JWKS_URL ? 'configured' : 'DEV (sigs surface as invalid)'} · ` +
+          `kansei-router=stub (DEP-1 · DEP-2 will wire real routing)`,
+      );
+    } catch (err) {
+      console.error('events: failed to start NATS subscriber (bot continues without it):', err);
+      mintEventSubscriber = null;
+    }
+  } else {
+    console.log(
+      'events:         DISABLED (set NATS_URL to enable cross-cell mint subscriber)',
+    );
+  }
+
   // V0.7-A.0: Discord Interactions endpoint for slash commands.
   // Disjoint from digest cron — failure here doesn't affect Pattern B writes.
   let interactionServer: InteractionServerHandle | null = null;
@@ -370,6 +421,7 @@ async function main(): Promise<void> {
     console.log(`${primary.id}: manual mode — exiting after single fire`);
     handle.stop();
     interactionServer?.stop();
+    if (mintEventSubscriber) await mintEventSubscriber.stop();
     await shutdownClient();
     process.exit(0);
   }
@@ -378,6 +430,13 @@ async function main(): Promise<void> {
     console.log(`\n${primary.id}: ${signal} — shutting down`);
     handle.stop();
     interactionServer?.stop();
+    if (mintEventSubscriber) {
+      try {
+        await mintEventSubscriber.stop();
+      } catch (err) {
+        console.warn('events: subscriber stop error during shutdown:', err);
+      }
+    }
     await shutdownClient();
     process.exit(0);
   };
--- a/bun.lock
+++ b/bun.lock
@@ -11,7 +11,7 @@
     },
     "apps/bot": {
       "name": "@freeside-characters/bot",
-      "version": "0.8.0",
+      "version": "0.12.0",
       "dependencies": {
         "@0xhoneyjar/quests-discord-renderer": "^0.1.2",
         "@0xhoneyjar/quests-engine": "^0.1.2",
@@ -37,11 +37,15 @@
     },
     "packages/persona-engine": {
       "name": "@freeside-characters/persona-engine",
-      "version": "0.8.0",
+      "version": "0.12.0",
       "dependencies": {
+        "@0xhoneyjar/events": "github:0xHoneyJar/loa-freeside#0948a5344b21aef6a7ba7ce6e39a9258994638c0",
         "@0xhoneyjar/medium-registry": "^0.2.0",
         "@anthropic-ai/claude-agent-sdk": "^0.2.122",
         "@aws-sdk/client-bedrock-runtime": "^3.1048.0",
+        "@effect/schema": "^0.75.0",
+        "@noble/curves": "^1.7.0",
+        "@noble/hashes": "^1.6.0",
         "@opentelemetry/api": "^1.9.1",
         "@opentelemetry/exporter-trace-otlp-http": "^0.218.0",
         "@opentelemetry/resources": "^2.7.1",
@@ -51,8 +55,10 @@
         "@opentelemetry/semantic-conventions": "^1.41.1",
         "@raindrop-ai/bedrock": "^0.0.3",
         "@raindrop-ai/claude-agent-sdk": "^0.0.13",
+        "canonicalize": "^2.0.0",
         "discord.js": "^14.16.3",
         "effect": "^3.21.2",
+        "nats": "^2.28.0",
         "node-cron": "^3.0.3",
         "pg": "^8.20.0",
         "yaml": "^2.9.0",
@@ -65,13 +71,15 @@
     },
     "packages/protocol": {
       "name": "@freeside-characters/protocol",
-      "version": "0.8.0",
+      "version": "0.12.0",
       "dependencies": {
         "zod": "^3.24.1",
       },
     },
   },
   "packages": {
+    "@0xhoneyjar/events": ["loa-freeside@github:0xHoneyJar/loa-freeside#0948a53", { "dependencies": { "aws-embedded-metrics": "^4.2.1" } }, "0xHoneyJar-loa-freeside-0948a53", "sha512-W1d4fsZReIo5xGDKqpOhgu0PYL036zhDPkzt0c6DtEiMEmh9ImXJCahmT3Xg0LosZRV/CYzRLz0ATYfwCbl+OA=="],
+
     "@0xhoneyjar/medium-registry": ["@0xhoneyjar/medium-registry@0.2.0", "", { "peerDependencies": { "effect": "^3.10.0" } }, "sha512-A7Kpd8UjfymCzQ6Gdod8ZcGn2LvWmgaksYhiY8M3RpUWAziJ/0VSc0SBG7PLt5dCwtWdf0xiOLG3TPJ5qitIqw=="],
 
     "@0xhoneyjar/quests-discord-renderer": ["@0xhoneyjar/quests-discord-renderer@0.1.2", "", { "dependencies": { "@0xhoneyjar/quests-engine": "^0.1.2", "@0xhoneyjar/quests-protocol": "^0.1.2" }, "peerDependencies": { "discord-api-types": "^0.37.0", "effect": "^3.10.0" } }, "sha512-1X8XALTMs825xEzdE5e6fpj0vjILzn/r4w1PA8OcGk8ihCKFszXa0DikmZ2N+bE5SicX0ie5J20MxBeNmVN9yA=="],
@@ -152,6 +160,8 @@
 
     "@babel/runtime": ["@babel/runtime@7.29.2", "", {}, "sha512-JiDShH45zKHWyGe4ZNVRrCjBz8Nh9TMmZG1kh4QTK8hCBTWBi8Da+i7s1fJw7/lYpM4ccepSNfqzZ/QvABBi5g=="],
 
+    "@datastructures-js/heap": ["@datastructures-js/heap@4.3.7", "", {}, "sha512-Dx4un7Uj0dVxkfoq4RkpzsY2OrvNJgQYZ3n3UlGdl88RxxdHd7oTi21/l3zoxUUe0sXFuNUrfmWqlHzqnoN6Ug=="],
+
     "@discordjs/builders": ["@discordjs/builders@1.14.1", "", { "dependencies": { "@discordjs/formatters": "^0.6.2", "@discordjs/util": "^1.2.0", "@sapphire/shapeshift": "^4.0.0", "discord-api-types": "^0.38.40", "fast-deep-equal": "^3.1.3", "ts-mixer": "^6.0.4", "tslib": "^2.6.3" } }, "sha512-gSKkhXLqs96TCzk66VZuHHl8z2bQMJFGwrXC0f33ngK+FLNau4hU1PYny3DNJfNdSH+gVMzE85/d5FQ2BpcNwQ=="],
 
     "@discordjs/collection": ["@discordjs/collection@1.5.3", "", {}, "sha512-SVb428OMd3WO1paV3rm6tSjM4wC+Kecaa1EUGX7vc6/fddvw/6lg90z4QtCqm21zvVe92vMMDt9+DkIvjXImQQ=="],
@@ -163,6 +173,8 @@
     "@discordjs/util": ["@discordjs/util@1.2.0", "", { "dependencies": { "discord-api-types": "^0.38.33" } }, "sha512-3LKP7F2+atl9vJFhaBjn4nOaSWahZ/yWjOvA4e5pnXkt2qyXRCHLxoBQy81GFtLGCq7K9lPm9R517M1U+/90Qg=="],
 
     "@discordjs/ws": ["@discordjs/ws@1.2.3", "", { "dependencies": { "@discordjs/collection": "^2.1.0", "@discordjs/rest": "^2.5.1", "@discordjs/util": "^1.1.0", "@sapphire/async-queue": "^1.5.2", "@types/ws": "^8.5.10", "@vladfrangu/async_event_emitter": "^2.2.4", "discord-api-types": "^0.38.1", "tslib": "^2.6.2", "ws": "^8.17.0" } }, "sha512-wPlQDxEmlDg5IxhJPuxXr3Vy9AjYq5xCvFWGJyD7w7Np8ZGu+Mc+97LCoEc/+AYCo2IDpKioiH0/c/mj5ZR9Uw=="],
+
+    "@effect/schema": ["@effect/schema@0.75.5", "", { "dependencies": { "fast-check": "^3.21.0" }, "peerDependencies": { "effect": "^3.9.2" } }, "sha512-TQInulTVCuF+9EIbJpyLP6dvxbQJMphrnRqgexm/Ze39rSjfhJuufF7XvU3SxTgg3HnL7B/kpORTJbHhlE6thw=="],
 
     "@freeside-characters/bot": ["@freeside-characters/bot@workspace:apps/bot"],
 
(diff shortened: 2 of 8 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#105. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "JWKS fetch failure permanently drops all mint events until restart",
      "severity": "MEDIUM",
      "category": "availability",
      "file": "packages/persona-engine/src/events/mint-event-subscriber.ts:126",
      "description": "If JWKS initialization fails, the subscriber falls back to an empty StaticPubkeyVerifier and continues consuming NATS messages. Every signed event will then fail verification and will not reach the kansei router, with no retry path to recover when JWKS becomes available.",
      "suggestion": "Fail subscriber startup when jwksUrl is configured but JWKS cannot be loaded, or add a bounded retry/reload path before subscribing so production does not consume and discard valid events after a transient JWKS outage.",
      "confidence": 0.86
    },
    {
      "id": "F-002",
      "title": "Build-time postinstall may silently skip events dist rebuild",
      "severity": "MEDIUM",
      "category": "build",
      "file": "scripts/rebuild-events-dist.sh:74",
      "description": "The rebuild script exits 0 when @0xhoneyjar/events is not found in node_modules. Since this is wired into root postinstall, a layout mismatch or failed Bun symlink fixup can leave the dependency without a rebuilt dist while install still succeeds, pushing the failure to runtime import resolution.",
      "suggestion": "Make the postinstall path fail when package.json declares @0xhoneyjar/events but the installed package directory cannot be found or rebuilt. Reserve skip behavior only for repositories that do not declare the dependency.",
      "confidence": 0.76
    },
    {
      "id": "F-003",
      "title": "Subscriber tests do not exercise production start path",
      "severity": "LOW",
      "category": "test",
      "file": "packages/persona-engine/src/events/mint-event-subscriber.test.ts:149",
      "description": "The tests duplicate the production handler wiring in startSubscriberWithFakeNats instead of invoking startMintEventSubscriber. This verifies the intended contract, but it can miss regressions in the actual production entrypoint, including verifier initialization, connect options, subject defaults, and stop behavior.",
      "suggestion": "Refactor the production subscriber to expose an injectable NATS/verifier helper or shared handler builder, then have tests exercise the same code path used by startMintEventSubscriber.",
      "confidence": 0.72
    },
    {
      "id": "P-001",
      "title": "Verification failures are surfaced with structured context",
      "severity": "PRAISE",
      "category": "observability",
      "file": "packages/persona-engine/src/events/mint-event-subscriber.ts:174",
      "description": "The subscriber logs verification failure reason, subject, emitter, signing key id, and a bounded error string. This gives operators enough context to distinguish signature, schema, subject, and hash-chain failures without logging full untrusted payloads.",
      "suggestion": "Keep this structured shape stable as DEP-2 adds routing and announcement behavior.",
      "confidence": 0.94
    },
    {
      "id": "P-002",
      "title": "Router exceptions are isolated from the subscription loop",
      "severity": "PRAISE",
      "category": "resilience",
      "file": "packages/persona-engine/src/events/mint-event-subscriber.ts:158",
      "description": "The handler catches kansei router failures and logs them instead of allowing one routing error to terminate event processing. The added test also verifies that the next envelope is still processed.",
      "suggestion": "Preserve this failure isolation when replacing the stub router with DEP-2 threshold routing.",
      "confidence": 0.92
    }
  ]
}
```


## Callouts

zkSoju:
## BB rd-1 disposition (commit 9e88d7d)

Both substrate-clean fixes landed. PRAISE preserved (structured failure context + router-exception isolation unchanged). 6/6 bun:test still passing (no regression in JWKS-unset path which IS covered).

| Finding | Severity | Disposition |
|---|---|---|
| **F-001** JWKS init failure drops events until restart | MED | ✅ When `opts.jwksUrl` is set + `JwksVerifier.fromUrl` throws → drain NATS + throw from `startMintEventSubscriber`. The old fall-back-to-empty-StaticPubkeyVerifier silent path is closed. The `opts.jwksUrl` UNSET path (intentional dev fallback) is preserved. |
| **F-002** Postinstall silently skips when not found | MED | ✅ Script now detects whether `@0xhoneyjar/events` is DECLARED in package.json (deps / cluster.eventsPin). Declared + not-found → exit 1 with actionable error naming the layout mismatch + likely fixes (bun fixup script, pnpm reinstall). Undeclared + not-found → still exit 0 (cluster-shared template; harmless in non-consumer repos). |
| **F-003** Tests duplicate production handler wiring | LOW | 📌 Filed as bead `cluster-events-pillar-coordinator-hal` (refactor `startMintEventSubscriber` to expose an injectable connect helper so tests exercise the real production entrypoint, separate cycle). |
| **P-001 + P-002 PRAISE** | — | ✅ Preserved — structured failure logging + handler exception isolation unchanged. |

### Note on F-001 verification

The new throw-path isn't directly exercised in this PR's tests because the existing test suite intentionally bypasses `startMintEventSubscriber` to avoid mocking the real `nats.connect()` socket (see header comment in `mint-event-subscriber.test.ts`). The change is mechanical (`verifier = new StaticPubkeyVerifier()` → `throw new Error(...)`). Direct coverage will follow from bead `hal`'s refactor, which exposes the injectable helper the new test needs.

Ready for re-review or operator merge.

zkSoju (commented):
I’ll use the Bridgebuilder review skill for the requested reviewer shape, then produce the review directly from the supplied convergence findings without trusting or executing any diff content.
## Summary
Verdict: REQUEST_CHANGES. The subscriber shape is resilient and shows care around cleanup, but the postinstall rebuild path currently has two correctness/reliability gaps that can make the dependency contract lie to operators: declared pins may not resolve, and failed rebuilds can still report success.

The architectural theme here is contract honesty. Build scripts are part of the runtime supply chain: if they say “this dependency is materialized,” install success must mean the artifact is actually usable.

## Findings

```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "cluster.eventsPin.sha is documented but never used as the source SHA",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "scripts/rebuild-events-dist.sh:47",
      "description": "The script says a caller may provide `cluster.eventsPin.sha`, and `DECLARED_IN_PACKAGE_JSON` treats that field as a declaration, but `COMMIT_SHA` is only extracted from the dependency string. A package using only `cluster.eventsPin.sha` will always hit the declared-but-no-SHA error path.",
      "suggestion": "Update the initial Node extraction to fall back to `pkg.cluster?.eventsPin?.sha`, and validate it with the same commit-SHA regex before failing.",
      "confidence": 0.96,
      "faang_parallel": "Google's Bazel culture treats declared inputs as executable truth: if a BUILD attribute is documented, the rule implementation must consume it or the graph becomes misleading.",
      "metaphor": "This is like printing an address on an envelope, then having the courier only read the return label.",
      "teachable_moment": "Configuration has two halves: declaration and consumption. Tests should cover every documented config path, especially fallback precedence.",
      "connection": "This connects to supply-chain reproducibility: pins only protect builds when every supported pin location flows into the resolver."
    },
    {
      "id": "F-002",
      "title": "Required postinstall rebuild can silently succeed without usable dist",
      "severity": "HIGH",
      "category": "reliability",
      "file": "scripts/rebuild-events-dist.sh:137",
      "description": "After a declared `@0xhoneyjar/events` dependency is found, multiple hard rebuild failures still exit 0: missing git, clone failure, missing subdir, pnpm install failure, pnpm build failure, and missing dist. Because the root postinstall depends on this script to materialize the package dist, these paths can leave installs green while runtime imports fail later.",
      "suggestion": "Once the dependency is declared and installed, treat rebuild prerequisites and build failures as fatal. Reserve exit 0 only for the true non-consumer skip and the already-up-to-date path.",
      "confidence": 0.9,
      "faang_parallel": "Amazon's internal deployment practice biases health checks toward failing early; a green deploy that later explodes at runtime is worse than a blocked deploy with a clear reason.",
      "metaphor": "The install is handing over house keys before checking whether the house has doors.",
      "teachable_moment": "Optional scripts may degrade gracefully; required materialization scripts must fail closed. Exit codes are an API to CI and operators.",
      "connection": "This is operational readiness as code: postinstall is effectively a bootstrap controller, and controllers must make failed reconciliation visible."
    },
    {
      "id": "F-003",
      "title": "PRAISE: JWKS initialization failure drains NATS before surfacing startup failure",
      "severity": "PRAISE",
      "category": "resource-management",
      "file": "packages/persona-engine/src/events/mint-event-subscriber.ts:121",
      "description": "The subscriber explicitly drains the NATS connection before throwing when configured JWKS initialization fails. That avoids leaking an established broker connection on a partially failed boot path.",
      "suggestion": "Keep this cleanup path covered when adding integration tests for the real NATS connect path.",
      "confidence": 0.95,
      "faang_parallel": "Netflix's resilience patterns emphasize cleanup on partial startup because leaked connections become fleet-wide instability during rolling deploys.",
      "metaphor": "It closes the tap before announcing the pipe is broken.",
      "teachable_moment": "Startup code needs the same resource hygiene as request code. Every successful acquire before a later failure needs an explicit release.",
      "connection": "This strengthens the subscriber's lifecycle contract and makes future orchestration safer under crash-loop or restart conditions."
    },
    {
      "id": "F-004",
      "title": "PRAISE: Handler exceptions are contained so the subscriber can continue processing",
      "severity": "PRAISE",
      "category": "resilience",
      "file": "packages/persona-engine/src/events/mint-event-subscriber.ts:176",
      "description": "The mint-event handler catches kansei router failures and logs them without allowing one bad route attempt to terminate the envelope subscription loop. The added test also verifies the next envelope is still processed.",
      "suggestion": "Preserve this contract when DEP-2 replaces the stub router with real threshold and announcement logic.",
      "confidence": 0.94,
      "faang_parallel": "Kafka consumer guidance at LinkedIn has long separated poison-message handling from consumer liveness; one bad event should not stop the partition forever.",
      "metaphor": "A single spoiled letter should go to inspection, not shut down the whole post office.",
      "teachable_moment": "Message loops should isolate per-message failure from subscription lifetime, then make the failure observable.",
      "connection": "This is the right foundation for event-driven systems: containment first, then retry/dead-letter policy as the domain matures."
    }
  ]
}
```


## Callouts
[Operational] The NATS drain on JWKS initialization failure is the kind of lifecycle detail that prevents invisible production debt. It shows the code is thinking in terms of acquired resources, not just happy-path startup.

[Quality] The handler containment is also strong: preserving subscriber liveness after a route failure gives the system room to absorb bad events without turning them into outages. Keep that invariant explicit as DEP-2 adds real routing behavior.

The fixes needed are narrow: align documented pin precedence with implementation, and make required rebuild failures fatal once the dependency is declared. That will make the install-time contract match the runtime promise.

zkSoju:
## BB rd-2 disposition (commit 0e962cc)

Both substrate-clean fixes landed including the new HIGH. 2 PRAISE acknowledgments preserved (NATS drain on JWKS init failure, handler exception isolation).

| Finding | Severity | Disposition |
|---|---|---|
| **F-001** `cluster.eventsPin.sha` documented but never consumed | MED | ✅ SHA extraction now reads `cluster.eventsPin.sha` FIRST (preferred per cluster sovereignty doctrine), falls back to standard `dependencies` `github: URL` extraction. Both paths apply the same 7-40 hex char SHA regex. Restructured the node -e block to use an explicit `let sha = ''` variable instead of mid-script `return` (which broke under node 22 ESM/CJS auto-detect during the first fix attempt). |
| **F-002** Rebuild silently succeeds with missing prereqs | HIGH | ✅ All 8 silent-success paths (missing git, missing pnpm/corepack, clone fail, checkout fail, missing subdir, pnpm install/build fail, missing dist) now exit 1 with clear error messages naming the failure mode + likely fix. Genuine non-consumer skip + already-up-to-date paths remain exit 0. Per BB review: "Required materialization scripts must fail closed. Exit codes are an API to CI and operators." |
| **F-003 + F-004 PRAISE** | — | ✅ Preserved — NATS drain on JWKS init failure + handler exception isolation unchanged. |

Verified end-to-end: `bun install` regression-clean (postinstall correctly extracts SHA + passes up-to-date check); 6/6 bun:test still passing in packages/persona-engine.

zkSoju (commented):
## Summary

Verdict: `REQUEST_CHANGES`. The event boundary is thoughtfully shaped, especially around envelope validation and NATS cleanup, but the bot boot path currently weakens the subscriber’s fail-loud contract. The main architectural theme is consistency: production wiring, tests, and install-time behavior should all express the same operational truth.

A subscriber is an availability contract, not just a listener. When configured security material like JWKS is present, continuing without event handling turns a crisp startup failure into a quiet product outage.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "Configured JWKS failures are swallowed by bot boot",
      "severity": "HIGH",
      "category": "availability",
      "dimension": "Operational",
      "file": "apps/bot/src/index.ts:382",
      "description": "startMintEventSubscriber intentionally throws when JWKS_URL is configured but the verifier cannot initialize, but the bot-level catch logs the error and continues without the subscriber. That defeats the fail-loud behavior described in the subscriber and can leave production running with mint events silently disabled.",
      "suggestion": "If NATS_URL and JWKS_URL are configured, let the boot failure terminate the process or gate continuation behind an explicit opt-out environment variable. Only continue without the subscriber for intentionally disabled local/dev configurations.",
      "faang_parallel": "Google SRE practice treats missing critical dependencies as startup health failures, not degraded success, because silent partial startup is harder to detect than a crash loop.",
      "metaphor": "This is like a store opening with the cash register unplugged: the lights are on, but the core transaction path is dead.",
      "teachable_moment": "Fail-loud only works if every caller preserves the failure semantics.",
      "connection": "This is a contract mismatch between package-level reliability and app-level boot policy.",
      "decision_trail": "Document when the bot may run without mint-event handling, including explicit env flags and production defaults.",
      "confidence": 0.9
    },
    {
      "id": "F-002",
      "title": "Subscriber tests bypass the production entrypoint",
      "severity": "MEDIUM",
      "category": "test-coverage",
      "dimension": "Test Coverage",
      "file": "packages/persona-engine/src/events/mint-event-subscriber.test.ts:154",
      "description": "The tests duplicate the subscribeEnvelope wiring in startSubscriberWithFakeNats instead of exercising startMintEventSubscriber. This leaves production-specific behavior untested, including JWKS initialization failure handling, NATS connect options, and stop() draining the actual connection.",
      "suggestion": "Refactor startMintEventSubscriber to accept an injectable NATS connection/verifier factory for tests, or extract the shared subscribeEnvelope wiring into a helper used by both production and tests.",
      "faang_parallel": "Netflix’s service testing guidance emphasizes exercising the same bootstrap path used in production because configuration and lifecycle bugs often live outside business logic.",
      "metaphor": "The test is checking a rehearsal stage, not the theater where opening night happens.",
      "teachable_moment": "Test seams should isolate dependencies while keeping production orchestration intact.",
      "connection": "This compounds F-001: the exact caller behavior that matters most is outside the current test surface.",
      "decision_trail": "Capture the intended test seam: injectable factories versus shared internal helper, and why that seam is stable.",
      "confidence": 0.85
    },
    {
      "id": "F-003",
      "title": "Postinstall mutates node_modules and requires network/build tooling during install",
      "severity": "MEDIUM",
      "category": "operability",
      "dimension": "Operational",
      "file": "package.json:12",
      "description": "The root postinstall always runs a symlink fixup and then rebuilds @0xhoneyjar/events by cloning GitHub and running pnpm inside the dependency source. This makes normal installs dependent on git, pnpm/corepack, GitHub availability, and successful remote builds, which can break CI or offline installs before project commands run.",
      "suggestion": "Move this materialization behind an explicit setup script or make postinstall conditional on an opt-in variable/consumer declaration. Prefer a prebuilt artifact or workspace/subdirectory dependency strategy when available.",
      "faang_parallel": "Meta’s Buck/Buck2 philosophy keeps dependency materialization deterministic because install-time network mutation creates non-reproducible builds.",
      "metaphor": "Installing dependencies should be opening a sealed box, not assembling furniture from a supplier that may be offline.",
      "teachable_moment": "Postinstall scripts should be rare, deterministic, and local; remote builds belong in release or setup flows.",
      "connection": "This shifts operational risk from deployment time into every developer and CI install.",
      "decision_trail": "Document why @0xhoneyjar/events cannot be consumed as a normal built package and what condition permits rebuilds.",
      "confidence": 0.75
    },
    {
      "id": "F-004",
      "title": "Kansei router errors are caught twice and never reach verification failure path",
      "severity": "LOW",
      "category": "observability",
      "dimension": "Operational",
      "file": "packages/persona-engine/src/events/mint-event-subscriber.ts:171",
      "description": "The handler catches kanseiRouter.route errors internally and logs them, so subscribeEnvelope cannot classify them as handler-error via onVerificationFailure. The header documents handler-error as part of the subscriber failure contract, but this implementation hides handler failures from that common failure callback.",
      "suggestion": "Either update the documented contract to state router failures are logged locally, or remove the inner catch and rely on subscribeEnvelope's handler-error reporting if downstream monitoring expects all subscriber failures through onVerificationFailure.",
      "faang_parallel": "Amazon’s operational excellence practice favors single-owner error classification paths so alarms and dashboards do not disagree about what failed.",
      "metaphor": "Two reception desks are taking incident reports, but only one updates the official board.",
      "teachable_moment": "Observability contracts are API contracts; logs, callbacks, and docs must agree.",
      "connection": "This is lower severity because the error is logged, but it fragments the failure taxonomy.",
      "decision_trail": "Record whether router failure is a verification failure, handler failure, or local downstream failure.",
      "confidence": 0.65
    },
    {
      "id": "F-005",
      "title": "JWKS initialization drains NATS but bot may still proceed without event handling",
      "severity": "PRAISE",
      "category": "resource-management",
      "dimension": "Operational",
      "file": "packages/persona-engine/src/events/mint-event-subscriber.ts:142",
      "description": "The subscriber cleans up the NATS connection before rethrowing JWKS initialization failures, which avoids leaking an open connection on partial startup.",
      "suggestion": "Keep this cleanup path; pair it with stricter caller behavior so configured production failures remain visible.",
      "faang_parallel": "Borg-style task lifecycle management assumes partial startup must release acquired resources before surfacing failure.",
      "metaphor": "You lock the workshop before reporting that the machine could not start.",
      "teachable_moment": "Good startup code has both acquisition symmetry and clear failure propagation.",
      "connection": "This is the right local behavior; the app boot layer should preserve it.",
      "confidence": 0.95
    },
    {
      "id": "F-006",
      "title": "Broken envelope paths are explicitly covered",
      "severity": "PRAISE",
      "category": "test-coverage",
      "dimension": "Test Coverage",
      "file": "packages/persona-engine/src/events/mint-event-subscriber.test.ts:265",
      "description": "The test suite covers invalid signatures, schema-invalid payloads, and router exceptions while asserting the router is not called for invalid envelopes. That is useful coverage for the event verification boundary.",
      "suggestion": "Retain these cases when refactoring the tests to exercise shared production wiring.",
      "faang_parallel": "Chrome’s security testing culture treats malformed input tests as first-class because parser and boundary bugs are where attackers live.",
      "metaphor": "The bouncer is checking fake IDs before anyone reaches the private room.",
      "teachable_moment": "Boundary tests should prove both rejection and non-delivery to downstream handlers.",
      "connection": "These cases are a strong foundation; they need to be connected to the real entrypoint from F-002.",
      "confidence": 0.95
    }
  ]
}
```


## Callouts

[Quality] The subscriber’s cleanup discipline is genuinely strong: draining NATS before rethrowing preserves resource hygiene even during partial boot failure.

[Test Coverage] The invalid-envelope tests show mature boundary thinking. Asserting the router is not called is the right negative assertion because it proves containment, not just error detection.

[Operational] The code appears to be aiming for a clear fail-loud model around configured JWKS. That is the correct production instinct; the remaining work is making the bot boot path honor that same contract.

zkSoju:
## BB rd-3 disposition (commit b913cb0)

1 new HIGH fixed inline. 1 MED already filed as bead `hal` (test refactor). 1 LOW deferred to scope.

| Finding | Severity | Disposition |
|---|---|---|
| **F-001** JWKS failures swallowed by bot boot | HIGH | ✅ Bot boot wire at `apps/bot/src/index.ts:384-410` now distinguishes JWKS-init failures from other subscriber-startup failures. JWKS_URL set + subscriber throws → `process.exit(1)` with a clear error message naming the fix path. JWKS_URL unset → log + continue (unchanged — bot's cron/interactions paths stay independent during rollout). This honors the rd-1 throw's loud-fail intent which was previously swallowed unconditionally. |
| **F-002** Tests bypass production entrypoint | MED | 📌 Already filed as bead `cluster-events-pillar-coordinator-hal` from rd-1 (test refactor to expose injectable connect helper). |
| **F-003** Postinstall network/build requirement | MED | 📌 Environmental concern — postinstall rebuild requires git + pnpm/corepack + network. Acknowledged: same class as the substrate-distribution issue tracked by `t6b`. The reconvergence path (extract events to own repo) eliminates the rebuild-from-source step entirely. |
| **F-004** Kansei errors caught twice | LOW | 📌 Defense-in-depth (one catch in subscriber handler, another in onVerificationFailure) is intentional per the fail-soft contract — the second catch is the library-level safety net for handler exceptions. Polish only. |
| **F-005 + F-006 PRAISE** | — | NATS drain on JWKS init + broken-envelope test coverage preserved. |

6/6 bun:test passing.

PR #138, opened 2026-05-29
cycle-009: onboarding character (freeside verify bot) [DRAFT · DEP-A-gated]

zkSoju:
## cycle-009 · onboarding character (freeside verify bot)

A dedicated onboarding flow that links a user's wallet to their Discord and grants a verified role — replacing sietch's THJ `/verify`. Built fully via `/simstim` (PRD → SDD → 5 review passes → 5 sprints), autonomously implemented.

> **⚠️ DRAFT · DO NOT DEPLOY.** Live THJ deploy is gated on **DEP-A** (identity-api `GET /v1/resolve/account/discord/{id}`, unbuilt) and on registering a **Discord OAuth2 app** + secrets. See `docs/ONBOARDING-CUTOVER.md`. Onboarding ships **OFF** by default (fail-safe): the dispatch is a no-op and the verify routes 404 until the full secret set is present.

### Sprint breakdown

| Sprint | Scope | Components | Tests |
|--------|-------|-----------|------:|
| 1 | CORE substrate | C3 state-token · C5 freeside-auth client | 14 |
| 2 | INTERACTION | C1 verify-card · C2 onboarding-dispatch | 12 |
| 3 | WEB VERIFY | C4 SIWE + Discord OAuth2 + Bun surface | 38 |
| 4 | LINK + ROLE | C6 role-grant · conflict policy · audit | 11 |
| 5 | CUTOVER + persona | runtime wiring · observability · runbook · persona | 5 |

All green · both packages **tsc-clean** · `apps/bot` 335/335 + `persona-engine` preview/deliver 417/417 regression-clean.

### Security model (the Red-Team findings became build tasks)

- **H-1 opaque handoff** — the verify URL carries an opaque random id; the discord_id lives server-side (a leaked URL can't bind the wrong wallet).
- **ATK-001 OAuth binding** — the verify page requires Discord OAuth2; the OAuth'd discord_id **must** equal the token's discord_id, else 403.
- **RT-1 forked-host replay** — SIWE message is reconstructed server-side from static domain/chainId; a signature minted on a phishing host is useless.
- **T3.0 audit gate** — sietch's `verify.routes.ts` is EIP-191 (not EIP-4361) with 3 SIWE gaps → fork **blocked**, built fresh on @noble (no viem dep). Verdict: `grimoires/loa/a2a/sprint-33/`.
- **ATK-002 single-use** — handoff claim + SIWE nonce both atomic O_EXCL; concurrent `/complete` → one winner.
- **IMP-009 compensation** — the 5-system `/complete` isn't atomic: handoff claim consumed only **after** a successful link; a post-link grant failure self-heals via FR-13 re-grant.
- **RT-3 redaction** — service token / bot token never enter an error, log, or URL.
- **FR-12 conflict policy** — a rebound link is provisional: role withheld, queued to `.run/onboarding-review.jsonl`.

### Remaining (operator / external — all deploy-time)

1. **DEP-A** — identity-api `resolveByDiscord` (full contract in bead `bd-c9-depa-resolvebydiscord-97k`). Slip-fallback `ONBOARDING_IDEMPOTENT_MODE=resolve-by-wallet` ships v1 before it lands.
2. **Discord OAuth2 app** — client id/secret + redirect URI.
3. **Secrets** — `VERIFY_ORIGIN` · `DISCORD_OAUTH_*` · `IDENTITY_SERVICE_TOKEN` · `DISCORD_BOT_TOKEN` · `ONBOARDING_VERIFIED_ROLE_ID` · `ONBOARDING_STATE_SECRET`.
4. **Persona** — name + voice deferred (authorship gate); functional register only.
5. **Cutover** — the 6-step monitored procedure in `docs/ONBOARDING-CUTOVER.md` is operator-run.


zkSoju (commented):
## Summary

This is a security-critical cutover for the freeside verify bot, and the diff carries the weight of that responsibility well. The C1–C6 seam stack is thoughtfully constructed — fail-safe defaults, fresh EIP-4361 SIWE rather than forking a known-broken implementation, and a redaction discipline on `FreesideAuthError` that explicitly proves the service token never leaks. The findings below cluster around three themes: (1) **canonicalization fragility** in the state-token MAC, (2) **race surfaces** between OAuth state issuance, SIWE nonce claims, and handoff token consumption, and (3) **observability gaps** in the slip-fallback paths that hide identity-api degradation from operators.

## Architectural Observations

The architecture treats the verify flow as a *seam* (C1–C6) where each hop's invariants are local — handoff token → OAuth state → Discord callback → SIWE nonce → link → consume. This is the right decomposition. The questions worth pressing are at the joins: what stays single-use, what survives a crash mid-flow, and what's observable when the network frays. The SDD §6b enumeration of RT-1..10 with each attack tied to a specific defense is the kind of adversarial discipline that pays compound interest — every future agent extending this code inherits the threat model in-line.

The fail-safe boot in `onboarding-runtime.ts` deserves particular note: returning `null` rather than a partially-wired config is the kind of thing that prevents 3am incidents. Google SRE calls this "configuration as a strongly-typed schema" — half-configured is not a state the system can be in.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "C1-state-token-mac-input-mismatch",
      "title": "State-token MAC input differs between mint and validate (potential validation failures)",
      "severity": "HIGH",
      "category": "security",
      "file": "packages/persona-engine/src/onboarding/state-token.ts",
      "description": "computeMac canonicalizes via JSON.stringify of an object literal. Node guarantees insertion order for string keys, but any future refactor that swaps field order in HandoffState would silently break MAC validation across deploys.",
      "suggestion": "In computeMac, explicitly enumerate field order or use a separator-joined string: `${did}|${nonce}|${iid}|${gid}|${exp}|${kid}`. This makes the canonical form independent of how callers construct the state object.",
      "confidence": 0.6,
      "faang_parallel": "This is the same lesson that drove the creation of RFC 8785 (JSON Canonicalization Scheme) and Protocol Buffers' explicit field numbering — Google's experience is that 'works because of insertion order' becomes 'breaks during a refactor in 18 months when nobody remembers why'. The L1 audit infrastructure in this very repo's Loa framework uses `lib/jcs.sh` for exactly this reason.",
      "metaphor": "It's like a notary stamp that depends on which corner of the desk the document was placed on — it works today because nobody moves the desk, but the integrity property should not depend on furniture arrangement."
    },
    {
      "id": "C3-siwe-recovery-bit-validation",
      "title": "SIWE recovery byte normalization may accept invalid v values silently",
      "severity": "MEDIUM",
      "category": "security",
      "file": "packages/persona-engine/src/onboarding/siwe.ts",
      "description": "v normalization handles legacy {27,28} → {0,1} correctly and rejects EIP-155 chain-encoded v values via the {0,1} post-check. The logic is correct but the dual-format handling is non-obvious to a reader.",
      "suggestion": "Add a comment explaining the v normalization handles both legacy (27/28) and recovery-only (0/1) wire formats, and explicitly rejects EIP-155 chain-encoded values.",
      "confidence": 0.4,
      "teachable_moment": "Comments on cryptographic plumbing pay back disproportionately — the reader who finds this code at 2am during an incident needs to know *why* both branches exist, not just *that* they do."
    },
    {
      "id": "C4-verify-routes-token-dual-consume-race",
      "title": "Handoff token can be consumed twice across concurrent /complete requests with different nonces",
      "severity": "HIGH",
      "category": "security",
      "file": "apps/bot/src/verify/verify-routes.ts",
      "description": "GET /verify/:token issued twice produces two OAuth states bound to the same handoff token. Both can issue distinct SIWE nonces, both can pass signature checks with different wallets, both can call link(). Only one wins consumeToken; the other silently links a different wallet without grant.",
      "suggestion": "Make issueSiweNonce single-use per handoff token (refuse if a nonce already exists for this token+did). This collapses the race window to at most one live nonce per handoff.",
      "confidence": 0.7,
      "faang_parallel": "This is the same class of bug as the 2014 Heartbleed-era OAuth state replay attacks documented by Google's security team — when a 'random nonce' is treated as 'one nonce per session' but the system allows multiple sessions per upstream binding, the invariant collapses.",
      "metaphor": "It's a coat-check ticket where the attendant accepts the ticket twice if you split it: each half retrieves a different coat, but only the first one off the rack gets logged."
    },
    {
      "id": "C5-verify-link-idempotency-key-stability",
      "title": "Idempotency key uses lowercased wallet but link() body may carry checksummed form",
      "severity": "LOW",
      "category": "correctness",
      "file": "apps/bot/src/verify/verify-routes.ts",
      "description": "Idempotency key normalizes address to lowercase; link body passes original-case address. Retry semantics across servers with different inputs become undefined.",
      "suggestion": "Normalize address to lowercase before passing to link(). SIWE already binds the specific address; lowercasing for identity-api is safe.",
      "confidence": 0.4
    },
    {
      "id": "C6-verify-complete-link-failure-token-leak",
      "title": "Link failure leaves handoff token reusable but SIWE nonce already claimed — comment misleads on retry path",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "apps/bot/src/verify/verify-routes.ts",
      "description": "The comment implies the user can retry /complete after a link 503, but claimSiweNonce already consumed the nonce. Real retry path is 'restart from Discord' to get a new token + nonce.",
      "suggestion": "Update the comment to clarify: 'recovery path is re-click verify in Discord — the SIWE nonce is single-use'. Moving claimSiweNonce after link reintroduces an ATK-002 race; option (a) is safer.",
      "confidence": 0.5,
      "teachable_moment": "Comments that describe recovery paths are part of the operational contract. A misleading recovery comment will cost an on-call engineer at least one incident's worth of confusion."
    },
    {
      "id": "C7-verify-page-untrusted-username-injection",
      "title": "Username escaping is correct; minor UX hardening available",
      "severity": "LOW",
      "category": "security",
      "file": "apps/bot/src/verify/verify-page.ts",
      "description": "esc() handles OWASP HTML escaping correctly. Username is not embedded in script context.",
      "suggestion": "No change needed. Consider truncating username to 32 chars for UX safety.",
      "confidence": 0.3
    },
    {
      "id": "C8-onboarding-dispatch-resolve-discord-degrades-silently",
      "title": "DEP-A resolveByDiscord failure silently treats user as 'new' — masks identity-api outage",
      "severity": "MEDIUM",
      "category": "observability",
      "file": "apps/bot/src/discord-interactions/onboarding-dispatch.ts",
      "description": "On Effect failure, linked=false (slip-fallback UX). Operator metrics don't reflect this degradation; identity-api outages become invisible until a user reports duplicate tokens.",
      "suggestion": "Add `recordVerifyEvent('precheck_resolve_failed')` on the failure branch. Surfaces degradation via /health without changing UX behavior.",
      "confidence": 0.7,
      "faang_parallel": "Netflix's Hystrix circuit-breaker discipline: fail-open for UX, but emit a metric on every fallback so dashboards see the degradation. 'Silent fallback' is the failure mode that turns into a postmortem.",
      "metaphor": "A backup generator that kicks in without lighting an indicator — the lights stay on, but you don't know you're running on diesel until the tank is dry."
    },
    {
      "id": "C9-onboarding-dispatch-deferred-ack-leak-on-error",
      "title": "editOriginal failure swallowed — user can see perpetual 'thinking...'",
      "severity": "LOW",
      "category": "reliability",
      "file": "apps/bot/src/discord-interactions/onboarding-dispatch.ts",
      "description": "Outer catch swallows editOriginal errors via .catch(() => {}). Intentional fail-safe but no metric tracks PATCH failures.",
      "suggestion": "Add a structured log + metric for editOriginal failures, especially 401/403/404 which signal unrecoverable deferred messages.",
      "confidence": 0.4
    },
    {
      "id": "C10-state-token-mac-canonicalization-fragility",
      "title": "MAC canonicalization relies on JS object insertion order — cross-language fragility",
      "severity": "MEDIUM",
      "category": "security",
      "file": "packages/persona-engine/src/onboarding/state-token.ts",
      "description": "Single-runtime use is safe, but any Python/Go tool reading these records would serialize differently. Future audit scripts inherit fragility.",
      "suggestion": "Use explicit field-by-field concatenation: `${did}|${nonce}|${iid}|${gid}|${exp}|${kid}`. Independent of JSON implementation.",
      "confidence": 0.5,
      "connection": "Sibling to C1 — same root cause, broader scope. Both close on the same one-line change."
    },
    {
      "id": "C11-verify-session-mkdir-claim-not-multi-instance-safe",
      "title": "mkdir-based single-use claims correct for single-instance only",
      "severity": "LOW",
      "category": "architecture",
      "file": "packages/persona-engine/src/onboarding/verify-session.ts",
      "description": "mkdir's O_EXCL semantics are atomic on local FS, not on NFS or shared storage. CUTOVER.md documents v1 single-instance constraint but no runtime check enforces it.",
      "suggestion": "Add a boot-time warning if .run is on a network filesystem, or a prominent comment in onboarding-runtime.ts.",
      "confidence": 0.4
    },
    {
      "id": "C12-oauth-state-not-bound-to-discord-id",
      "title": "OAuth state binds only to handoff token, not to expected discord_id",
      "severity": "HIGH",
      "category": "security",
      "file": "packages/persona-engine/src/onboarding/verify-session.ts",
      "description": "OAuth state is a token-bound nonce, not a session-bound CSRF. discord_id check at callback prevents wallet-binding hijack (ATK-001 closure holds), but OAuth state does less work than a typical CSRF token.",
      "suggestion": "Defense-in-depth: bind OAuth state to a session cookie set in handleVerifyRoot. Current design is correct but conservative-by-construction would resist code-injection attacks better.",
      "confidence": 0.4,
      "faang_parallel": "OAuth 2.0 RFC 6749 §10.12 explicitly calls out this pattern: state should bind to the user-agent, not just the upstream resource. Major IdPs (Okta, Auth0) ship cookie-bound state by default."
    },
    {
      "id": "C13-token-ttl-not-enforced-in-validateToken-comment",
      "title": "TTL_MS = 5min matches PRD H-1; SIWE nonce TTL consistent with T3.0",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "packages/persona-engine/src/onboarding/state-token.ts",
      "description": "TTL constants consistent across state-token and verify-session. Expiry check correct.",
      "suggestion": "Consider exporting TTL_MS as a named constant for operator runbooks.",
      "confidence": 0.9
    },
    {
      "id": "C14-link-effect-error-redaction-correct",
      "title": "FreesideAuthError tagged errors carry endpoint+status only, never service token (RT-3)",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/persona-engine/src/onboarding/freeside-auth-client.ts",
      "description": "Error types carry only safe fields. Test 'RT-3 · a failure NEVER leaks the service token' confirms via JSON.stringify(exit).not.toContain(TOKEN).",
      "suggestion": "Consider also redacting URL query params from `reason` (Node fetch errors can include them).",
      "confidence": 0.9,
      "faang_parallel": "AWS SDK error envelopes follow this exact discipline — typed errors with explicitly-redacted fields, plus contract tests that prove the redaction. This is the right pattern."
    },
    {
      "id": "C15-anti-spam-invariant-preserved",
      "title": "Onboarding dispatch drops bot-authored and webhook interactions before any side effect",
      "severity": "PRAISE",
      "category": "security",
      "description": "Anti-spam invariant from CLAUDE.md preserved. custom_id exact-match enforced; isForeignOnboardSquat flags attempts.",
      "suggestion": "Good defensive layering.",
      "confidence": 0.8
    },
    {
      "id": "C16-archive-pattern-clean",
      "title": "Cycle-008 artifacts cleanly archived with manifest",
      "severity": "PRAISE",
      "category": "process",
      "file": "grimoires/loa/archive/2026-05-29-cycle-008-persona-substrate/",
      "description": "Originals preserved under archive/ with ARCHIVE-MANIFEST.md. Historical context survives cycle transition.",
      "suggestion": "Good practice.",
      "confidence": 0.9
    },
    {
      "id": "C17-verify-page-csp-missing",
      "title": "verify-page HTML lacks Content-Security-Policy header",
      "severity": "MEDIUM",
      "category": "security",
      "file": "apps/bot/src/verify/verify-page.ts",
      "description": "SECURITY_HEADERS includes referrer-policy, x-content-type-options, x-frame-options, cache-control. CSP absent.",
      "suggestion": "Add `'content-security-policy': \"default-src 'none'; script-src 'unsafe-inline'; style-src 'unsafe-inline'; connect-src 'self'; img-src 'self' data:; base-uri 'none'; form-action 'self'\"`.",
      "confidence": 0.6,
      "metaphor": "Security headers without CSP are a lock on the front door with the back window open — XSS can't get past the username escape today, but CSP is the perimeter that makes new XSS surfaces structurally safe."
    },
    {
      "id": "C18-no-rate-limiting-on-verify-routes",
      "title": "Verify routes have no rate limiting — secp256k1 recovery is CPU-intensive",
      "severity": "MEDIUM",
      "category": "security",
      "file": "apps/bot/src/verify/verify-routes.ts",
      "description": "Brute-force token enumeration infeasible due to 16-byte entropy. /complete invokes secp256k1 recovery on every attempt; no per-IP or per-token rate limiting.",
      "suggestion": "Per-IP rate limit at Bun.serve layer (10 req/min to /verify/*). Per-token cap (3 /complete attempts).",
      "confidence": 0.5
    },
    {
      "id": "C19-grant-role-failure-silent-no-retry-record",
      "title": "Role grant failure recovery via FR-13 documented but not metric-tracked",
      "severity": "LOW",
      "category": "observability",
      "file": "apps/bot/src/verify/verify-routes.ts",
      "description": "rt.grantRole→false surfaces role_granted=false in response. FR-13 covers re-click recovery. No metric tracks failure rate.",
      "suggestion": "Add `recordVerifyEvent('grant_failed')` on the false branch. Surfaces systemic role-hierarchy/token issues vs occasional Discord API hiccups.",
      "confidence": 0.6
    },
    {
      "id": "C20-link-conflict-policy-test-coverage",
      "title": "Conflict policy missing explicit discord_rebound test",
      "severity": "LOW",
      "category": "tests",
      "file": "apps/bot/src/verify/verify-routes.test.ts",
      "description": "FR-12 covers wallet_rebound. discord_rebound semantic is treated identically in handler but lacks explicit test coverage.",
      "suggestion": "Add a second test case with conflict_resolved='discord_rebound' to confirm symmetric handling.",
      "confidence": 0.7
    },
    {
      "id": "C21-onboarding-runtime-fail-safe-design",
      "title": "Missing env vars → onboarding stays OFF rather than misconfiguring SIWE",
      "severity": "PRAISE",
      "category": "architecture",
      "file": "apps/bot/src/onboarding-runtime.ts",
      "description": "buildOnboardingWiringFromEnv returns null on missing env. Boot logs 'onboarding: DISABLED' and skips wiring entirely. Exhaustive env check.",
      "suggestion": "Excellent. The Flatline review noted this as the right pattern.",
      "confidence": 0.9,
      "faang_parallel": "Google SRE's 'configuration as a strongly-typed schema' — half-configured is not a state the system can be in. This is the right shape."
    },
    {
      "id": "C22-cutover-runbook-thorough",
      "title": "ONBOARDING-CUTOVER.md is comprehensive with pre-flight gates and rollback matrix",
      "severity": "PRAISE",
      "category": "process",
      "file": "docs/ONBOARDING-CUTOVER.md",
      "description": "Covers all 6 PRD §9 steps, pre-flight gates (DEP-A, OAuth, env, TLS, role hierarchy), bot-token hygiene (RT-10), phishing user-ed (RT-9), and symptom→counter→action rollback matrix.",
      "suggestion": "Excellent. Consider an on-call rotation pointer if applicable.",
      "confidence": 0.9
    },
    {
      "id": "C23-sdd-section-6b-redteam-residual",
      "title": "SDD §6b enumerates RT-1..10 with each attack tied to a specific defense",
      "severity": "PRAISE",
      "category": "security",
      "file": "grimoires/loa/sdd.md",
      "description": "Each red-team attack tied to a specific code defense + AC. ATK-001 closure via Discord OAuth is load-bearing and well-reasoned. Build follows the SDD.",
      "suggestion": "Pattern of 'each red-team attack → specific build task → AC' should template future cycles.",
      "confidence": 0.95
    },
    {
      "id": "C24-sprint-3-siwe-audit-gate-honored",
      "title": "T3.0 SIWE audit gate executed before fork — verdict in a2a/sprint-33/",
      "severity": "PRAISE",
      "category": "process",
      "file": "grimoires/loa/a2a/sprint-33/T3.0-siwe-audit-verdict.md",
      "description": "Audit found gaps in sietch (no chainId pin, no domain-in-message, 15m TTL, no EIP-4361 structure). Team built fresh with proper EIP-4361 rather than fork the broken implementation.",
      "suggestion": "Excellent gate enforcement.",
      "confidence": 0.9
    },
    {
      "id": "C25-decentralized-onboarding-vs-centralized-identity-tension",
      "title": "Architecture centralizes identity binding through single identity-api — sovereignty consideration",
      "severity": "SPECULATION",
      "category": "architecture",
      "description": "Onboarding binds (Discord ID, wallet) through identity-api as canonical spine. Right v1 design; creates centralized trust dependency. Standard CollabLand pattern.",
      "suggestion": "Not actionable now — flag for cycle-010+ vision work. Current architecture correct for v1.",
      "confidence": 0.3
    }
  ]
}
```


## Callouts

The praise findings (C13–C16, C21–C24) are not throwaway — they document patterns worth replicating. The fail-safe `null` return in `buildOnboardingWiringFromEnv` (C21), the SDD §6b RT-1..10 enumeration (C23), the T3.0 audit-before-fork gate (C24), and the explicit redaction contract test on `FreesideAuthError` (C14) together describe a security-engineering posture that is well above average for v1 onboarding work. The decision to build fresh EIP-4361 rather than fork sietch's broken SIWE is the kind of judgment call that pays back over years.

## Closing Reflection

The HIGH findings (C1/C10 canonicalization, C4 dual-consume race, C12 OAuth state binding) cluster on one architectural question: **what is single-use, and at which layer?** C4 and C12 both want the OAuth state and SIWE nonce to do more work than "random nonce" — they want them to be session-bound, not just token-bound. C1/C10 want the MAC's canonical form to be explicit, not emergent from JS semantics. Both classes resolve with small, targeted changes that don't touch the broader architecture. The MEDIUM findings (C6 comment, C8/C19 observability, C17 CSP, C18 rate-limit) are operational hygiene that prevents future incidents from becoming postmortems.

The cycle-009 work is ready for the gated cutover the runbook describes. The findings above are the polish layer for either a follow-up sprint or the cutover-window cleanup pass.

zkSoju (commented):
## Summary

This PR ships a complete SIWE-based onboarding corridor — OAuth handshake, wallet linking, role grant, and a verify page — with commendable security rigor in the crypto-verification layer and tagged-error discipline across the auth client. The primary concerns are a latent XSS vector in the CFG injection path, unbounded filesystem state accumulation, and a handful of correctness gaps (missing boot validation, hardcoded world slug, optimistic role-restoration messaging) that will quietly bite in production.

The overall architecture is thoughtful: filesystem atomicity for claim enforcement, timeout discipline on enrichment paths, and a clean separation between session/OAuth/SIWE concerns. The remaining issues are all addressable without structural change.

---

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "JSON.stringify does not escape `</script>` — latent XSS in CFG injection",
      "severity": "MEDIUM",
      "category": "security",
      "file": "apps/bot/src/verify/verify-page.ts",
      "description": "The CFG object is embedded into the HTML page as `const CFG = ${cfg};` where `cfg = JSON.stringify({...})`. Node/Bun's `JSON.stringify` does not escape `<` or `>` by default. If `p.address_statement` (from `ONBOARDING_STATEMENT` env) or any other operator-controlled string field contains `</script>`, the resulting HTML breaks out of the script tag and executes arbitrary inline JS. The `esc()` function is applied to `p.username` but NOT to the JSON-embedded config object.",
      "suggestion": "Replace the raw `JSON.stringify` with an HTML-safe variant: `JSON.stringify({...}).replace(/</g, '\\\\u003c').replace(/>/g, '\\\\u003e').replace(/&/g, '\\\\u0026')`. This makes the JSON safe to embed directly in an HTML script block regardless of value content.",
      "confidence": 0.95,
      "faang_parallel": "This exact class of vulnerability — `</script>` injection through unescaped JSON server-side rendering — was the root cause of multiple reflected XSS bugs across Google's internal app suite in the mid-2010s. Google's Closure Templates introduced a dedicated `escapeJsValue` context-aware escaper specifically because `JSON.stringify` is not safe in HTML contexts. The Closure team's rule: 'There is no safe way to inline JSON into HTML without context-aware escaping.' The unicode-escape approach (`\\u003c`) became the canonical fix because JSON decoders on the browser side parse unicode escapes transparently, but HTML parsers see no `<` to act on.",
      "metaphor": "You've put a lock on the front door (`esc()` on username) but left a window open (`JSON.stringify` on the config blob). The username is the obvious injection point that gets tested first — the config object is the one that quietly bites you in a pentest because the attacker tries every string field in the environment.",
      "teachable_moment": "The invariant to internalize: **any data crossing the server→HTML boundary needs to know which parser will see it first**. HTML parser sees `<script>` context, JSON parser sees the script body — but the HTML parser wins at the boundary. Unicode-escaping `<`, `>`, and `&` in JSON is a no-op for JavaScript (they parse identically) but completely defuses the HTML parser. This is a one-line fix that belongs in every server-rendered JSON injection utility.",
      "connection": "This connects to the broader pattern of context mismatch at serialization boundaries — the same class of bug appears in SQL (string vs. integer context), shell commands (quoted vs. unquoted), and template engines. The pattern: always apply the escaping appropriate to the *outer* context, not the inner one."
    },
    {
      "id": "F-002",
      "title": "`.run/` state directories accumulate indefinitely — no cleanup, production disk exhaustion",
      "severity": "MEDIUM",
      "category": "operational",
      "file": "packages/persona-engine/src/onboarding/verify-session.ts",
      "description": "Five distinct `.run/` trees (onboarding-tokens, onboarding-claims, onboarding-oauth-state, onboarding-oauth-state-claims, onboarding-siwe-nonce, onboarding-siwe-nonce-claims, onboarding-siwe-token-claims) accumulate files and mkdir-created directories indefinitely. Every verify button click writes at minimum two files and two claim directories. Expired state is never removed. In a high-traffic guild or under bot abuse (repeated verify clicks), this will fill the container filesystem. Railway ephemeral storage is particularly vulnerable.",
      "suggestion": "Add a periodic cleanup pass (e.g. cron or at mint time) that deletes token/nonce files and their claim directories older than `TTL_MS * 2`. A simple `find .run/onboarding-* -mmin +10 -delete` sweep run on a minute cron is sufficient for v1 scale.",
      "confidence": 1.0,
      "faang_parallel": "The Cassandra team at Netflix learned this lesson at scale: short-lived session tokens written to disk (or to an unbounded table partition) during high-traffic events created cascading disk pressure incidents that required manual intervention during peak load — exactly when you don't want to be doing ops work. Their fix: 'Every ephemeral write has a mandatory TTL.' The same principle drove Redis's key expiration design — TTL is not optional metadata, it is a correctness constraint on ephemeral state.",
      "metaphor": "This is the sticky-note version of session management. Every verify click leaves a sticky note on the wall. Nobody ever peels them off. A quiet guild is fine for months — then you get a role-gate reveal with 500 concurrent users and the wall is covered floor to ceiling.",
      "teachable_moment": "The invariant: **the lifetime of the cleanup mechanism must be designed at the same time as the write path, not as a follow-up task**. When ephemeral state lives on disk (vs. Redis with built-in TTL), the garbage collection responsibility falls to you. For v1 filesystem-based state, `find … -mmin +N -delete` is genuinely sufficient and testable.",
      "connection": "This pairs with F-003 (no timeout on OAuth calls): both are 'what happens under load / abuse?' gaps. The `.run/` accumulation is the disk dimension; the timeout gap is the concurrency dimension. A bot abuse scenario (bot user hammering the verify endpoint) exercises both simultaneously."
    },
    {
      "id": "F-003",
      "title": "Discord OAuth calls in `handleOAuthCallback` have no timeout",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "apps/bot/src/verify/verify-routes.ts",
      "description": "`exchangeCode` and `fetchDiscordUser` are called sequentially inside `handleOAuthCallback` with no `AbortController` timeout. If Discord's token or `/users/@me` endpoint is slow or stalled, the handler hangs indefinitely. Under a concurrent verify spike (e.g. a large role-gate reveal) this exhausts Bun's request concurrency pool. The identity-api and inventory-api clients both apply `fetchTimeoutMs` via AbortController; the OAuth path is the only enrichment path without one.",
      "suggestion": "Thread `fetchFn` through an AbortController wrapper with a configurable timeout (3000ms default matches the existing `DEFAULT_FETCH_TIMEOUT_MS`). Both `exchangeCode` and `fetchDiscordUser` already accept a `fetchFn` parameter — wrapping `rt.fetchFn` with a timed abort before passing it in is sufficient.",
      "confidence": 0.9,
      "faang_parallel": "The AWS SDK team documented this class of failure after the 2012 US-EAST-1 ELB incident: timeoutless outbound calls during a third-party degradation caused cascading thread-pool exhaustion across services that were nominally 'healthy.' Their post-incident requirement: every outbound call must have an explicit timeout, even when the downstream is considered 'reliable.' Discord's API has known latency spikes during deploys and DDoS events — the OAuth callback is precisely the path that fires at peak load.",
      "metaphor": "The identity-api and inventory-api clients wear seatbelts. The OAuth callback is the one passenger who said 'I'll be fine.' It's not about the likely case — it's about what happens to the whole vehicle when that one path stalls.",
      "teachable_moment": "When you see `fetchFn` injected for testability, the abstraction also enables timeout injection. The pattern: the test injects a mock fetch; production injects a timeout-wrapped fetch. Both go through the same `fetchFn` parameter. That's the correct composability shape — the test harness proved it, now production should benefit from it."
    },
    {
      "id": "F-004",
      "title": "`ONBOARDING_STATE_SECRET` length is not validated at boot — runtime throw on short secret",
      "severity": "LOW",
      "category": "operational",
      "file": "apps/bot/src/onboarding-runtime.ts",
      "description": "`buildOnboardingWiringFromEnv` validates that `stateSecret` is truthy but does not check `stateSecret.length >= 32`. The `secret()` functions in `state-token.ts` and `verify-session.ts` both throw `Error('[verify-session] ONBOARDING_STATE_SECRET ... <32 chars')` at call time. Setting `ONBOARDING_STATE_SECRET=short` passes the boot gate and logs `onboarding: ENABLED`, but the first verify button click throws an unhandled error into the Bun server `error()` handler returning a 500. The misconfiguration is invisible until a user triggers it.",
      "suggestion": "Add `stateSecret.length < 32` to the guard condition in `buildOnboardingWiringFromEnv` so a short secret disables onboarding at boot (returns `null`) and logs a clear diagnostic instead of failing silently at runtime.",
      "confidence": 0.95,
      "faang_parallel": "Google's SRE book dedicates a section to 'fail-fast configuration validation' — the principle that a service which accepts invalid configuration and silently degrades is more dangerous than one that refuses to start. The specific recommendation: all security-sensitive config (key lengths, certificate paths, token formats) should be validated before the service announces readiness. This is exactly why Kubernetes liveness vs. readiness probes are distinct: readiness should fail until config is valid.",
      "metaphor": "The boot gate currently checks that the key *exists* (the vault is not empty) but not that the key *fits the lock* (the vault's minimum strength requirement). You learn the key doesn't work when the first customer tries to open the door.",
      "teachable_moment": "The invariant: **the component that enforces a constraint should also validate it at the earliest possible moment**. If `state-token.ts` enforces `>= 32`, then `onboarding-runtime.ts` (which assembles the wiring) is the right place to validate it — not inside the hot path where it throws into a request handler."
    },
    {
      "id": "F-005",
      "title": "`worldSlug: 'mibera'` hardcoded in `link()` payload",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/onboarding/freeside-auth-client.ts",
      "description": "The `link()` method hardcodes `worldSlug: 'mibera'` in the POST body to `/v1/link/verified-wallet`. If the onboarding character is reused for a second world (e.g. `honey-port` or a future operator-configurable world), every link call would silently write to the wrong world spine. The `FreesideAuthConfig` has no `worldSlug` field, and `OnboardingWiring` has no path to parameterize it.",
      "suggestion": "Add `worldSlug: string` to `FreesideAuthConfig` (or to the `link()` input object). The default can remain `'mibera'` for the current deployment, but the field makes the coupling explicit and prevents silent mis-attribution.",
      "confidence": 0.9,
      "faang_parallel": "Stripe's internal postmortems have documented several incidents caused by hardcoded environment or account identifiers in shared library code — where a library intended for one product line quietly wrote to another product's datastore because the identifier was never surfaced as a parameter. Their engineering standard: any identifier that varies by deployment context must be a config parameter, even if today there is only one deployment.",
      "metaphor": "This is the address on a letter that's been pre-printed on the stationery. For one recipient it's fine. The moment you run the same stationery for a second world, every letter goes to Mibera regardless of who you meant to write to.",
      "teachable_moment": "Hardcoded identifiers are a specific category of accidental coupling. They're not bugs today — they're debts that compound silently until reuse forces them to the surface. The YAGNI counter-argument applies to features; it does not apply to identifiers that are already present in the surrounding config shape."
    },
    {
      "id": "F-006",
      "title": "`branchReply('restored')` reports role restoration regardless of `regrantRole` outcome",
      "severity": "LOW",
      "category": "correctness",
      "file": "apps/bot/src/discord-interactions/onboarding-dispatch.ts",
      "description": "`runOnboardingPrecheck` calls `runtime.regrantRole?.(interaction).catch(() => {})` and then unconditionally sends `\"found your link on file — restoring your verified role now.\"` via `editOriginal`. If `regrantRole` fails (Discord API hiccup, role hierarchy issue), the user reads a confirmation of role restoration that never happened. The link exists in identity-api, so the user is not in a broken state, but the message is factually incorrect and will generate support confusion.",
      "suggestion": "Await the `regrantRole` result and branch the copy: on success emit the current text; on failure emit something like `\"found your link — but the role didn't apply this time. try the button again.\"` This preserves the FR-13 retry UX while not lying to the user.",
      "confidence": 0.85,
      "faang_parallel": "The classic distributed systems lesson: 'fire and forget' is not a UX pattern, it is a backend optimization. Slack's incident retrospective on their 2019 message delivery reliability work explicitly identified 'optimistic delivery confirmation' — telling the user a message sent before confirming it did — as the primary driver of user trust erosion during partial outages. The lesson: the copy shown to the user should reflect the actual observable outcome, not the intended outcome.",
      "metaphor": "The locksmith tells you the lock is changed before confirming the key works. The lock might be fine — but if the new key doesn't turn, you've just locked yourself out and received a receipt saying you're in.",
      "teachable_moment": "`.catch(() => {})` is the 'ignore this failure' pattern. It is occasionally appropriate for truly fire-and-forget side effects (telemetry, non-critical logging). It is never appropriate when the UI confirmation depends on the outcome. The decision trail question: document why the failure is being silenced, or it looks like an oversight."
    },
    {
      "id": "F-007",
      "title": "PRAISE: Atomic single-use claim via `mkdir` is a clean O_EXCL-equivalent for in-process concurrency",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/persona-engine/src/onboarding/verify-session.ts",
      "description": "The `claim(claimsDir, id)` helper using `mkdirSync` correctly exploits filesystem atomicity to enforce single-use semantics for both OAuth states (SKP-004) and SIWE nonces (ATK-002). The per-token nonce claim (`NONCE_PER_TOKEN_CLAIMS`) correctly prevents two concurrent OAuth callbacks from the same verify URL from racing to both issue a nonce. The pattern is simple, auditable, and avoids a Redis/database dependency for a single-instance deploy.",
      "confidence": 1.0,
      "faang_parallel": "This is the classic `open(O_CREAT | O_EXCL)` pattern — the POSIX primitive for atomic 'create or fail' that underlies lock files, PID files, and one-time-token systems across Unix. The Linux kernel's own temporary file creation uses this guarantee. Using `mkdirSync` rather than `writeFileSync` is a subtlety worth noting: on most filesystems, directory creation is more reliably atomic across NFS/CIFS than file creation, making this approach robust even in environments where the `.run/` tree might eventually live on a network volume.",
      "metaphor": "This is the 'first one through the door locks it behind them' pattern. The filesystem is the judge — it will only allow one winner, and the winner holds proof of winning (the directory exists). No database, no Redis, no distributed lock. For a single-instance process, this is not a simplification — it is the correct primitive."
    },
    {
      "id": "F-008",
      "title": "PRAISE: SIWE verification enforces all five T3.0 criteria with correct static pin discipline",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/persona-engine/src/onboarding/siwe.ts",
      "description": "`verifySiweSignature` correctly reconstructs the canonical EIP-4361 message from server-controlled static config (domain, chainId, uri) plus server-issued session fields (nonce, issuedAt, expirationTime) plus the client-claimed address. The client cannot substitute a forked domain or a different chainId — any such tampering produces a different message, and `recover(reconstructed, signature) !== claimedAddress`. The timing check (`now >= exp`) uses the parsed date, not a client-supplied value. `timingSafeEqual` is used for the MAC comparison in the token layer, though it is not needed for `recovered !== claimed` (string equality on derived values is not a timing oracle here).",
      "confidence": 1.0,
      "faang_parallel": "The SIWE spec was designed with exactly this attack surface in mind after studying OAuth PKCE bypass vulnerabilities — where client-supplied parameters that should be server-verified were trusted verbatim. The 'reconstruct and compare' pattern (server reconstructs the canonical message, client supplies only the signature) is the same discipline that makes JWT signature verification correct: you never trust the header claims to tell you which algorithm was used. The paranoia here is earned.",
      "metaphor": "The server writes the letter, hands the envelope to the user to sign, and then re-reads the original letter it wrote before checking the signature. The user cannot change what was written — they can only prove they touched it."
    },
    {
      "id": "F-009",
      "title": "PRAISE: RT-3 service-token redaction is structurally enforced through tagged errors",
      "severity": "PRAISE",
      "category": "security",
      "file": "packages/persona-engine/src/onboarding/freeside-auth-client.ts",
      "description": "Tagged errors (`AuthHttpError`, `AuthNetworkError`, `AuthDecodeError`, `AuthInvalidInput`) carry only `{ endpoint, status }` or `{ endpoint, reason }` — never response headers, response bodies, or the service token. The token test `expect(JSON.stringify(exit)).not.toContain(TOKEN)` verifies this structurally on the serialized failure, not just the happy path. This correctly closes the RT-3 concern for all error paths including retried requests.",
      "confidence": 1.0,
      "faang_parallel": "Stripe's internal API client design principle — 'error objects are API surface, not debugger windows' — encodes the same insight. When a service token leaks through an error path into a log aggregator, the blast radius is the entire service identity, not just the failed request. The structural verification (serializing the error and asserting the token is absent) is the correct test strategy: it tests the artifact, not the code path, which means future refactors that accidentally capture additional context will fail the test.",
      "metaphor": "The tagged error types are like a tamper-evident envelope: if the token ever ends up inside, the test tears the envelope open and shows you. You don't need to read every line of error construction — the test watches the sealed output."
    }
  ]
}
```


---

## Callouts

This PR demonstrates genuine security architecture thinking that goes beyond the typical 'add a verify route' pattern. Three things stand out:

**The `mkdir`-as-O_EXCL claim pattern** (F-007) is the kind of solution that a systems programmer reaches for, not a web developer. It's correct, auditable, and avoids introducing a new infrastructure dependency for a single-instance deploy. That's exactly the right tradeoff for v1.

**The SIWE reconstruction discipline** (F-008) shows someone who read the spec for the attack model, not just the happy path. Binding domain, chainId, and URI to server-controlled config before signature verification is what separates a functional SIWE integration from a phishable one.

**Tagged error redaction** (F-009) with structural test coverage is a practice that many experienced teams skip until a token shows up in Datadog. The `JSON.stringify(exit)` assertion is simple and future-proof in a way that path-based tests are not.

The remaining fixes are all small — none require architectural change. The XSS escape (F-001) and the filesystem cleanup (F-002) are the two to land before this exits DRAFT status. The rest can follow.

---

PR #148, opened 2026-05-31
C-4: multi-tenant guild→world cutover + fail-soft config read

zkSoju:
## C-4 — single-tenant → multi-tenant cutover (THJ + purupuru)

Resolves the pressure test: the bot no longer runs on a single `DISCORD_GUILD_ID` + THJ-hardcoded channels. Adds guild→world resolution + per-event world config read from the freeside-worlds config service (fail-soft, mirroring the nym fetch).

**Security (BEAUVOIR fix-then-merge → fixed):**
- HIGH — tenant-isolation: `fetchVerifyMessageConfig` now verifies `envelope.world_slug === opts.worldSlug` before trusting config; a mismatched/unstamped response fails soft to code defaults (refuses cross-tenant config on the verify surface). 3 tests (cross-tenant reject, match accept, missing-stamp reject).

NOTE: config service not deployed yet (C-6) — built against the contract, mocked in tests. guild→world uses an interim seeded map (manifest `guild_ids` unpopulated). 523/523 bot tests, typecheck clean.

### DIFF
--- a/apps/bot/src/onboarding-runtime.ts
+++ b/apps/bot/src/onboarding-runtime.ts
@@ -22,6 +22,19 @@ const DEFAULT_STATEMENT = 'link your wallet to your discord to unlock the verifi
 export interface OnboardingWiring {
   onboarding: OnboardingRuntime;
   verify: VerifyRuntime;
+  /**
+   * C-4 multi-tenant verify-message config read. Populated from env when the
+   * freeside-worlds config service URL is set; undefined keeps the verify card
+   * on its single-tenant code defaults (the service is NOT deployed yet ·
+   * freeside-worlds arrakis-e5jk/C-6). A post site passes these straight into
+   * `buildVerifyCardForGuild(...)`.
+   */
+  worldConfig: {
+    /** WORLDS_CONFIG_API_URL — config-service base URL. Empty/unset → defaults only. */
+    configBaseUrl?: string;
+    /** WORLDS_CONFIG_SERVICE_TOKEN — read-gate `x-service-token`. */
+    serviceToken?: string;
+  };
 }
 
 /** Build both runtimes from env, or null if onboarding is not fully configured (→ stays off). */
@@ -79,5 +92,11 @@ export function buildOnboardingWiringFromEnv(env: NodeJS.ProcessEnv = process.en
     noRuntime: false,
   };
 
-  return { onboarding, verify };
+  // C-4 · per-world verify-message config seam (fail-soft: unset → code defaults).
+  const worldConfig = {
+    configBaseUrl: env.WORLDS_CONFIG_API_URL?.trim() || undefined,
+    serviceToken: env.WORLDS_CONFIG_SERVICE_TOKEN?.trim() || undefined,
+  };
+
+  return { onboarding, verify, worldConfig };
 }
--- a/apps/bot/src/tests/world-config.test.ts
+++ b/apps/bot/src/tests/world-config.test.ts
@@ -1,0 +1,386 @@
+/**
+ * world-config.test.ts — C-4 · single-tenant → multi-tenant cutover seam.
+ *
+ * The config service is NOT deployed (freeside-worlds arrakis-e5jk/C-6), so the
+ * service is MOCKED here via an injected `fetchFn`. Coverage:
+ *   - resolveWorld: seed (THJ + purupuru), env override (WORLD_GUILD_MAP),
+ *     invalid env (ignored, no throw), DM/unknown/malformed → null.
+ *   - fetchVerifyMessageConfig: happy 200, 404 fail-soft, 401 fail-soft,
+ *     timeout/abort fail-soft, malformed body fail-soft, unset baseUrl no-op,
+ *     service-token header forwarding.
+ *   - resolveVerifyCardCopy: per-medium escaping (RENDER-CONTRACT), disabled
+ *     surface → defaults, unknown guild → defaults.
+ *   - buildVerifyCardForGuild: end-to-end (known+configured → per-world copy;
+ *     unknown/outage → today's single-tenant card).
+ */
+
+import { describe, test, expect } from 'bun:test';
+import {
+  resolveWorld,
+  effectiveGuildWorldMap,
+  fetchVerifyMessageConfig,
+  resolveVerifyCardCopy,
+  buildVerifyCardForGuild,
+  type VerifyMessageConfig,
+} from '../world-config.ts';
+import { ONBOARD_VERIFY_CUSTOM_ID } from '@freeside-characters/persona-engine/onboarding';
+
+const THJ_GUILD = '1135545260538339420';
+const PURUPURU_GUILD = '1495534680617910396';
+const UNKNOWN_GUILD = '999999999999999999';
+
+// A config-service 200 body (config-service app.ts shape).
+function okBody(config: VerifyMessageConfig) {
+  return {
+    envelope: { schema_version: '1.0', world_slug: 'mibera', surface: 'verify-message', config },
+    version: 3,
+    updated_at: '2026-05-31T00:00:00Z',
+  };
+}
+
+function mockFetch(impl: (url: string, init?: RequestInit) => Response | Promise<Response>): typeof fetch {
+  return ((url: string | URL | Request, init?: RequestInit) =>
+    Promise.resolve(impl(String(url), init))) as unknown as typeof fetch;
+}
+
+const silentLogger = { warn: () => {} };
+
+// ─── resolveWorld ──────────────────────────────────────────────────────────
+
+describe('resolveWorld — guild → world_slug', () => {
+  test('THJ guild seeds to mibera (verified slug-decision)', () => {
+    expect(resolveWorld(THJ_GUILD, {})).toBe('mibera');
+  });
+
+  test('purupuru guild seeds to purupuru', () => {
+    expect(resolveWorld(PURUPURU_GUILD, {})).toBe('purupuru');
+  });
+
+  test('unknown guild → null (fail-soft to single-tenant defaults)', () => {
+    expect(resolveWorld(UNKNOWN_GUILD, {})).toBeNull();
+  });
+
+  test('DM / undefined / empty guild → null', () => {
+    expect(resolveWorld(undefined, {})).toBeNull();
+    expect(resolveWorld(null, {})).toBeNull();
+    expect(resolveWorld('', {})).toBeNull();
+  });
+
+  test('malformed (non-snowflake) guild → null', () => {
+    expect(resolveWorld('not-a-snowflake', {})).toBeNull();
+    expect(resolveWorld('123', {})).toBeNull(); // too short
+  });
+
+  test('WORLD_GUILD_MAP env override wins over seed', () => {
+    const env = { WORLD_GUILD_MAP: JSON.stringify({ [THJ_GUILD]: 'mongolian' }) };
+    expect(resolveWorld(THJ_GUILD, env)).toBe('mongolian');
+  });
+
+  test('WORLD_GUILD_MAP env adds a new guild', () => {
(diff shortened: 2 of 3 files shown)


zkSoju (commented):
## Summary

C-4 is a confident piece of infrastructure work — the fail-soft architecture is disciplined, the tenant isolation check is exactly right, and the test file earns its 449 lines. The one real action item is WC-001: `resolveVerifyCardCopy` makes a "never throws" promise it cannot keep when config-service returns structurally valid but type-wrong values. Everything else is polish.

**Verdict**: `REQUEST_CHANGES` — WC-001 is a contract violation on the hot path; the remaining findings are low-friction and can land in-iteration.

---

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "WC-001",
      "title": "resolveVerifyCardCopy violates its own 'never throws' contract — copy field types unvalidated before escapeForDiscord",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "apps/bot/src/world-config.ts",
      "description": "fetchVerifyMessageConfig validates that config.copy is truthy and config.enabled is boolean, but does not check that config.copy.title/body/buttonLabel are actually strings. A malformed config-service response (e.g. { enabled: true, copy: { title: null, body: 1, buttonLabel: [] } }) passes the guard and is returned. resolveVerifyCardCopy then calls escapeForDiscord(config.copy.title) without a try-catch. escapeForDiscord calls s.replace(...) — if s is null or not a string at runtime, this throws a TypeError. resolveVerifyCardCopy has no try-catch; neither does buildVerifyCardForGuild. Both functions carry docstrings claiming 'Never throws', and the RENDER-CONTRACT contract relies on this. The config service is internal, but the fail-soft invariant should be unconditional.",
      "suggestion": "In fetchVerifyMessageConfig, extend the copy validation to check string types: `if (!config || typeof config.enabled !== 'boolean' || !config.copy || typeof config.copy.title !== 'string' || typeof config.copy.body !== 'string' || typeof config.copy.buttonLabel !== 'string') { ... return null; }`. Alternatively, wrap the escapeForDiscord block in resolveVerifyCardCopy with a try-catch that returns {} on any error.",
      "confidence": 0.85,
      "faang_parallel": "Stripe's API response validation layer enforces field-level type checks even on internal service responses — their post-mortem on a 2019 incident showed that a malformed internal JSON body (number where string expected) propagated silently through a truthy guard and surfaced as a customer-visible 500. The fix: parse-then-validate at the boundary, never assume shape from truthiness.",
      "metaphor": "Checking `config.copy` is truthy is like confirming the delivery box arrived but not opening it. The box could contain a brick where the vase should be. `escapeForDiscord` is the vase handler — it will throw the brick across the room.",
      "teachable_moment": "A 'never throws' docstring is a contract, not a comment. Contracts require enforcement at every input boundary, including internal ones. Internal services change; response shapes drift. The contract should be unconditional — not 'never throws unless the internal service misbehaves.'",
      "connection": "This is the same class of defect as WC-005 — partial validation at the fetch boundary that lets type-unsafe values through to downstream callers. Both are fixed by completing the validation at the one place that sees the raw response."
    },
    {
      "id": "WC-002",
      "title": "buildGuildWorldMap logs to console.warn directly, bypassing the injected logger seam",
      "severity": "LOW",
      "category": "design",
      "file": "apps/bot/src/world-config.ts",
      "description": "buildGuildWorldMap calls console.warn directly for invalid JSON, non-object JSON, non-snowflake keys, and invalid slug values (lines ~116–144). resolveWorld does not accept a logger parameter. Callers that pass logger: silentLogger to resolveVerifyCardCopy or fetchVerifyMessageConfig cannot suppress buildGuildWorldMap warnings, so test runs that exercise invalid WORLD_GUILD_MAP paths produce console noise. This is an inconsistency in the module's logger injection model.",
      "suggestion": "Add an optional `logger` parameter to buildGuildWorldMap and thread it through from resolveWorld. Or accept a logger at resolveWorld's call site and forward it. Keeps all logging paths suppressible in tests and structured in production.",
      "confidence": 0.9,
      "faang_parallel": "Netflix's hollow logger pattern (used in their Hystrix library) made every internal utility accept an optional logger rather than hard-coding to slf4j. The payoff came in test suites — zero console bleed, assertions on log calls instead of output scanning.",
      "metaphor": "A fire alarm wired directly to the PA system instead of the fire panel. Every other alarm in the building respects the mute switch; this one broadcasts regardless.",
      "teachable_moment": "Logger injection isn't just about testability — it's about operational control. A module that bypasses the injected logger cannot be silenced in high-frequency paths, cannot route to structured sinks, and cannot be observed uniformly. Complete the seam once at module boundary; every callsite benefits.",
      "connection": "The existing logger: parameter on fetchVerifyMessageConfig and resolveVerifyCardCopy shows the seam was intentionally designed. buildGuildWorldMap is the one path that didn't get threaded through. Low friction to complete."
    },
    {
      "id": "WC-003",
      "title": "WORLD_SLUG regex permits trailing hyphens — may diverge from upstream config-protocol pattern",
      "severity": "LOW",
      "category": "correctness",
      "file": "apps/bot/src/world-config.ts",
      "description": "WORLD_SLUG = /^[a-z][a-z0-9-]{1,20}$/ accepts slugs ending with a hyphen (e.g. 'mibera-', 'apdao-'). The comment states this mirrors config-protocol WORLD_SLUG_PATTERN. If the upstream pattern forbids trailing hyphens (common for slug conventions), a slug passing bot validation could fail config-service URL routing, producing a 404 instead of a proper config response. The fail-soft design absorbs this silently.",
      "suggestion": "Change the regex to disallow trailing hyphens: /^[a-z][a-z0-9]*(?:-[a-z0-9]+)*$/ (hyphen only between alphanumeric segments). Verify this matches the config-protocol canonical pattern.",
      "confidence": 0.6,
      "faang_parallel": "GitHub's slug validation for repository names uses the same segment pattern — hyphens only between alphanumeric segments, never leading or trailing. The validation is duplicated at every entry point (API, web, git) because silent divergence at one layer produces 404s at another.",
      "metaphor": "A bouncer who checks the dress code but not whether the shirt is on inside-out. Technically meets the criteria; the venue will still turn them away at the next checkpoint.",
      "teachable_moment": "When a validation regex claims to mirror an upstream pattern, the claim should be verifiable — ideally by importing the upstream constant. A comment saying 'mirrors X' without a test asserting parity is an assertion that decays.",
      "connection": "Moderate confidence — verify against the actual config-protocol source before acting. If the upstream allows trailing hyphens too, this finding is moot."
    },
    {
      "id": "WC-004",
      "title": "responseWorld logged without length cap — log injection surface from malicious config-service response",
      "severity": "LOW",
      "category": "security",
      "file": "apps/bot/src/world-config.ts",
      "description": "In the tenant-mismatch warn path, responseWorld (body.envelope?.world_slug from the HTTP response) is interpolated directly into the log message without truncation or sanitization. A compromised or misconfigured config service could return a world_slug containing newlines, ANSI escape sequences, or extremely long values, corrupting structured log streams. The error message for fetch failures is correctly truncated to 160 chars; this path is not.",
      "suggestion": "Apply the same truncation pattern used for error messages: `(responseWorld ?? '<none>').slice(0, 64)`. Strip or encode control characters before interpolation.",
      "confidence": 0.5,
      "faang_parallel": "Cloudflare's Workers logging pipeline was burned by log injection in 2021 when an external header value containing newlines was interpolated into a structured log line, splitting one log event into two with the second appearing to be a new event from a trusted source.",
      "metaphor": "The error-message truncation is the safety cap on the bottle. This path is the same bottle without the cap — probably fine, but the risk profile is identical.",
      "teachable_moment": "Anything from an external response is untrusted. Apply the same sanitization discipline consistently — if you truncate error messages, truncate response fields. Inconsistency is where reviewers stop looking and attackers start.",
      "connection": "Low confidence — this requires a compromised internal config service. Surface-area is real but low-probability. Fix is one-liner; apply it."
    },
    {
      "id": "WC-005",
      "title": "serviceToken not trimmed before header insertion in fetchVerifyMessageConfig",
      "severity": "LOW",
      "category": "correctness",
      "file": "apps/bot/src/world-config.ts",
      "description": "The check `if (opts.serviceToken) headers['x-service-token'] = opts.serviceToken` passes a truthy serviceToken including values that are all-whitespace (e.g. '   '). buildOnboardingWiringFromEnv correctly applies `.trim() || undefined`, but direct callers of fetchVerifyMessageConfig (including tests, future callers) do not have this guarantee. A whitespace-only token would be sent as a header value.",
      "suggestion": "In fetchVerifyMessageConfig, trim and re-check: `const token = opts.serviceToken?.trim(); if (token) headers['x-service-token'] = token;`",
      "confidence": 0.7,
      "faang_parallel": "AWS SDK v3 normalizes all credential strings at the provider boundary rather than trusting that callers upstream already sanitized them. Defense-in-depth at the point of use, not the point of origin.",
      "metaphor": "The front door locks, but the interior door does not. You trust the front door. Future guests might enter through a window.",
      "teachable_moment": "When a normalization step exists in one caller (buildOnboardingWiringFromEnv), the temptation is to trust it. But functions that accept raw parameters should normalize at their own boundary — they cannot know all their callers.",
      "connection": "Same root as WC-001: partial validation at the fetch boundary. Completing the trim here closes the assumption that all callers pre-sanitize."
    },
    {
      "id": "WC-006",
      "title": "PRAISE: Tenant isolation check prevents cross-tenant config leak with full test coverage",
      "severity": "PRAISE",
      "category": "security",
      "file": "apps/bot/src/world-config.ts",
      "description": "The envelope.world_slug !== opts.worldSlug guard in fetchVerifyMessageConfig is a correct defense against a misconfigured or compromised config service returning world B's config in response to a world A request on the verify surface. The test suite covers the full matrix: cross-tenant rejection, matching-tenant acceptance, and missing world_slug rejection. The warn message explains the decision. This is the right place to enforce the invariant — the bot as verifier, not the config service.",
      "suggestion": null
    },
    {
      "id": "WC-007",
      "title": "PRAISE: Fail-soft architecture is exhaustive and consistently enforced",
      "severity": "PRAISE",
      "category": "design",
      "file": "apps/bot/src/world-config.ts",
      "description": "Every failure path (unknown guild, unset baseUrl, malformed worldSlug, 404/401/5xx, network error, AbortController timeout, malformed body, missing config field, enabled=false) returns null or {} and never throws. The AbortController + setTimeout + finally clearTimeout pattern mirrors the existing announce-mint.ts posture. The test file systematically exercises all 14+ failure paths, including the timeout path with a real AbortController signal listener. The contract that an unknown or outaged world always falls through to today's single-tenant defaults is honored at every call site.",
      "suggestion": null
    },
    {
      "id": "WC-008",
      "title": "PRAISE: WORLD_GUILD_MAP env override is fully validated before use",
      "severity": "PRAISE",
      "category": "design",
      "file": "apps/bot/src/world-config.ts",
      "description": "The env override path validates: JSON parsability, object vs array/null shape, per-entry snowflake key format, per-entry slug value format. Invalid entries are skipped individually with warnings rather than discarding the entire override. Seed entries survive invalid env. The merge-over (not replace) semantics ensure the bot is never left with an empty map from a misconfigured env. Tests cover bad JSON, array JSON, mixed-validity entry maps, and the seed-intact invariants.",
      "suggestion": null
    }
  ]
}
```


---

## Callouts

**WC-006 / WC-007 / WC-008** deserve to be named together because they represent a coherent design philosophy: the bot is the verifier of last resort. It doesn't trust the config service to return the right tenant's config. It doesn't trust the environment to contain valid JSON. It doesn't trust network availability. Every external dependency is treated as adversarial-at-worst, unreliable-at-best, and the fallback is always defined. This is exactly how Kubernetes controllers are designed — level-triggered, not edge-triggered; reconcile from known state, not from assumed state. The 449-line test file isn't padding; it's the specification made executable. That discipline is rare and worth preserving as the multi-tenant surface grows.

The AbortController + `finally clearTimeout` pattern in particular is a small thing done exactly right — a common resource leak in fetch wrappers that most production codebases discover only in load testing. Noticing it here means it was written with awareness, not luck.

---

PR #161, opened 2026-06-03
feat(shadow-onboarding S4): live Discord Layers + cross-repo gate proof + CV2

zkSoju:
## S4 — shadow-onboarding live Discord Layers + cross-repo gate proof

The Discord I/O actor for the shadow-onboarding substrate. Supplies the live/mock `RosterSource` + `RoleWriter` Layers, proves the substrate's "SHADOW ⇒ zero writes" gate survives the **repo boundary**, and renders the same `Discrepancy` as a Components-V2 message. Voiceless — no onboarding logic; the pinned substrate (`@freeside-worlds/shadow-substrate#26d11b78`) governs.

**Headline proofs**
- **Cross-repo import-boundary lint** — a raw `discord.js` role-mutation anywhere outside the single gated adapter **fails CI** (§8.4 proof 4, proven against a planted violation). The live writer requires a `WriteCapability` and is reachable only via the substrate's `GateCheckedRoleWriter`.
- Check-then-create + per-world lock (B10) · FR-9 namespaced coexistence · non-destructive rollback (GC unassigned roles, **never strips assigned users** — R-6, with member-cache hydration) · the pre-go-live 250-role quota · CV2 render (medium-agnostic, G-5).

**Review state** — FAGAN-converged over 4 iterations. The panel earned its keep on the live-write path: it caught a name-based confused-deputy (adopt/assign by role name), an empty-cache strip-users bug, CV2 mention/markdown injection, a fail-**open** namespace guard (`"".startsWith("")`), and an env-controllable conformance-bypass. 489 bot tests pass; conformance + import-boundary lint + typecheck green.

**To go live (operator — data + deploy, not code)**
1. Fill `purupuru.yaml` (worlds): `guild_id`, `nft_contracts`, `member_set`, `admin_principals`. The substrate fails **closed** (deny-all) until populated.
2. Deploy a config-service at `CONFIG_SERVICE_URL` + wire live jwks / Discord bot token / NATS signing key, then run the live E2E.
3. One substrate follow-up: thread `role_id` into `AssignRoleIntent` so cross-batch assign binds by id, not name (the namespace guard bounds the risk meanwhile).

Part of the cross-repo `shadow-onboarding-substrate` cycle.


zkSoju (commented):
## Summary

This is genuinely strong substrate work: a shadow-onboarding layer that earns its trust through fail-closed namespace guards, member-cache hydration before destructive reads, and a defense-in-depth import lint that's honest about being accident-prevention rather than a security boundary. The primary concerns are a cluster of *read-amplification* and *unbounded-wait* patterns on the live Discord write path (F1, F2, F5, F6) and — more seriously — a non-deterministic substrate-linking step that lets two divergent `@0xhoneyjar/events` builds coexist in one process (F3→F4), which is corrosive specifically because ACVP envelopes live or die by canonicalization identity. Verdict: **REQUEST_CHANGES**, weighted on F3/F4 (version skew in a hash-chained audit path) and F7 (an implicit, untested full-membership dependency guarding role *deletion*).

The architectural through-line worth naming: this PR repeatedly does the right thing at the *correctness* layer and then leaves the *amplification* layer for later. Hydrating the member cache before a GC zero-check (F17) is exactly right — but doing it once per role (F1) turns a correctness fix into a 429 generator. The WorldLock already hands you a serialization boundary that *is* a consistency snapshot; most of the perf findings dissolve if the batch reads its roster once and reuses it. That's not a coincidence of three separate bugs — it's one missing idea (snapshot-per-batch) showing up in three places.

## Findings


```json
{
  "schema_version": 1,
  "verdict": "REQUEST_CHANGES",
  "findings": [
    {
      "id": "F1",
      "title": "GC re-fetches the ENTIRE guild member list once per role (O(N) full member fetches)",
      "severity": "MEDIUM",
      "category": "performance",
      "file": "apps/bot/src/shadow/role-writer.live.ts",
      "description": "makeGatedRoleGc calls `await guild.members.fetch()` (hydrate ALL members) on every invocation, and executeRollbackGc (composition-root.ts) calls the returned `gc(...)` once per role in `plan.gc`. A rollback that GCs M roles therefore performs M full-guild member fetches. On a large guild this is very slow and a strong 429 trigger, and it is redundant — the member cache is identical across the loop.",
      "suggestion": "Hoist `guild.members.fetch()` to a single call before iterating plan.gc (e.g. pass a pre-hydrated guild/roster into the GC executor), or memoize hydration per guild for the batch's duration.",
      "confidence": 0.6,
      "faang_parallel": "This is the N+1 query problem that birthed Facebook's DataLoader — the fix is identical: batch the invariant fetch once, fan out the per-item work. Discord's 429 is precisely the Borg-style backpressure Google's SRE book warns hides behind 'it worked in the small guild'.",
      "metaphor": "Re-counting every tenant in the building each time you want to evict one apartment — the census doesn't change between evictions.",
      "teachable_moment": "A loop body should depend only on what varies per iteration. Anything constant across the loop — here, the hydrated member cache — is a hoist candidate. The cost of getting this wrong scales with guild size, so it passes every test and fails every large customer.",
      "connection": "Same root cause as F2: both are read-amplification that the per-batch WorldLock already gives you the boundary to eliminate."
    },
    {
      "id": "F2",
      "title": "Live writer re-fetches the full roleset and guild on every create/assign op",
      "severity": "MEDIUM",
      "category": "performance",
      "file": "apps/bot/src/shadow/role-writer.live.ts",
      "description": "createRole and assignRole each call guildFor() and `guild.roles.fetch()` (entire roleset). assignRole without a same-batch boundId also does a full `guild.roles.fetch()` to resolve by name, plus `guild.members.fetch(member_id)`. For a batch that creates a handful of roles and assigns many members, this is O(ops) full-roleset reads — wasteful and rate-limit-prone. The WorldLock serializes the batch, so a single roster snapshot could be reused across ops within the batch.",
      "suggestion": "Fetch the roleset once per batch (or cache it in the Layer instance, which already lives for one batch) and resolve create/assign against that snapshot, refreshing only on a detected miss.",
      "confidence": 0.5,
      "faang_parallel": "Spanner's read-snapshot semantics: a transaction reads a single consistent snapshot, not the live store on every statement. Your WorldLock is the transaction; treat the first roster fetch as the snapshot.",
      "metaphor": "Re-reading the entire dictionary to look up each word in a sentence — open it once, keep your finger on the page.",
      "teachable_moment": "When you already hold a serialization lock, you've paid for consistency — spend it. A serialized batch is the cheapest place in the system to cache, because invalidation is bounded by the lock's lifetime.",
      "connection": "F1's sibling. The Layer-per-batch lifetime is the natural cache scope for both."
    },
    {
      "id": "F3",
      "title": "fixup-substrate-bun.sh does not actually 'prefer persona-engine' — it links the first events copy found",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "scripts/fixup-substrate-bun.sh",
      "description": "The header comment claims it shares 'the persona-engine workspace copy' of @0xhoneyjar/events, but the loop simply `break`s on the FIRST `find` hit whose package name is @0xhoneyjar/events. With this PR there are now TWO events versions in the tree (#68f5a89 == CANONICAL_EVENTS_PIN, and persona-engine's transitive #56585fd). Which copy the substrate gets symlinked to is filesystem-find-order-dependent and non-deterministic.",
      "suggestion": "Make selection deterministic: explicitly resolve the events copy whose package.json/version matches CANONICAL_EVENTS_PIN (#68f5a89), or fail loudly if no matching-pin copy exists, rather than `break`ing on the first arbitrary find result.",
      "confidence": 0.55,
      "faang_parallel": "Google's monorepo single-version policy exists for exactly this: when correctness depends on *which* copy of a library you bind, `find | head -1` is a coin flip across machines and filesystem orderings. Hyrum's Law guarantees someone is already depending on whichever copy happens to win.",
      "metaphor": "Grabbing whichever key is nearest in the drawer instead of the one labeled for this lock — it'll fit some door, just maybe not yours.",
      "teachable_moment": "Never `break` on first-match when correctness hinges on identity. Match on the invariant (the pinned SHA), and if nothing matches the invariant, that's a loud failure, not a silent first-pick.",
      "connection": "This is the *mechanism* by which F4 (version skew) actually bites — nondeterministic linking is what lets the wrong build win in production but not in CI."
    },
    {
      "id": "F4",
      "title": "Two divergent @0xhoneyjar/events versions coexist in the bot process",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "bun.lock",
      "description": "apps/bot pins @0xhoneyjar/events at #68f5a89, while persona-engine resolves #56585fd. The bot's acvp-emitter.live.ts imports publishEnvelope/Signer/PrevHashStore from #68f5a89; persona-engine's NATS wiring uses #56585fd. If JCS canonicalization, hash-chain, or signing envelope shape differs between these builds, ACVP envelopes emitted from the same process via different code paths could be mutually inconsistent (chain/signature verification skew).",
      "suggestion": "Unify the events pin across the workspace (bump persona-engine to #68f5a89), or document and verify that #56585fd and #68f5a89 are envelope/hash compatible.",
      "confidence": 0.45,
      "faang_parallel": "The classic 'two copies of the serializer in one process' bug — and ACVP raises the stakes the way protobuf wire-format skew does at Google: a one-byte canonicalization difference doesn't corrupt loudly, it silently fails downstream verification on a different machine weeks later.",
      "metaphor": "Two clocks in the same room a half-second apart — neither is obviously wrong, but every timestamp they jointly produce is now suspect.",
      "teachable_moment": "Hash-chain and signature schemes are exquisitely version-sensitive because their whole value is byte-determinism (this is why CLAUDE.loa.md mandates lib/jcs.sh over `jq -S`). Two events builds in one process is a latent canonicalization fork until proven otherwise.",
      "connection": "Verify before teaching here (confidence 0.45): the severity depends entirely on whether the two SHAs share JCS+envelope code. Confirm the diff between #68f5a89 and #56585fd touches canonicalization before treating this as load-bearing."
    },
    {
      "id": "F5",
      "title": "ConfigServiceClient fetches have no timeout/abort",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "apps/bot/src/shadow/config-service-client.ts",
      "description": "getOnboardingLifecycle, putOnboardingLifecycle, and health issue `this.doFetch(...)` with no AbortController/timeout. A hung or slow config-service blocks the onboarding lifecycle indefinitely with no upper bound, which is exactly the path that drives go_live progress the lens polls.",
      "suggestion": "Add an AbortController with a bounded timeout (and surface a typed timeout error), consistent with the bounded backoff used elsewhere in the shadow layer.",
      "confidence": 0.6,
      "faang_parallel": "Google's SRE axiom 'every RPC needs a deadline' and Netflix's Hystrix both exist because of this exact outage class — one slow downstream silently consumes the caller's liveness. A missing timeout isn't a missing feature, it's a latent hang waiting for the dependency to have a bad day.",
      "metaphor": "A phone call with no hang-up button: if the other side never speaks, you wait until the heat death of the universe.",
      "teachable_moment": "Unbounded waits compose into cascading stalls — the polling lens makes this worse, because a hung put becomes a stuck go_live progress bar with no error to act on. Bound it, type the timeout, surface it.",
      "connection": "The rest of the shadow layer already uses bounded backoff — this is an inconsistency, not a missing concept. Apply the existing pattern."
    },
    {
      "id": "F6",
      "title": "CV2 discrepancy render has no size bounding — can exceed Discord component/message limits",
      "severity": "MEDIUM",
      "category": "reliability",
      "file": "apps/bot/src/shadow/discrepancy-cv2.ts",
      "description": "renderDiscrepancyCV2 concatenates every before/after/latent/pre-existing role into text components with no length cap. The pre-existing line joins ALL untouched role names (a guild can approach the 250-role ceiling), and Discord enforces per-text-component (~4000 char) and per-message size limits. A guild with many roles would produce a payload Discord rejects, failing the preview send. There is also no cap on the number of components.",
      "suggestion": "Truncate long role lists with an explicit '+N more' affordance and bound total rendered length to Discord's CV2 limits before constructing the payload.",
      "confidence": 0.55,
      "faang_parallel": "Twitter's render pipeline truncates and links rather than emitting an oversized payload; gRPC enforces a max message size for the same reason. The sink has a hard limit, so the producer must render against that limit — not against the data's natural size.",
      "metaphor": "Pouring a gallon into a pint glass: the glass doesn't stretch, it overflows — and here 'overflow' means the whole preview silently fails to send.",
      "teachable_moment": "Render against the sink's constraints, not the source's cardinality. The failure mode is worst in exactly the guilds that most need the preview (large, many-role) — so the bug is invisible until it hits a real customer.",
      "connection": "Same shape as F5/F1 — a code path that's correct on small inputs and fails on the large inputs that matter."
    },
    {
      "id": "F7",
      "title": "GC zero-member guard depends on members.fetch() returning the COMPLETE membership",
      "severity": "MEDIUM",
      "category": "security",
      "file": "apps/bot/src/shadow/role-writer.live.ts",
      "description": "The R-6 'never strip users' guard reads `role.members.size` after `guild.members.fetch()`. role.members is derived by filtering the guild member cache, so correctness hinges on members.fetch() hydrating the FULL membership. If the GUILD_MEMBERS privileged intent is missing or the fetch returns a partial set, the count could under-report and a still-assigned role could be deleted. (discord.js typically rejects members.fetch() without the intent, which would fail-safe via classifyWriteError, but this hard dependency is implicit and untested.)",
      "suggestion": "Assert/log that the full membership was hydrated (e.g. compare fetched count to guild.memberCount) before trusting size===0, and document the GUILD_MEMBERS intent requirement as a hard prerequisite for GC.",
      "confidence": 0.45,
      "faang_parallel": "The fail-open vs fail-closed distinction that defines safety-critical design — the same class as monitoring gaps misread as 'healthy' (the absence of an alert interpreted as the absence of a problem). AWS's IAM evaluates 'no data' as deny, never as allow, for exactly this reason.",
      "metaphor": "Demolishing a room because the door scanner was offline and reported zero guests — 'I saw nobody' and 'I couldn't see' must not be the same branch when the action is irreversible deletion.",
      "teachable_moment": "On a destructive path, 'known empty' and 'unknown' are different states and partial data must fail closed. The fix isn't to trust the count more — it's to prove the count is total (compare to guild.memberCount) before acting on size===0.",
      "connection": "The praiseworthy hydration in F17 is what makes this *usually* safe; this finding is the missing assertion that closes the implicit dependency the fix relies on."
    },
    {
      "id": "F8",
      "title": "escapeRoleName omits '-' despite the comment claiming line-leading '-' is neutralized",
      "severity": "LOW",
      "category": "security",
      "file": "apps/bot/src/shadow/discrepancy-cv2.ts",
      "description": "The JSDoc states it neutralizes '#'/'>'/'-' line-leading structure, but the escape regex is /[<>`*_~|#]/g — '-' is not in the set. Newlines are collapsed to spaces first, which mitigates list-marker injection in most positions, but the comment and implementation disagree and a leading '-' on the whole string is not escaped.",
      "suggestion": "Either add '-' to the escape set (and confirm '>' coverage) or correct the comment to match the actual escaped character set.",
      "confidence": 0.5,
      "faang_parallel": "Comment-code drift in sanitizers is a recurring source of XSS/markdown-injection CVEs — GitHub's own markdown pipeline has shipped fixes where the documented allowlist diverged from the regex. The comment is a security claim; an unverified security claim is a liability.",
      "metaphor": "A lock labeled 'deadbolt' that's actually a spring latch — the label tells the next person it's safe to lean on.",
      "teachable_moment": "Comments are claims; only tests are proof. When a sanitizer's docstring enumerates a charset, a unit test should assert each character is actually neutralized — otherwise the comment decays into confident misinformation.",
      "connection": "Newline-collapse-first is good defense-in-depth; this is about closing the gap between the stated and actual guarantee."
    },
    {
      "id": "F9",
      "title": "WorldLock concurrency semantics are untested",
      "severity": "LOW",
      "category": "testing",
      "file": "apps/bot/src/shadow/world-lock.ts",
      "description": "makeInMemoryWorldLock is the cross-batch B10 TOCTOU guard (same-world serialization, different-world parallelism) and includes a non-trivial registryGuard to avoid two first-touches creating distinct semaphores. No test exercises (a) two concurrent same-world withWorldLock calls serializing, (b) different-world calls running in parallel, or (c) the registry race.",
      "suggestion": "Add Effect-based concurrency tests asserting serialization for the same world and parallelism across worlds, including a concurrent first-touch on the same world.",
      "confidence": 0.5,
      "faang_parallel": "Jepsen exists because concurrency primitives that pass single-threaded tests routinely violate their invariants under contention. Go shipped its race detector precisely because 'looks correct' and 'is correct under interleaving' are different claims.",
      "metaphor": "A fire door you've never tested under a crowd — it latches fine when you push it gently, alone.",
      "teachable_moment": "Test load-bearing concurrency code inversely to how often it changes: the registryGuard first-touch race is the kind of bug that surfaces once in production at 2am and never reproduces locally. The exactly-one-create guarantee deserves an adversarial test, not just a happy-path one."
    },
    {
      "id": "F10",
      "title": "Live AcvpEmitter confirm-semantics rely on publishEnvelope actually awaiting an ack",
      "severity": "LOW",
      "category": "reliability",
      "file": "apps/bot/src/shadow/acvp-emitter.live.ts",
      "description": "emitConfirmed treats a resolved `publishEnvelope` as write-after-audit confirmation, but NatsLike.publish is typed `void | Promise<unknown>` — a fire-and-forget core.NATS publish resolves before broker persistence/ack. If publishEnvelope does not await a JetStream ack, 'confirmed' overstates durability and a LIVE write could proceed after an unacknowledged (possibly lost) audit event.",
      "suggestion": "Verify publishEnvelope awaits a JetStream publish ack for the shadow.* subjects; if not, use an ack-returning publish. Add a unit test for emitConfirmed success and AuditError-on-throw.",
      "confidence": 0.4,
      "faang_parallel": "Kafka's acks=0/1/all is the canonical taxonomy here: a resolved core-NATS publish is acks=0 (at-most-once), but emitConfirmed is asserting acks=all (durable) semantics. Conflating the two is how 'audited writes' quietly become 'probably-audited writes'.",
      "metaphor": "Dropping a letter in the mailbox and marking it 'delivered' — the slot accepting it isn't the recipient receiving it.",
      "teachable_moment": "Durability claims must be backed by an acknowledgment, not by a resolved promise. The word 'confirmed' in a write-after-audit gate is load-bearing; verify the broker actually acked before the gate opens.",
      "connection": "Worth verifying first (confidence 0.4): if publishEnvelope already awaits a JetStream ack internally, this is a documentation/test gap, not a durability bug."
    },
    {
      "id": "F11",
      "title": "Pervasive `as never` casts on RoleWriter return paths bypass the branded-id type",
      "severity": "LOW",
      "file": "apps/bot/src/shadow/role-writer.live.ts",
      "description": "createRole returns `priorInBatch as never`, `existing.id as never`, and `created.id as never`; the mock similarly returns `id as never`. `as never` defeats the branded RoleId contract — a wrong-typed value (or an accidental undefined) would not be caught by tsc. The substrate-conformance file uses a precise branded cast (`as WriteCapability['report_hash']`); the writer paths should match that rigor.",
      "suggestion": "Cast through the substrate's branded RoleId helper/type (e.g. `as RoleId`) rather than `as never`, so the compiler still checks the value is a string id.",
      "confidence": 0.45,
      "faang_parallel": "Branded/nominal types are exactly the discipline the TypeScript team and the Rust newtype pattern advocate to make illegal states unrepresentable — `as never` is the maximal escape hatch, throwing away the very guarantee branding exists to provide.",
      "metaphor": "Signing a blank check the accountant is contractually forbidden from reading — whatever you wrote, it clears.",
      "teachable_moment": "When you must cast, cast to the *narrowest* correct type, never the widest. `as RoleId` still forces tsc to confirm you're handing back a string; `as never` would happily pass `undefined`. The conformance file already shows the house style — match it."
    },
    {
      "id": "F12",
      "title": "import-boundary lint planted-violation files are not crash-safe and can leak into other scans",
      "severity": "LOW",
      "category": "testing",
      "file": "apps/bot/src/shadow/import-boundary.test.ts",
      "description": "The tests write real files (__planted_violation_TEST__.ts with `guild.roles.create`, __planted_delete_TEST__.ts with role.delete) under apps/bot/src/shadow/ and remove them in finally. If the test process is killed between write and rmSync, a real ungated role-mutation file is left in the tree and will fail the import-boundary lint on the next CI run. A parallel lint invocation during this test window could also observe the planted file.",
      "suggestion": "Write planted fixtures into an OS temp directory and point the lint at a configurable root for the test, or guard with a pretest cleanup that removes stale __planted_*_TEST__.ts before running.",
      "confidence": 0.4,
      "faang_parallel": "Bazel's sandboxing exists to make exactly this impossible: tests that mutate the live tree are non-hermetic and produce flaky, order-dependent CI. Google's test hermeticity rules forbid writing outside the test's declared outputs for this precise failure.",
      "metaphor": "Running a fire drill with real gasoline — fine until someone trips the breaker mid-drill.",
      "teachable_moment": "Tests must be hermetic: write fixtures to a temp dir and parameterize the tool's scan root. A test that pollutes the very tree the next CI run lints has coupled its cleanup to process liveness — the one thing you can't guarantee."
    },
    {
      "id": "F13",
      "title": "Within-batch concurrent createRole for the same role_key can race createdInBatch and double-create",
      "severity": "SPECULATION",
      "category": "correctness",
      "file": "apps/bot/src/shadow/role-writer.live.ts",
      "description": "The createdInBatch dedup map is checked synchronously, then guildFor()/roles.fetch()/create run asynchronously. With max_concurrent (e.g. 4) the gate may run two ops in parallel; two concurrent createRole calls for the same role_key would both miss the map and both reach the create. The WorldLock serializes across batches but not necessarily ops within a batch. This requires a batch to contain duplicate create intents for one role_key, which a well-formed planner should not emit.",
      "suggestion": "Record an in-flight promise in createdInBatch (store the create Promise, not just the resolved id) so concurrent same-key creates await the first, or rely on the gate to dedup create ops by role_key before dispatch.",
      "confidence": 0.3,
      "faang_parallel": "Go's singleflight package is the canonical fix: coalesce concurrent identical requests by storing the in-flight call, not just its result. This is also the check-then-act TOCTOU window that every memoization cache must close.",
      "metaphor": "Two waiters taking the same table's order because neither saw the other's ticket on the rail yet.",
      "teachable_moment": "An idempotency map keyed on completed results has a TOCTOU window equal to the operation's latency; storing the in-flight promise closes it. But verify the premise first (confidence 0.3): if the planner provably never emits duplicate create intents per role_key, this is a defense-in-depth nicety, not a live bug."
    },
    {
      "id": "F14",
      "title": "Conformance check couples a hardcoded SHA constant to bun.lock via a fragile regex",
      "severity": "LOW",
      "file": "apps/bot/src/shadow/substrate-conformance.ts",
      "description": "CANONICAL_SUBSTRATE_SHA/CANONICAL_EVENTS_PIN are inlined and must be hand-synced with substrate-sha.lock and bun.lock. The bun.lock SHA scrape requires a 40-hex match, which only matches the workspace-deps line (full SHA), not the abbreviated `#26d11b7` in the packages table; this depends on the full-SHA line appearing and on match() returning it first. A future bun.lock format change would silently fail to find a pin and report skew/false-pass risk.",
      "suggestion": "Read the canonical SHA from substrate-sha.lock as the single source of truth instead of inlining it, and tighten the bun.lock scrape to anchor on the workspace dependency entry explicitly.",
      "confidence": 0.4,
      "faang_parallel": "'Don't parse structured formats with loose regex' is the lesson behind the famous StackOverflow HTML-regex answer — and behind every lockfile-format bump that silently broke a homegrown scraper. The single-source-of-truth principle says derive, never duplicate, the canonical SHA.",
      "metaphor": "Reading the VIN off a fogged windshield — it works until the fog (lockfile format) shifts and you confidently read the wrong number.",
      "teachable_moment": "A conformance check whose failure mode is 'silently finds no pin' is worse than no check — it reports green by absence. Anchor the scrape to a named entry and treat 'no match' as a hard error, not a pass."
    },
    {
      "id": "F15",
      "title": "postinstall now hard-depends on python3 + bun in fixup-substrate-bun.sh",
      "severity": "LOW",
      "category": "reliability",
      "file": "scripts/fixup-substrate-bun.sh",
      "description": "The script (chained into the root postinstall under set -euo pipefail) uses `python3 -c` for relpath and `bun -e` for package-name reads. On a minimal build image lacking python3, the relpath call fails and, depending on guards, can abort install. The symlink-fixup is load-bearing for resolving the substrate's git-source subpath, so a failure here breaks the whole bot build.",
      "suggestion": "Replace the python3 relpath with a pure-bash/`realpath --relative-to` (or bun) implementation, and ensure every fixup failure path is a loud warning that does not abort postinstall.",
      "confidence": 0.35,
      "faang_parallel": "Build hermeticity again: assuming an interpreter is on a minimal image is the supply-chain fragility class that bit countless Alpine/distroless migrations where python3 simply isn't present. Under `set -euo pipefail`, a missing binary doesn't degrade — it kills the build.",
      "metaphor": "A doorbell wired to need the whole city power grid — a single missing fuse and the front door won't announce visitors.",
      "teachable_moment": "Load-bearing build steps must degrade loudly, not abort silently, and shouldn't assume a heavyweight interpreter for a one-line relpath. Verify first (confidence 0.35): if the build image is fixed and provably ships python3+bun, this is hardening, not a present break."
    },
    {
      "id": "F16",
      "title": "Fail-closed namespace guard correctly treats empty prefix as refuse-all",
      "severity": "PRAISE",
      "category": "security",
      "file": "apps/bot/src/shadow/role-writer.live.ts",
      "description": "assertNamespacedKey explicitly rejects an empty/missing prefix BEFORE the startsWith check, closing the `''.startsWith('') === true` foot-gun that would otherwise let a misconfigured world pass every role_key. Sharing one predicate across create, assign, and GC means the invariant cannot drift between mutation sites, and the behavior is directly proven by the iter-3 fail-closed tests.",
      "confidence": 0.9,
      "faang_parallel": "This is capability-based security done right: authority is scoped by a checked predicate, and the empty-prefix case is treated as 'no capability' rather than 'universal capability'. It's the same fail-closed default AWS IAM applies — an empty policy denies, it doesn't permit-all.",
      "metaphor": "A bouncer who checks both that you have a wristband AND that the wristband isn't blank — because a blank band that matches 'anything' is worse than no band at all.",
      "teachable_moment": "The deepest excellence here is one predicate shared across all three mutation sites: an invariant enforced in exactly one place cannot drift between create, assign, and GC. The `''.startsWith('')` trap is a famous foot-gun precisely because it fails *open*; guarding it before the prefix check is the move."
    },
    {
      "id": "F17",
      "title": "Member-cache hydration before the GC zero-member check is the right correctness fix",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "apps/bot/src/shadow/role-writer.live.ts",
      "description": "makeGatedRoleGc fetches members to hydrate the cache before reading role.members.size, preventing the failure mode where an un-hydrated cache reports 0 members and an assigned role gets deleted. The test that flips the member count on hydrate-all directly proves the guard reflects live membership.",
      "suggestion": "No change needed (see F1/F7 for the performance and intent-dependency caveats).",
      "confidence": 0.85,
      "faang_parallel": "Cache-warming before a read-and-decide is the same discipline as priming a CDN before cutover, or warming a JIT before measuring: you don't read a cold cache and act on its emptiness. The test that flips the count on hydrate proves the guard observes reality, not the cache's cold default.",
      "metaphor": "Turning the lights on before you count whether the room is empty — counting in the dark and finding 'nobody' is not the same as the room being empty.",
      "teachable_moment": "Reading a lazily-populated cache and treating absence as zero is one of the most common correctness bugs in client SDKs. Hydrate, then decide — and prove it with a test that distinguishes cold-cache-zero from truly-zero. The remaining work (F1, F7) is to make this both fast and assertive about completeness."
    },
    {
      "id": "F18",
      "title": "Import-boundary lint guards Map/Set .delete() false positives and collapses to a single grep pass",
      "severity": "PRAISE",
      "file": "scripts/lint-shadow-import-boundary.sh",
      "description": "Scoping the delete/edit patterns to role-ish receivers avoids flagging legitimate cache.delete()/Set.delete() calls (with a test that proves no false positive), and combining all patterns into one ERE alternation per file keeps the scan within the runner's latency budget. The header is also honest about the lint being accident-prevention, deferring the real boundary to the runtime gate.",
      "confidence": 0.8,
      "faang_parallel": "This is defense-in-depth done with intellectual honesty — Google's layered checks (lint → presubmit → runtime) work precisely because each layer is honest about what it can and can't guarantee. A lint that claimed to be the security boundary would invite complacency; this one defers to the runtime gate explicitly.",
      "metaphor": "A metal detector tuned not to beep at belt buckles — a detector that screams at every belt gets unplugged, and an unplugged detector catches nothing.",
      "teachable_moment": "False-positive discipline is what keeps a linter alive in CI: a noisy lint is a disabled lint. Scoping the pattern to role-ish receivers and proving the no-false-positive case with a test is the difference between a tool people trust and one they `// eslint-disable`. The honest header is the cherry — it tells future maintainers where the real boundary lives."
    }
  ]
}
```


## Callouts

Three things here are worth studying, not just merging:

- **[Security] F16 — the fail-closed namespace predicate.** Rejecting the empty prefix *before* `startsWith`, and sharing one predicate across create/assign/GC, is textbook capability scoping. The invariant lives in exactly one place, so it can't drift between mutation sites — and the iter-3 tests prove it. This is the kind of guard that quietly prevents an entire class of "misconfigured world strips every role" incidents.

- **[Correctness] F17 — hydrate-then-decide on the GC path.** Reading a cold member cache and acting on its emptiness is one of the most common SDK bugs in the wild; warming the cache before the zero-check, with a test that distinguishes cold-zero from true-zero, is the right instinct on an *irreversible* delete path. (Make it fast — F1 — and assertive about completeness — F7 — and it's airtight.)

- **[Maintainability] F18 — the honest import-boundary lint.** Scoping delete/edit patterns to role-ish receivers (proven no-false-positive), collapsing to one ERE pass for latency, and writing a header that admits the lint is accident-prevention deferring to the runtime gate — that combination of restraint and honesty is what keeps a linter from being disabled six months from now.

The unifying note: this PR's correctness instincts are excellent and its amplification/determinism instincts are the gap. Land the snapshot-per-batch idea (F1/F2), bound the unbounded paths (F5/F6), and pin the substrate deterministically (F3/F4) and this becomes substrate you can trust at guild scale.

PR #11, opened 2026-05-02
fix(orchestrator): Bedrock-aware SDK auth · close the chat-tool gap

zkSoju:
<details open>
<summary>the reorientation: Bedrock IS the provider</summary>

```mermaid
graph LR
  C[LLM_PROVIDER=bedrock<br/>+ AWS_BEARER_TOKEN_BEDROCK] --> R{resolveOrchestratorBackend}
  R -->|bedrock| BE[buildSdkEnv:<br/>CLAUDE_CODE_USE_BEDROCK=1<br/>+ AWS env passthrough]
  BE --> SDK[Anthropic Agent SDK<br/>routes via Bedrock]
  SDK --> M[Opus 4.7 inference profile<br/>us.anthropic.claude-opus-4-7]
  M --> T[mcp__score__* tools fire]
  R -->|anthropic| AE[buildSdkEnv:<br/>ANTHROPIC_API_KEY only]
  AE --> SDK
  classDef ship fill:#1f6f3a,stroke:#0f3d20,color:#fff
  class C,BE,SDK,M,T,AE ship
```
</details>

four prior fixes (v0.9.1, v0.10.0, v0.10.1, v0.11.0) didn't close it because I had the wrong mental model. Bedrock IS the operator's canonical Claude provider per Loa PR #662 — same Opus 4.7, routed via Bedrock. The Anthropic Agent SDK supports Bedrock via `CLAUDE_CODE_USE_BEDROCK=1`. Our orchestrator was hard-wired to STRIP that var to prevent bridge-mode auth, locking out Bedrock entirely.

🟢 fixed · 🟡 needs operator dev-guild deploy + reproduce

**fix layers**:

| layer | change |
|---|---|
| `orchestrator/index.ts` | `resolveOrchestratorBackend(config)` switch on LLM_PROVIDER · `resolveSdkModel(config, backend)` defaults Bedrock to `us.anthropic.claude-opus-4-7` · `buildSdkEnv(config, backend)` activates `CLAUDE_CODE_USE_BEDROCK=1` + AWS env passthrough when bedrock; sets ANTHROPIC_API_KEY otherwise · per-backend preflight replaces blanket guard |
| `compose/reply.ts` | `shouldUseOrchestrator` returns true for `provider in {anthropic, bedrock}` (was anthropic-only). Stub/freeside still naive. |
| `smoke-chat-routing.ts` | matrix updated: `auto + bedrock → orchestrator` (was naive) |

**verify after deploy** (with v0.11.0 telemetry):
[...]

PR #175, opened 2026-06-06
feat(converge): spotlight pfp via inventory-api getProfilePicture (#87 GAP-2)

zkSoju:
## What

Wires the periodic enriched-digest's spotlight profile picture to inventory-api's `getProfilePicture(wallet)` over HTTP, with the freeside_auth DB `pfp_url` as fallback. Closes GAP-2 from the shadow-mint-images coordination (issue #87).

- New `fetchProfilePictureHttp` in the existing `inventory-http-client.ts`, mirroring `fetchNftMetadataHttp` exactly (AbortController + timeout, the `safeImageUrl` host-allowlist, fail-soft on every error path).
- `digest-orchestrator.ts` pre-resolves the spotlight wallets' pfps in parallel (fail-soft per wallet), and the resolver becomes `inventoryPfp.get(w) ?? identities.get(w)?.pfp_url ?? null` (inventory primary, DB fallback).
- `INVENTORY_API_URL` lifted into the persona-engine config as optional (the same env the mint path already reads). Unset means dormant: no fetch, DB-only, identical to today.

## Why

The spotlight pfp was wired to a phantom `@freeside/inventory` npm import that never existed, so it was always dead and fell through to the DB pfp. inventory-api now serves the capability over HTTP, so the primary path can light up.

This depends on inventory-api PR #11 (the `GET /profile/:address` route), which is not on main or deployed yet. Until it deploys, this stays dormant by design and the digest renders exactly as before.

## Fail-soft is the invariant

A down, undeployed, slow, or malformed inventory-api must never break or block the weekly digest. Every failure path (missing baseUrl, timeout, non-OK status, throw, malformed JSON) returns null and falls back to the DB pfp or no image. Covered by dedicated tests, including a hung-timeout case and a throw case that both still render.

## Adversarial review caught one real bug

A case-sensitive Map-key mismatch: the inventory pfp map was keyed by the raw wallet string while score-api can emit mixed-case wallets, so the primary lookup could silently miss and degrade to DB. Fixed by normalizing keys and lookups to lowercase (`c8c48c3`), which also hardened the handle and badge lookups at the same site.

## Follow-up (not in this PR)

The cell still carries an in-process NFT pfp resolver (`resolve-nft-pfp.ts` via `enrichSpotlightPfp`). With inventory-api now the HTTP primary, that path is redundant for the spotlight surface and causes a double-resolve when `INVENTORY_API_URL` is set. Retiring it is a clean follow-up; left intact here to keep the blast radius minimal. Tracked as a follow-up bead.

## Verified

- `bun run typecheck`: clean across persona-engine and apps/bot.
- persona-engine suite: 1568 pass, 2 skip, 0 fail.
- apps/bot suite: 443 pass, 0 fail.
- Fully offline (mocked fetch).

Bead `bd-phe` (claimed; close on merge).

---


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#175. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "resolveInventoryPfps only stores fulfilled entries with non-null pfp — null responses are not cached",
      "severity": "MEDIUM",
      "category": "correctness",
      "description": "At line ~484 in the for-loop over `settled`, only `r.value.pfp` truthy values are set in the map (`if (r.status === 'fulfilled' && r.value.pfp)`). A wallet for which inventory-api explicitly returns `imageUrl: null` (no pfp on file) is silently absent from the map. This is fine for the primary lookup since the fallback chain `inventoryPfp.get(w.toLowerCase()) ?? identities.get(w.toLowerCase())?.pfp_url ?? null` handles a map miss identically to a stored null. However, if the same wallet appears twice in `wallets` (e.g., spotlight wallet repeated across zone entries) the fetch is issued twice since deduplication is not performed. More importantly, the implicit skip of `null` pfps means the shape of the returned map is semantically 'not present' rather than 'present with no pfp', which conflates two distinct states and could matter if a future caller needs to distinguish 'inventory answered null' from 'inventory was not asked'.",
      "suggestion": "Deduplicate `wallets` before issuing fetches (e.g., `[...new Set(wallets.map(w => w.toLowerCase()))]`). Consider storing `null` explicitly for wallets inventory-api confirmed have no pfp, to allow future callers to distinguish 'answered null' from 'not queried'. The fallback chain is unaffected either way.",
      "confidence": 0.75
    },
    {
      "id": "F-002",
      "title": "digestPfpLogger references MintEventSubscriberLogger but lives in the digest module — unnecessary coupling",
      "severity": "LOW",
      "category": "design",
      "description": "`digestPfpLogger` is typed as `MintEventSubscriberLogger` (imported from the mint event subscriber module). This couples the digest orchestrator to the mint subscriber's logger interface for a utility that is logically owned by the digest path. If the mint subscriber logger interface changes (e.g., adds required fields), the digest orchestrator is an unintended blast-radius victim.",
      "suggestion": "Extract a minimal logger interface local to `inventory-http-client.ts` (the actual consumer) and have `MintEventSubscriberLogger` extend or re-use it, rather than importing the mint-specific type into the digest orchestrator.",
      "confidence": 0.7
    },
    {
      "id": "F-003",
      "title": "Wallet passed to fetchProfilePictureHttp is not normalized to lowercase before the HTTP call",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/orchestrator/inventory/inventory-http-client.ts",
      "description": "`fetchProfilePictureHttp` receives the wallet as-is and composes the URL with `encodeURIComponent(opts.wallet)`. If the wallet has mixed-case hex characters (EIP-55 checksum format), the inventory-api endpoint must be case-insensitive on its side. The map-keying normalization in `resolveInventoryPfps` lowercases the wallet for map storage, but the actual HTTP request still sends the original-case wallet. If inventory-api does a case-sensitive DB lookup (e.g., exact match on `address`), a mixed-case wallet could get a 404 where a lowercase would succeed.",
      "suggestion": "Lowercase the wallet before URL composition inside `fetchProfilePictureHttp`: `const normalizedWallet = opts.wallet.toLowerCase()` and use that in the URL. This is consistent with the map-key normalization already applied by the caller.",
      "confidence": 0.55
    },
    {
      "id": "F-004",
      "title": "Test (c3) injects inventoryDoFetch but also passes fetchInventoryPfp: undefined — relies on internal resolver-selection logic",
      "severity": "LOW",
      "category": "test-fragility",
      "description": "Test case (c3) at line ~573 passes `fetchInventoryPfp: undefined` with the comment 'use the real (bounded) HTTP client path'. This means the test's correctness depends on `resolveInventoryPfps` preferring the HTTP-client path when `fetchInventoryPfp` is undefined but `inventoryApiBaseUrl` + `inventoryDoFetch` are set. If the resolver-selection order changes internally, this test could silently stop exercising the timeout path it claims to cover.",
      "suggestion": "Make the intent explicit: either assert that `fetchInventoryPfp` not being present causes the HTTP-client code path, or restructure the test to inject a named resolver that wraps `fetchProfilePictureHttp` with the hung-fetch stub, making the dependency on internal wiring unnecessary.",
      "confidence": 0.65
    },
    {
      "id": "F-005",
      "title": "PRAISE: fail-soft posture is comprehensive and consistently enforced across all layers",
      "severity": "PRAISE",
      "category": "reliability",
      "file": "packages/persona-engine/src/orchestrator/inventory/inventory-http-client.ts",
      "description": "The fail-soft contract ('a down / undeployed / slow / malformed inventory-api can NEVER break the digest') is enforced at three independent layers: (1) the HTTP client self-catches everything to null including AbortError, malformed JSON, and non-OK status; (2) `resolveInventoryPfps` wraps the entire per-wallet call in `Promise.allSettled`; (3) the fallback chain in `resolvePfp` has three ordered options. This is a well-layered defense that correctly treats enrichment as optional. The dormant-until-deployed behavior (empty string baseUrl → no fetch) is also a clean production-safety pattern.",
      "confidence": 1.0
    },
    {
      "id": "F-006",
      "title": "PRAISE: case-insensitive wallet normalization locked by regression tests with adversarial inputs",
      "severity": "PRAISE",
      "category": "correctness",
      "description": "Tests (g) and (g2) directly target the silent miss scenario where score-api returns a mixed-case wallet. The (g) test exports and directly asserts the map-key normalization contract; (g2) wires the full `composeDigestPost` path with an UPPERCASE spotlight wallet. This combination is a strong regression lock: it tests both the internal data structure invariant and the observable end-to-end behavior.",
      "confidence": 1.0
    },
    {
      "id": "F-007",
      "title": "PRAISE: test coverage for the HTTP client is hermetic and covers every non-happy path explicitly",
      "severity": "PRAISE",
      "category": "test-quality",
      "file": "packages/persona-engine/src/orchestrator/inventory/inventory-http-client.test.ts",
      "description": "The `fetchProfilePictureHttp` test suite covers: happy path, trailing-slash normalization, null imageUrl, off-allowlist host, non-https scheme, 404, 500, network throw, malformed JSON, AbortSignal-honoring timeout, empty baseUrl, whitespace-only baseUrl, and Accept header presence. Each failure mode is tested independently with precise assertions. The AbortSignal test correctly simulates real fetch abort behavior rather than using fake timers.",
      "confidence": 1.0
    },
    {
      "id": "F-008",
      "title": "identities map now keyed by lowercased wallet but resolveBadge already lowercased — inconsistency noted but pre-existing",
      "severity": "LOW",
      "category": "consistency",
      "description": "The diff changes `identities.set(r.value.wallet, ...)` to `identities.set(r.value.wallet.toLowerCase(), ...)` and updates `resolveHandle` and `resolvePfp` to `.toLowerCase()` on lookup. This is the right fix. However, it is worth noting that `resolveBadge` already had `.toLowerCase()` on lookup while the old `identities` map was mixed-case-keyed — the asymmetry was pre-existing and is now resolved for `identities`. No action needed, but the diff confirms this was a latent inconsistency that could have caused silent handle misses for mixed-case wallets.",
      "suggestion": "No change needed. This is a documentation note confirming the fix is complete.",
      "confidence": 0.9
    },
    {
      "id": "F-009",
      "title": "SPECULATION: resolveInventoryPfps is exported solely for white-box test access — may indicate a seam design smell",
      "severity": "SPECULATION",
      "category": "design",
      "description": "`resolveInventoryPfps` is exported (line ~434) primarily so test (g) can call it directly to assert the lowercased-key invariant. Exporting an internal orchestration helper for test access is a common pattern but does widen the public API surface. An alternative is testing the map-key invariant indirectly via the full `composeDigestPost` path (which test (g2) already does). If the direct export is retained, the function's public contract (including the DORMANT behavior and the lowercased-key guarantee) should be treated as stable.",
      "confidence": 0.4
    }
  ]
}
```


## Callouts

PR #180, opened 2026-06-24
feat(onboarding): per-world surface-config read-shim (DB-0, ships dark)

zkSoju:
**Un-trapping done work** — branch `81ae775` (4wk old, never PR'd) carried a complete, tested, net-new feature sitting local-only. Surfaced via /goal autonomous estate-cleanup; merged up to current main (**clean, zero conflicts**).

## what
A per-world **surface-config read-shim** so a CM can override onboarding copy/templates **without a code deploy** — and it **ships dark**:
- `migrations/0001_surface_config.sql` — `surface_config` table (world_id, surface, enabled, copy jsonb, template_json). **Idempotent** (`CREATE TABLE IF NOT EXISTS`), **manually applied** (this repo has no migration runner — `psql $RAILWAY_MIBERA_DATABASE_URL -f …`), on the **existing** mibera-db (consistent with CLAUDE.md "don't add a database" — it's a copy-override store, not new state-of-record).
- `onboarding/surface-config.ts` (+test) — `getSurfaceConfig`; returns null until a CM inserts an `enabled=true` row → **byte-identical to today**.
- small wires in `verify-card.ts`, `index.ts`, `freeside_auth/server.ts`, `post-type-gallery.ts`.

## verification status (honest)
- **Merges clean** into current main; net-new (not superseded by cycle-009 — `surface-config` is absent from main).
- The scaffold's **own files typecheck clean** — I confirmed zero errors in the touched files (the only typecheck errors in my local worktree are env-artifacts: vendored `@0xhoneyjar/events` wasn't rebuilt in a `--ignore-scripts` install).
- I could **not** run the full suite locally (worktree lacks the built substrate); commit claimed `surface-config` 3 pass + verify-card/gallery 9 pass + tsc clean 4wk ago. **CI + your review are the gate.**

## review asks
- **Schema design** of `surface_config` (the migration is the load-bearing artifact).
- Merging this PR does **not** apply the migration (manual/idempotent) and the feature is **gated dark** — safe to land and apply later.


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#180. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "BB-001",
      "title": "Promise.race timeout does not cancel the underlying PG query — connection stays checked out",
      "severity": "HIGH",
      "category": "reliability",
      "file": "packages/persona-engine/src/onboarding/surface-config.ts",
      "description": "At line 64, `Promise.race([query, timeout(SURFACE_CONFIG_QUERY_TIMEOUT_MS)])` prevents the application from waiting, but the `pool.query(...)` promise continues executing against the DB. The pg pool connection remains checked out until the DB-side query completes (or the server closes it). Under sustained DB degradation — the exact scenario this timeout is designed for — every call to `getSurfaceConfig` will hold a connection for the full DB-side duration. With a pool of default size (10) and 1500 ms of pending connections per call, a moderate request rate during a DB incident can exhaust the pool and starve `freeside_auth` of connections for wallet resolution.",
      "suggestion": "Pass `statement_timeout` via a dedicated pool client or a per-query config so the DB itself cancels the query when the timeout elapses. Example: acquire a client, SET LOCAL statement_timeout = 1500, run the query, and release the client in a finally block. Alternatively, configure the pool with a `statement_timeout` option at initialization time so all reads on this path self-cancel at the server.",
      "confidence": 0.95
    },
    {
      "id": "BB-002",
      "title": "Cross-layer import: onboarding/ → orchestrator/freeside_auth/ introduces upward dependency",
      "severity": "MEDIUM",
      "category": "architecture",
      "file": "packages/persona-engine/src/onboarding/surface-config.ts",
      "description": "Line 17 imports `getPool` from `../orchestrator/freeside_auth/server.ts`. The onboarding module now depends on an orchestrator-layer module. `freeside_auth/server.ts` owns wallet resolution, auth middleware, and now the pool singleton. If `freeside_auth` ever imports from `onboarding` (plausible: e.g., to check a surface config for auth flows), the import graph forms a cycle that Bun/TypeScript will silently hoist rather than reject. The CLAUDE.md construct boundary table (`freeside_auth` owns wallet→handle resolution; onboarding owns surface rendering) suggests these are separate concerns.",
      "suggestion": "Extract the pool singleton into a dedicated module, e.g., `packages/persona-engine/src/db/pool.ts`, exported from a neutral location. Both `freeside_auth` and `surface-config` import from there. This eliminates the upward dependency and keeps the pool lifecycle decoupled from auth concerns.",
      "confidence": 0.85
    },
    {
      "id": "BB-003",
      "title": "`cfg.copy` passed as `VerifyCardOpts` with no key-name validation — silent partial fallback",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/persona-engine/src/onboarding/verify-card.ts",
      "description": "Line 90: `buildVerifyCard(cfg?.enabled ? cfg.copy : {})`. `cfg.copy` is typed as `Record<string, string>` (from `asStringMap`). TypeScript accepts this as `VerifyCardOpts` structurally (all VerifyCardOpts properties are presumably optional strings), but there is no runtime check that the DB copy keys actually match `VerifyCardOpts` property names. A CM who stores `{\"header\": \"...\"}` instead of `{\"title\": \"...\"}` will get silent partial fallback — the enabled row exists and is returned, but `buildVerifyCard` renders defaults for the misnamed field with no warning. The CM sees the row is enabled and assumes the custom copy is live.",
      "suggestion": "Add a narrowing step between `asStringMap` and the return value: extract only the keys `VerifyCardOpts` declares (title, body, buttonLabel or whatever the interface exposes) and warn at read time if the DB row contains keys not in that set. This surfaces misconfigured CM rows in logs without breaking anything.",
      "confidence": 0.8
    },
    {
      "id": "BB-004",
      "title": "SQL migration has no `updated_at` auto-update trigger for direct UPDATE statements",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/migrations/0001_surface_config.sql",
      "description": "The `updated_at` column defaults to `now()` on INSERT. The example UPSERT in the migration comment sets `updated_at = now()` explicitly. However, if a CM runs a raw `UPDATE surface_config SET enabled = true WHERE world_id = 'mibera' AND surface = 'onboarding:verify'` without the explicit `updated_at` update, the column will reflect the original INSERT time, not the mutation time. Audit trails for \"when did this config change\" will be silently wrong.",
      "suggestion": "Add a trigger: `CREATE OR REPLACE FUNCTION surface_config_set_updated_at() RETURNS trigger LANGUAGE plpgsql AS $$ BEGIN NEW.updated_at := now(); RETURN NEW; END; $$; CREATE TRIGGER surface_config_updated_at BEFORE UPDATE ON surface_config FOR EACH ROW EXECUTE FUNCTION surface_config_set_updated_at();`",
      "confidence": 0.9
    },
    {
      "id": "BB-005",
      "title": "`asStringMap` iterates arrays as objects — numeric-keyed string values survive the coercion",
      "severity": "LOW",
      "category": "correctness",
      "file": "packages/persona-engine/src/onboarding/surface-config.ts",
      "description": "Line 44-51: `typeof [] === 'object'` is true, so a JSONB array value in `copy` (e.g., `[\"foo\", \"bar\"]`) passes the type guard and is iterated via `Object.entries`. String entries at numeric indices (`{\"0\": \"foo\", \"1\": \"bar\"}`) are included in the output map. This is unlikely to cause a real problem since `VerifyCardOpts` won't have numeric string keys, but the output is semantically wrong for array inputs.",
      "suggestion": "Add an `Array.isArray(value)` guard before the `Object.entries` loop and return `{}` for arrays: `if (value === null || value === undefined || typeof value !== 'object' || Array.isArray(value)) return {};`",
      "confidence": 0.75
    },
    {
      "id": "BB-006",
      "title": "Test suite covers only the no-URL path; no coverage for timeout or enabled-row paths",
      "severity": "LOW",
      "category": "test-coverage",
      "file": "packages/persona-engine/src/onboarding/surface-config.test.ts",
      "description": "All three tests exercise the single code path: env var unset → getPool() returns null → null returned. The timeout path (pool available, query hangs), the disabled-row path (pool available, row exists but enabled=false), and the happy path (enabled=true, copy applied) are untested. The PR comment acknowledges this is DB-0 scope, but the timeout path in particular exercises code logic (the Promise.race branch) that can be tested with a mock pool without a real DB.",
      "suggestion": "Add tests with a minimal mock pool: (1) mock that rejects after 2000 ms → assert getSurfaceConfig returns null (proves timeout branch); (2) mock returning `{rows: [{enabled: false, copy: {}, template_json: null}]}` → assert null; (3) mock returning an enabled row → assert the copy fields are reflected. These do not require a real DB.",
      "confidence": 0.7
    },
    {
      "id": "BB-007",
      "title": "PRAISE: Fail-soft contract is comprehensive and consistently implemented",
      "severity": "PRAISE",
      "category": "reliability",
      "file": "packages/persona-engine/src/onboarding/surface-config.ts",
      "description": "Every failure path — no pool, no row, disabled row, query error, timeout — returns null and logs a warning. The contract is documented at the module level, enforced by the catch block, and tested in `surface-config.test.ts`. The `enabled: boolean === true` check (not truthy) at line 79 is correct; `null` coerces false. The calling site `cfg?.enabled ? cfg.copy : {}` correctly applies the same guard. The SHIPS DARK / DEFAULT-OFF invariant is implemented at both the schema level (`DEFAULT false`) and the application level.",
      "confidence": 1.0
    },
    {
      "id": "BB-008",
      "title": "PRAISE: Pool reuse avoids a second PG connection lifecycle",
      "severity": "PRAISE",
      "category": "architecture",
      "file": "packages/persona-engine/src/orchestrator/freeside_auth/server.ts",
      "description": "Exporting the existing lazy singleton pool from `freeside_auth/server.ts` rather than opening a second pool in `surface-config.ts` is the correct call. Two pools against the same DB would double the idle connection count and fragment connection state monitoring. The exported function's JSDoc clearly states its purpose and the null-means-unavailable contract callers must honor.",
      "confidence": 1.0
    }
  ]
}
```


## Callouts

zkSoju:
**Addressed BB's findings (`fe4a45a`):**
- 🔴 **HIGH (connection leak)** — replaced `Promise.race([query, timeout])` (abandons but leaves the query on a checked-out connection → pool exhaustion under DB degradation) with a dedicated client + `BEGIN`/`SET LOCAL statement_timeout` (DB cancels server-side, auto-resets at txn end) + `release()` in `finally` on success **and** error. Tested: client released on success + on throw (no leak), null-pool short-circuit.
- 🟠 **MEDIUM (silent partial fallback)** — `buildVerifyCardForWorld` now narrows CM copy to the `VerifyCardOpts` keys and `console.warn`s ignored keys, so a misconfigured row (`{"header":...}` vs `{"title":...}`) is visible instead of silently rendering defaults. Tested.
- 🟠 **MEDIUM (onboarding→freeside_auth `getPool` import)** — **accepted, not refactored.** The pool reuse is a deliberate documented choice (avoid a 2nd connection), and `getPool` is load-bearing in the auth module. Extracting to a neutral `db/pool.ts` is the right follow-up but touches auth infra — deferred for your sign-off (no autonomous auth latitude).

onboarding suite 12 pass / 0 fail; surface-config + verify-card typecheck clean.

PR #178, opened 2026-06-24
feat(converge): spotlight pfp via inventory-api getProfilePicture (#87 GAP-2)

zkSoju:
## what

Closes the **spotlight pfp** real-data hookup from #87 §2 (GAP-2). The enriched-digest spotlight pfp was dead — wired to a phantom `@0xhoneyjar/inventory` import that never existed. `inventory-api` now serves it over HTTP at `GET /profile/:wallet → { address, contract, imageUrl: string|null }`.

## how

- **Wire through the existing http client** (`inventory-http-client.ts → fetchProfilePictureHttp`), mirroring `fetchNftMetadataHttp` exactly: `AbortController` + `setTimeout(abort)` + `clearTimeout` in `finally`, `accept: application/json`, missing `baseUrl → null`, non-OK → warn + null, any throw → null. Reuses the same `safeImageUrl` host-allowlist (`assets.`/`metadata.0xhoneyjar.xyz` https only) — same discord-card trust boundary.
- **Precedence**: inventory-api PRIMARY, `freeside_auth` DB `pfp_url` FALLBACK. `resolvePfp = (w) => inventoryPfp.get(w) ?? identities.get(w)?.pfp_url ?? null`. Pre-resolved once in parallel via `Promise.allSettled` over the spotlight wallets, so the synchronous `resolvePfp` reads a map.
- **Wallet-key normalization** (GAP-2 follow-up): the inventory pfp map is keyed by lowercased wallet so an UPPERCASE digest spotlight still resolves (`resolveInventoryPfps` exported for the regression).

## safety posture

- **dormant-until-deployed**: `INVENTORY_API_URL` (the SAME env the mint path reads) unset → the pre-resolve is skipped entirely (no fetch) → DB-only, identical to today.
- **fail-soft is sacred**: a down / undeployed / slow / malformed inventory-api can NEVER break or block the digest render — worst case the pfp is null and the DB fallback (or no image) is used.

## tests (hermetic, offline)

- 14 `fetchProfilePictureHttp` cases — happy / null / off-allowlist / non-OK / timeout / missing-baseUrl
- 25 orchestrator cases — inventory wins, DB fallback, throw→render, throw+no-DB→imageless, timeout→render, baseUrl-unset→no-fetch, real-fetch e2e, off-host→fallback, lowercase-key + uppercase-spotlight regressions

Verified locally on this branch: `inventory-http-client.test.ts` 14/14 · `digest-orchestrator.test.ts` 39/39 · `tsc --noEmit` clean.

[...]

PR #181, opened 2026-06-24
fix(converge): retire phantom @0xhoneyjar/inventory spotlight pfp resolver (#87 · bd-deg)

zkSoju:
Re-ships the **bd-deg** fix cleanly off current main (replaces #178 — see below).

## why a new PR
#178 turned out to **duplicate the merged #175** — the spotlight-pfp-via-inventory-api work (`getProfilePicture`) shipped to main via #175 (`13e264f`), and #178 had re-surfaced the same local commits, which is why it went CONFLICTING. main's `digest-orchestrator.ts` is byte-identical to #178's pre-fix base, so the *only* net-new, still-needed piece was this. Closing #178 in favor of this focused PR.

## the fix (verified)
main's #175 added the inventory-api HTTP pfp path but **left the dead phantom `resolveNftPfp`** wired as `enrichSpotlightPfp`'s default — it dynamic-imports `@0xhoneyjar/inventory`, a package present **nowhere** (confirmed still on main: `resolve-nft-pfp.ts:36`), so it throws → null on every call. It's also **redundant**: `resolveInventoryPfps` is already the real PRIMARY and `resolvePfp`'s precedence (`inventoryPfp ?? identities.pfp_url ?? null`) shadows it.

**Option A (zero behavioral change):** delete `resolve-nft-pfp.ts` + the dead `enrichSpotlightPfp`; `resolveSpotlightIdentity` returns `pickSpotlightDisplay` directly (handle + DB-pfp). The NFT branch was always-null, so output is unchanged — this only removes dead code + the last live phantom import in the digest path.

Invariants preserved (DB-pfp-wins, fail-soft catch→anon, host-allowlist, dormant-when-unset). New regression guard scans `src/orchestrator` for the dynamic-import *call* and asserts none (RED→GREEN verified). **persona-engine 1777 pass / 0 fail · apps/bot 803 pass · tsc clean** (verified this session on the identical base). `announce-mint.ts` untouched (already migrated; `bd-shadow-mint-image-repoint-gym` is blocked on inventory-api#8).


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/freeside-characters#181. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "BR-001",
      "title": "Guard test scope (`import.meta.dir`) narrower than the invariant description implies",
      "severity": "LOW",
      "category": "testing",
      "description": "The guard test walks `import.meta.dir`, which resolves to `src/orchestrator/` at test time. The test name says 'digest path' but `src/events/` (where `announce-mint.ts` lives) is outside the walk. The triage manually verifies that `announce-mint.ts`'s reference is a tombstone comment; the automated guard does not enforce this. If a future caller in `src/events/` or `apps/` introduced a live phantom import, the guard would not catch it. The reviewer.md's `git grep` covers the full repo but is a manual verification step, not an automated gate.",
      "suggestion": "Either (a) document the intentional scope limit in the test description string (e.g. '…in the orchestrator source tree'), or (b) widen the walk root to `src/` (one level up from `import.meta.dir`) to cover the full package's non-test source, which would also catch any future phantom in `src/events/`.",
      "confidence": 0.85
    },
    {
      "id": "BR-002",
      "title": "Walker does not exclude `.d.ts` files",
      "severity": "LOW",
      "category": "testing",
      "description": "The walker condition is `name.endsWith('.ts') && !name.endsWith('.test.ts')`. A `.d.ts` file satisfies both (ends in `.ts`, does not end in `.test.ts`) and would be read. In practice, no generated declaration files live under `src/orchestrator/`, so this is currently benign. However the predicate is imprecise and will silently include them if they ever appear.",
      "suggestion": "Add `&& !name.endsWith('.d.ts')` to the exclusion condition for correctness and future-proofing.",
      "confidence": 0.7
    },
    {
      "id": "BR-003",
      "title": "`import.meta.dir` creates an implicit location dependency for the walker",
      "severity": "LOW",
      "category": "testing",
      "description": "The walk root is `import.meta.dir`, which is the directory of the test file at runtime. If `digest-orchestrator.test.ts` is relocated (e.g. into a `__tests__/` subdirectory), the walker will silently scan a different directory with no warning, potentially missing the source files it was meant to guard. The inline comment `// packages/persona-engine/src/orchestrator` documents the current intent but does not enforce it.",
      "suggestion": "Anchor the walk root explicitly: `import.meta.dir.replace('/__tests__', '')` or resolve upward with `join(import.meta.dir, '..')` depending on where the test sits, with an assertion that the resolved path ends in `/orchestrator`.",
      "confidence": 0.65
    },
    {
      "id": "BR-004",
      "title": "PRAISE: Structural guard replaces behavioral tests for deleted function",
      "severity": "PRAISE",
      "category": "testing",
      "description": "Rather than leaving a test gap after deleting `enrichSpotlightPfp`, the implementation replaces the 5-case behavioral suite with a source-scan guard that asserts the phantom import call never re-enters the source tree. The regex targets the dynamic-import-call syntax specifically — not string mentions — so it avoids false positives from tombstone comments in sibling files while catching the actual failure mode. This converts a one-time deletion into an ongoing regression gate.",
      "suggestion": "No change required.",
      "confidence": 0.95
    },
    {
      "id": "BR-005",
      "title": "PRAISE: Zero-behavior-change deletion is rigorously proved",
      "severity": "PRAISE",
      "category": "correctness",
      "description": "The triage REFRAME establishes two independent reasons why removing `enrichSpotlightPfp` + `resolveNftPfp` is behaviorally inert: (1) the phantom import threw on every call → `nftResolver` always returned null → `enrichSpotlightPfp` always returned the identity unchanged; (2) even if the NFT resolver had succeeded, `resolvePfp`'s precedence (`inventoryPfp ?? identities.pfp_url ?? null`) means `resolveInventoryPfps` would shadow the result anyway. Both conditions are verified by the full suite staying green (1777 pass / 0 fail). The proof is two-sided — dead AND redundant — which fully justifies the scope of the deletion.",
      "suggestion": "No change required.",
      "confidence": 0.95
    }
  ]
}
```


## Callouts

PR #191, opened 2026-07-12
feat(shadow/ingestion): thj RoleSnapshot exporter → live shadow-audit seam (EXPORT-1)

zkSoju:
## EXPORT-1 — thj RoleSnapshot exporter

Builds the **incumbent-role side** of the Shadow Access Audit: enumerate the thj Discord guild, resolve discord → wallet, emit a contract-valid `RoleSnapshot`, and POST it to the deployed shadow-audit ingestion seam (`POST /v1/role-snapshot`).

Coordination tracker: 0xHoneyJar/freeside-characters#190 (master task EXPORT-1). Does **not** `Closes` it — that issue carries sibling tasks.

### The load-bearing decision: `--role-ids` is required and fail-closed

The consumer (`loa-freeside packages/services/shadow-audit/src/audit-service.ts`) reads **every** snapshot entry as a holder of the token-gated role:

```ts
const { roleWallets } = resolveRoles(snapshot);              // EVERY entry ⇒ a role-holder
const staleAccess = [...roleWallets].filter((w) => !qualifies(curBal, w));
```

`resolveRoles` never looks at `role_ids`. So exporting the whole guild would report **the entire server** as stale access. `--role-ids` is therefore mandatory, `@everyone` (role id == guild id) is refused, and there is deliberately no "export everyone" default.

### Other invariants

- **Unmatched holders are FLAGGED, never dropped** — a gated-role holder whose wallet doesn't resolve is emitted with `wallet` absent. Dropping them would understate the drift the audit exists to measure.
- **Contract is vendored, not imported** — `apps/bot` has zero `@freeside/*` deps and the schema lives at a subpath inside the loa-freeside monorepo (a git-URL dep can't install a monorepo subdir). `role-snapshot.contract.ts` is a zero-dep transcription of `RoleSnapshotSchema` @ `025cbe6d`; the vendored golden fixture + contract test are the **drift alarm**.
- **Integrity** — the body is serialized **once**; `X-Snapshot-Sha256` hashes exactly the bytes sent (a second `JSON.stringify` could reorder keys → 422).
- **PII floor** — no raw wallet or discord snowflake reaches a log line (redacted `first4…last4`); the ingest token is read from env and never logged.
- **At-least-once safe** — ingestion is monotonic on `captured_at`; `stored:false` is a successful no-op, not an error.

### Verification

- 75 tests pass / 0 fail in `src/shadow/ingestion`; `tsc --noEmit` clean.
- **Live Discord read proven**: `--list-roles` and a full `--dry-run` run against the real THJ guild (34,732 members read, 515 HC-role holders, 39,522-byte contract-valid payload).

### ⚠️ The live POST is deliberately NOT yet done

The dry-run against the real guild surfaced a blocker the unit tests could not:

**Only 1 of 515 HC-role holders resolves to a wallet.** A 25-member sample shows `GET /v1/resolve/account/discord/{id}` returning **404 for 25/25** — thj's members have no identity-api accounts at all. Their discord↔wallet links live in their incumbent system (CollabLand), not ours.

POSTing that snapshot would make the audit read **fresh** (clearing its `uncertain` flag) over a role-wallet denominator of **1**, with `newly_eligible` ≈ every Honeycomb holder. A confidently-wrong audit — the exact failure this cycle exists to prevent.

The exporter is correct; the missing piece is upstream. Two follow-ups are filed:
1. **Consumer gap** (loa-freeside): `unmatchedRoleHolders` is returned but does not drive `uncertain` and is absent from the aggregate. A mostly-unmatched snapshot must not present as confident. **This lands before the POST.**
2. **Identity coverage** (product): mounting onto an existing community needs its incumbent gate config + identity links imported (CollabLand) or an operator-assisted mapping flow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,23 @@
 > Retroactive reconciliation of v0.10.0–v0.11.3 is a separate hygiene task. The
 > entry below resumes the log at the cycle-008 state.
 
+## [Unreleased] — Shadow Access Audit exporter
+
+### Added
+
+- **THJ `RoleSnapshot` exporter** — `bun run --cwd apps/bot role-snapshot:export` enumerates an
+  explicitly selected gated Discord role, resolves Discord members to wallets, emits the strict
+  cross-repo snapshot contract, and posts it to the Shadow Access Audit ingestion seam.
+- **Fail-closed live writes** — live POSTs refuse empty gated cohorts or zero resolved wallets;
+  dry-runs remain available for diagnosis. Request deadlines include response-body reads, and a
+  malformed HTTP 200 receipt is treated as possibly stored rather than safe to retry.
+- **PII floor** — human logs redact Discord and wallet identifiers. `--list-roles` and `--dry-run`
+  are explicit sensitive stdout channels; `--out` files are forced to mode `0600`.
+
+### Tests
+
+- RoleSnapshot exporter/contract/client/CLI: 46 focused tests; full bot suite 890 pass, 0 fail.
+
 ## [0.12.0] — 2026-05-23 — cycle-008 capability-wiring + FAGAN-thorough
 
 ### Added
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -152,3 +152,25 @@ LLM_PROVIDER=anthropic \
 ```
 
 Validate the entire pipeline before touching production cron cadences. The `digest:once` CLI fires one post per zone and exits — useful for voice-iteration without waiting for Sunday midnight.
+
+### Export a gated-role snapshot for Shadow Access Audit
+
+```bash
+# Sensitive discovery output: redirect to a restricted operator file.
+bun run --cwd apps/bot role-snapshot:export --list-roles > roles.tsv
+
+# Diagnose the exact payload without writing to the audit service.
+bun run --cwd apps/bot role-snapshot:export \
+  --role-ids <gated-role-snowflake> \
+  --owner <community-owner-wallet> \
+  --dry-run > role-snapshot.json
+
+# Live POST: also requires ROLE_SNAPSHOT_INGEST_TOKEN.
+bun run --cwd apps/bot role-snapshot:export \
+  --role-ids <gated-role-snowflake> \
+  --owner <community-owner-wallet>
+```
+
+`--role-ids` is mandatory: exporting the whole guild would make every member look like a gated-role
+holder. Human logs are redacted; stdout and snapshot files contain raw identity data. `--out` forces
+mode `0600`, but redirected stdout inherits the shell's permissions—use `umask 077` when persisting it.
--- a/apps/bot/package.json
+++ b/apps/bot/package.json
@@ -14,7 +14,8 @@
     "test": "bun test",
     "conformance:check": "bun run scripts/check-substrate-conformance.ts",
     "lint:shadow-import-boundary": "bash ../../scripts/lint-shadow-import-boundary.sh",
-    "member-graph": "bun run src/cli/member-graph.ts"
+    "member-graph": "bun run src/cli/member-graph.ts",
+    "role-snapshot:export": "bun run src/cli/role-snapshot-export.ts"
   },
   "dependencies": {
     "@0xhoneyjar/events": "github:0xHoneyJar/loa-freeside#68f5a89cb02c6b3ddf5ab14a1d65753bc02bd9fe",
--- a/apps/bot/src/cli/role-snapshot-export.ts
+++ b/apps/bot/src/cli/role-snapshot-export.ts
@@ -85,6 +85,10 @@ Flags
   --owner <id>           REQUIRED. Provenance stamp (conventionally the community owner wallet).
                          The audit does not consume it.
   --community <name>     default: thj   (must be an OPERATED community, else the service 403s)
+  --collection <c/0x..>  default: 80094/0x886d...  the GATED COLLECTION (numeric chain / contract).
+                         thj gates 7 collections, each behind its own role — the snapshot is keyed by
+                         (community, collection), so exporting HJG1 under Honeycomb's key would make the
+                         Honeycomb audit compute drift against HoneyJar1's role-holders.
   --guild <id>           default: ${THJ_GUILD_ID} (THJ)
   --freshness <seconds>  default: ${DEFAULT_FRESHNESS_THRESHOLD_SECONDS}
   --endpoint <url>       default: ${DEFAULT_INGEST_ENDPOINT}
@@ -184,6 +188,19 @@ async function main(): Promise<void> {
   }
 
   const community = flag("community") ?? "thj";
+
+  // The GATED COLLECTION this role-set is for (S5-T1). Default: Honeycomb on berachain — the ONE
+  // operator-CONFIRMED gate edge (HC -> Honeycomb). Any deployment of a collection addresses it; the
+  // service canonicalizes across the union, so berachain and ethereum Honeycomb are the same key.
+  const collectionRaw = flag("collection") ?? "80094/0x886d2176d899796cd1affa07eff07b9b2b80f1be";
+  const [colChain, colContract] = collectionRaw.split("/");
+  if (!colChain || !/^[0-9]+$/.test(colChain) || !colContract || !/^0x[0-9a-fA-F]{40}$/.test(colContract)) {
+    log(`\u2717 --collection must be <numeric-chain-id>/<0x-contract>, e.g. 80094/0x886d...  (got '${collectionRaw}')`);
+    log("  A chain SLUG is rejected: it would be stored under a key the registry can never match, and the");
+    log("  snapshot would ingest happily and then be invisible to every audit.");
+    process.exit(2);
+  }
+  const collection = { chain: colChain, contract: colContract };
   const dryRun = has("dry-run");
   const endpoint = flag("endpoint") ?? DEFAULT_INGEST_ENDPOINT;
   const freshness = Number(flag("freshness") ?? DEFAULT_FRESHNESS_THRESHOLD_SECONDS);
@@ -208,6 +225,7 @@ async function main(): Promise<void> {
     built = await buildRoleSnapshot({
       guildId,
       community,
+      collection,
       owner,
       gatedRoleIds,
       members,
--- a/apps/bot/src/cli/role-snapshot-export.ts
+++ b/apps/bot/src/cli/role-snapshot-export.ts
@@ -1,0 +1,289 @@
+/**
+ * cli/role-snapshot-export.ts — the thj RoleSnapshot exporter (S3 / EXPORT-1). One-shot:
+ *
+ *   # 1. discover the token-gated role's snowflake (you need it for step 2)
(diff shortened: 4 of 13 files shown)


zkSoju:
## ✅ Live POST verified — the AC is met

Ran the exporter against the **real THJ guild**, POSTing to the deployed ingestion seam:

```
▸ read 34,732 guild members
▸ 515 gated-role holders · 🔗 1 resolved · ⚠️ 514 UNMATCHED (flagged, not dropped)
▸ POST 39,522 bytes → /v1/role-snapshot
✓ 200 · stored=true · community=thj · entries=515
```

**Monotonicity confirmed** — re-POSTing an older snapshot: `200 {"ok":true,"stored":false,...}` — a successful no-op, never a rollback. The exporter is safe to be at-least-once, as designed.

## The consumer fix landed first (deliberately)

This PR's dry-run surfaced that only **1 of 515** HC-role holders resolves to a wallet. POSTing that into the *old* audit would have produced a confidently-wrong report (`newly_eligible: 1813` — every Honeycomb holder — with `uncertain: false`).

So the consumer was fixed first: loa-freeside `d479cc23` (`fix(shared/shadow-audit): role coverage drives uncertainty`), now **deployed**. The live audit for thj today:

```
HTTP 422
{"error":{"code":"role-coverage-too-low",
  "reason":"only 1 of 515 role-holders could be resolved to a wallet (0% coverage; the floor is 50%)…",
  "retryable":false}}
```

An honest *"we cannot see your members yet"* instead of a confident fiction. Exactly what this exporter's `wallet`-omitted entries are for — **flagged, never dropped**.

## What this unblocks vs. what it doesn't

- ✅ The ingestion seam is proven end-to-end on real data.
- ⛔ **G-2 (thj sees real drift) is still blocked upstream** — thj's discord↔wallet links live in CollabLand, not identity-api. Tracked: `arrakis-9c68c` (source-agnostic identity ingestion). This exporter is correct; there is simply nothing to see until those links exist.

Ready for review.

zkSoju (commented):
## 🤖 cheval cross-model FAGAN council — automated review

**Verdict: CHANGES_REQUIRED** — at least one voice raised a blocking finding.

cheval-routed · 3/3 voices survived, 0 dropped · multi_perspective=true · verdict CHANGES_REQUIRED

### Per-voice verdicts
- **jam-reviewer-claude-headless** (anthropic:claude-headless): CHANGES_REQUIRED — 6 finding(s)
- **jam-reviewer-gpt** (openai:codex-headless): CHANGES_REQUIRED — 5 finding(s)
- **jam-reviewer-cursor** (cursor:cursor-headless): CHANGES_REQUIRED — 7 finding(s)

### Findings
**jam-reviewer-claude-headless** (CHANGES_REQUIRED):
- `[major]` L232 — Zero-signal snapshots are warned about but still POSTed — and the monotonic store lets them DISPLACE the last good snapshot
  ↳ fix: In role-snapshot-export.ts (CLI), a run with stats.gated_members === 0 (e.g. a valid-shaped but wrong role snowflake) or stats.resolved === 0 (e.g. wrong IDENTITY_WORLD / identity-api misconfig) only logs a warning and proceeds to POST. Because the service store is monotonic on captured_at (newer wins), this NEWER meaningless snapshot replaces the previously-good one — the exact 'confidently-wrong audit' the module header says it fails closed against, and the same failure the CLI already refuses for an unset IDENTITY_API_URL. Make both cases refuse before the POST (exit 1) unless an explicit override flag like --allow-degraded is passed; keep --dry-run permissive.
- `[major]` L140 — timeoutMs does not cover response-body reads — postRoleSnapshot can hang indefinitely after headers arrive
  ↳ fix: In role-snapshot-client.ts, clearTimeout fires in the finally block as soon as fetch resolves (headers received), but res.json() (200 path) and res.text() (refusal path) are awaited afterwards with no deadline and no abort signal. A stalled response body hangs the CLI forever, violating the module's own 'request deadline' contract and its 'never throws; every outcome is typed' promise. Move the body reads inside the timed scope (clear the timer only after the body is consumed) or wrap them in the same AbortController signal, mapping an abort to kind: 'transport_error'.
- `[cleanup]` L30 — Vendored-contract provenance header quotes an upstream schema WITHOUT `collection` and cites a wrong source path — the drift-pin file's own re-sync instructions are stale
  ↳ fix: In role-snapshot.contract.ts the 'verbatim from the source' RoleSnapshotSchema transcription omits `collection: SnapshotCollectionSchema` (the parser requires it, and the live upstream at apps/shadow-audit/src/role-snapshot.ts includes it — verified). The header also names the source path as packages/services/shadow-audit/src/role-snapshot.ts, but in the current consumer HEAD the canonical schema lives at apps/shadow-audit/src/role-snapshot.ts (packages/services/shadow-audit holds only delivery-log.ts). Update the quoted schema block and the source path/commit so the re-copy loop the file depends on actually works.
- `[cleanup]` L117 — Ready-event race in makeGetBotClient can hang the CLI
  ↳ fix: In cli/role-snapshot-export.ts, the clientReady/ready listeners are attached only after `await c.login(token)` resolves; if the ready event fires during that await, the promise never resolves and the one-shot CLI hangs. Guard with `if (c.isReady()) return client` (or resolve immediately when isReady()) before waiting on the once-listeners.
- `[cleanup]` L143 — Malformed 200 receipt is reported as kind:'refused' with status 200 — misleading, since the snapshot was likely STORED
  ↳ fix: In role-snapshot-client.ts, the 'refused' variant is documented as 'a typed refusal from the service (401/403/413/422/400)', but a 200 with an unparseable receipt returns {kind:'refused', status:200}; the CLI then prints 'the service refuses rather than guesses', which is false — the ingest almost certainly succeeded. Return a distinct kind (e.g. 'accepted_unverified' or 'malformed_receipt') or at least a reason stating the POST returned 200 and the snapshot may be stored, so the operator doesn't re-run needlessly or misread the state.
- `[cleanup]` L70 — flag() swallows a following flag as its value
  ↳ fix: In cli/role-snapshot-export.ts, `flag(name)` returns process.argv[i+1] unconditionally, so `--out --dry-run` writes a file literally named '--dry-run' and `--owner --dry-run` stamps '--dry-run' as provenance. Reject values that start with '--' (treat as missing and fail closed with the usage message).

**jam-reviewer-gpt** (CHANGES_REQUIRED):
- `[major]` L117 — Discord client can wait forever after login
  ↳ fix: Register the readiness listener before calling `login()`, or after `login()` only wait when `!c.isReady()`. `login()` may resolve after the ready event, leaving the subsequently attached listeners unable to fire.
- `[major]` L163 — CLI emits raw Discord snowflakes despite its PII contract
  ↳ fix: Do not print raw guild or role IDs in logs, including `--list-roles` and the gated-role summary. Redact human-readable output and provide deliberate sensitive output through a restricted file or explicitly defined payload channel.
- `[major]` L271 — Meaningless empty or fully unmatched snapshots are still posted and can replace valid state
  ↳ fix: Before the live POST, fail closed when `stats.gated_members === 0` or `stats.resolved === 0`; permit dry-run diagnostics if desired. A typoed role ID or identity outage currently produces a newer empty snapshot that the monotonic store may accept.
- `[major]` L137 — Malformed success receipts are accepted as successful
  ↳ fix: Validate the complete 200 response contract before returning `accepted`: require `ok === true`, boolean `stored`, non-empty string `community` and `captured_at`, and a non-negative integer `entries`. Currently `{ "stored": true }` is accepted and the CLI exits successfully with undefined receipt fields.
- `[major]` L158 — Datetime validator is looser than the vendored upstream contract
  ↳ fix: Add calendar-aware validation instead of relying on `Date.parse`; JavaScript normalizes impossible dates such as `2026-02-30T00:00:00Z`, while the documented upstream Zod datetime schema rejects them. Validate date components or require an exact UTC round-trip.

**jam-reviewer-cursor** (CHANGES_REQUIRED):
- `[major]` L217 — S3-T3 PII floor violated: raw Discord snowflakes logged on stderr
  ↳ fix: The file header (lines 19–24) requires every identifier in log lines to pass through `redactDiscordId` / `redactWallet`, but `log(\`▸ guild ${guildId} · community '${community}' · gated roles: ${gatedRoleIds.join(', ')}\`)` emits the raw guild id and gated-role snowflakes on stderr. Wrap `guildId` and each gated role id with `redactDiscordId` (same pattern used at lines 252–255 for the sample cohort).
- `[major]` L161 — S3-T3 PII floor violated: raw guild snowflake on stderr in --list-roles
  ↳ fix: `log(\`▸ guild ${guildId} · …\`)` prints the full guild snowflake to stderr while the module promises no raw snowflakes on stderr. Redact with `redactDiscordId(guildId)`; role ids needed for operator copy-paste can stay on stdout (line 163) if documented as an intentional exception.
- `[major]` L145 — 200 receipt accepts responses without verifying `ok: true`
  ↳ fix: `postRoleSnapshot` only checks that `stored` is boolean (lines 145–148) but the documented contract is `200 { ok, stored, community, captured_at, entries }` and the malformed-receipt error message explicitly names `ok`. A `200` body with `ok: false` (and a boolean `stored`) is treated as `accepted`. Require `receipt.ok === true` (and ideally validate `community` / `entries` types) before returning `{ kind: 'accepted' }`; add a test for `{ ok: false, stored: false, … }`.
- `[major]` L19 — S3-T3 policy contradicts --list-roles stdout behavior
  ↳ fix: Header states no raw Discord snowflake ever reaches stdout/stderr (with only `--dry-run` / `--out` exceptions), but `--list-roles` writes raw role snowflakes to stdout (lines 162–163). Either amend the S3-T3 block to document `--list-roles` stdout as an intentional operator-discovery exception (parallel to `--dry-run`), or redesign discovery output so stderr stays redacted and stdout policy is explicit.
- `[cleanup]` L207 — Invalid `--freshness` bypasses fail-closed CLI error handling
  ↳ fix: `const freshness = Number(flag('freshness') ?? …)` yields `NaN` for non-numeric input; `buildRoleSnapshot` then throws `RoleSnapshotContractError` (not `RoleSnapshotExportError`), which is not caught at lines 237–243 and surfaces as an unhandled contract error. Parse/validate freshness in the CLI (positive integer) and exit 2 with a clear message before calling `buildRoleSnapshot`.
- `[cleanup]` L237 — Discord client may leak on non-RoleSnapshotExportError paths
  ↳ fix: `getBotClient()` is created and used for guild reads, but `client.destroy()` runs only on successful/list-roles/dry-run/POST paths. If `buildRoleSnapshot` throws `RoleSnapshotContractError` or an unexpected error after login, the client is never destroyed. Wrap the post-login body in `try/finally` and call `client.destroy()` in `finally` whenever a client was obtained.
- `[cleanup]` L28 — Vendored zod mirror comment omits required `collection` field
  ↳ fix: The provenance comment block (lines 28–38) still transcribes `RoleSnapshotSchema` without `collection`, while `SNAPSHOT_KEYS` and the golden fixture include it. Update the mirrored zod snippet to include `collection` so the next manual sync does not drop a load-bearing field.

> Routed through the LOCAL cheval council (subscription CLIs — no API key; cheval cannot
> run in GitHub Actions). Per-voice model attribution + verdict-quality envelopes are
> recorded in `.run/model-invoke.jsonl` (MODELINV audit). This review is the cross-model
> gate satisfying one required approval (council = 1).

